PR #33, opened 2015-01-26
update

mariapirespacheco:


rmtfleming:
Hi Maria,
please use the location 
https://github.com/opencobra/cobratoolbox/tree/master/modelGeneration/fluxConsistency/FASTCORE
for the updated files.
Avoid renaming the files as we only want to be able to see the updates before merging the request. Also, there should be no .mat files outside the testing folder.
Regards,
Ronan

mariapirespacheco:
Hi,
ok will change it.
Best
Maria

2015-01-26 18:56 GMT+01:00 Ronan M.T. Fleming <notifications@github.com>:

PR #34, opened 2015-01-27
latest

mariapirespacheco:


mariapirespacheco:
Hi, 
Is there still something missing ? 
Best
Maria

rmtfleming:
Hi Maria, I looked before at the files, why does it show a cut and paste view of what is changed rather than the individual lines? The former means I have to check manually much more lines, which needs a longer window of time. You could help by only editing the parts of the files you change, then committing that.

mariapirespacheco:
Hi,
No problem I will do that.
Best
Maria

2015-02-17 13:23 GMT+01:00 Ronan M.T. Fleming <notifications@github.com>:

mariapirespacheco:
Hi,
The files have been updated again. Now it should show the changes as requested. In addition a switch between the original cplex code and the cobra code was implemented. It appears, that the glpk solver often has issues with the algorithm for no apparent reason while it is working without problems using CPLEX.
Let me know if there are any further issues.

PR #38, opened 2015-02-02
Master

akaviaLab:
Made the code of removeRxns.m simpler, and easier to add fields


rmtfleming:
This commit cannot be automatically merged and there are a lot of changes in it on many files, some of which may be old versions, from before the time when I hand merged the old cobra-devel version with the github version, merged with master on 30th Jan, therefore I am reluctant to merge this pull request as a monolithic block.

akaviaLab:
Okay. I'm closing this pull request. I'll try to create a branch which shows only the small amount of changes I made and use a pull request on it.

PR #46, opened 2015-04-29
A problem with initCobraToolbox.m is fixed

longfeimao:
Some errors related to calling "TranslateSBML" function in "initCobraToolbox.m" are fixed.


longfeimao:
Another minor error is fixed in the initCobraToolbox.m

PR #56, opened 2015-07-08
Update estimate_pKa.m for cxcalc 15.6.15.0

zappingseb:
Line 72-79: Due to the infeasibility of calculating the pKa for H+ and H2 these are eliminated by changing the bool function
Line 94-119: Due to the new format of the cxcalc version 15.6.15.0 output, first the table headers have to be eliminated, further on the lines have to be split and kommas and dots have to be exchanged
Line 156: The exchange of kommas and dots does not have to be used here anymore.

### DIFF
--- a/thermo/vonBertalanffy/estimate_pKa.m
+++ b/thermo/vonBertalanffy/estimate_pKa.m
@@ -68,7 +68,16 @@ else
 end
 
 % Only estimate pKa for unique metabolites to increase speed
-bool = ~cellfun('isempty',inchi);
+bool
+% do not estimate for H+ and H2
+for i=1:length(inchi)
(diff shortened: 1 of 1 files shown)


huldash:
Hi Sebastian, I was unable to run your updated function. Line 71 gave me an error and line 95 caused problems in line 138. Also, the variable newesresult that you create in lines 107-119 is not used. What was the original bug? Older versions of chemaxon also output a header for pKa estimates, which was removed in lines 102-104 of the original code (lines 138-140 of updated code).

zappingseb:
Hi Hulda,

I corrected the errors. Sorry it took so long, but I was on holidays

Cheers,

Sebastian

PR #57, opened 2015-07-09
Commented out all "pause(eps)" statements

keithma:
This is a bugfix for a problem identified in solvers/getCobraSolverParams.m.
Profiling results indicated that pause(eps) calls accounted for a large majority
of the time spent in this function (and in the program that called it). Further
discussion with Ronan Fleming confirmed that this behavior is not intentional,
and the pause statement should be removed. In this commit I remove it from all
(non-doc) files.

See here for our (brief) communication in the Google group:
https://groups.google.com/forum/#!topic/cobra-toolbox/spNdyPaPtG8

Keith Ma
Research Computing Services
Boston University

### DIFF
--- a/external/m2html/mdot.m
+++ b/external/m2html/mdot.m
@@ -48,7 +48,7 @@ else
             URLnames{i}=mfiles{i}(length(options.mFiles{1})+2:end-2);
         end
     end
-    pause(eps)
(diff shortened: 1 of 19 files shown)


rmtfleming:
Thanks a lot Keith!

PR #58, opened 2015-07-09
Cleanup/init cobra toolbox

keithfma:
"Cleaned up" the first half of the initCobraToolbox function. The changes are mostly cosmetic, and aimed at making it easier for the user to understand what this function does. I stopped editing at the SBML installation section.


rmtfleming:
I don't understand why one would want to remove the reporting of solver status during initialisation of the cobra toolbox (e.g. as below). Please could you explain the idea?

-    if LPsolverOK
-        fprintf('LP solver set to %s successful\n',char(CobraLPSolver));
-    	break;
-    end
+    if LPsolverOK; break; end

PR #60, opened 2015-07-11
Bugfix/mcc

keithfma:
This branch fixes several errors that prevented compilation of a standalone executable including CobraToolbox.  

During compilation, mcc parses all the source files to search for dependencies. This process turned up several typos in CobraToolbox files, as well as a few other problems specific to compiling a standalone (e.g. empty files are not allowed, .m extension implies a MATLAB script or function). I fixed these so that I could successfully compile, and did my best to ensure that any changes I made do not change the behavior of the program at all. 

Keith Ma,
Boston University RCS


rmtfleming:
Hi Keith, the summer holidays are over so there will be time for us to look into your contribution here. Sorry about the delay.

keithfma:
Thanks for looking into this!

PR #65, opened 2015-08-26
added missing negFluxAllowedFlag to reduceModel

Bounciness:
Also changed sampleCbModel to allow negative fluxes.


rmtfleming:
model.rev will be depreciated in favour of model.lb and model.ub so it would be consistent not to start relying on model.rev

PR #91, opened 2015-12-08
Generalize findRxnsFromGenes for gene identifiers.

hhefzi:
Function now can parse gene ids appropriately even if starting with a
digit, without user input flag.

Also checks for nonstandard characters in geneids (anything but letter, digit, underscore) and replaces them for parsing within the function.

### DIFF
--- a/findRxnsFromGenes.m
+++ b/findRxnsFromGenes.m
@@ -1,4 +1,4 @@
-function [results ListResults] = findRxnsFromGenes(model, genes, humanFlag,ListResultsFlag)
+function [results ListResults] = findRxnsFromGenes(model, genes, ListResultsFlag)
 %findRxnsFromGenes print every reaction associated with a gene of interest
 %
 % [results ListResults] = findRxnsFromGenes(model, genes, humanFlag,ListResultsFlag)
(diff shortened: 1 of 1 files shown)


rmtfleming:
Please can you write a function testFindRxnsFromGenes to verify that your edits work as intended?

hhefzi:
Doesn't appear I can attach .m files here but the code is simple enough:

function status = testFindRxnsFromGenes(model)

a=[];
b=[];

%Checks to see if first gene starts with a digit, if not, prepend a 1
if ~regexp(model.genes{1}(1),'\d')
    model.genes=strcat('1',model.genes);
end

%See if function works for first gene
try
a=findRxnsFromGenes(model,model.genes(1));
end

%Create a model with gene ids containing disallowed characters (here I use
%colon)
model_badgeneids=model;
model_badgeneids.genes=strcat(model.genes,':appended');

%See if function works for a gene containing a disallowed character
try
b=findRxnsFromGenes(model_badgeneids,model_badgeneids.genes(1));
end

%Only return true if both results are not empty
if ~isempty(a) && ~isempty(b)
    status=1;
else
    status=0;
end
end

PR #102, opened 2016-01-12
Fixes adding external constraints used by optKnock.

AndrasHartmann:
It works with optKnock, but should be tested through wherever this method is used.


rmtfleming:
Please can you run the necessary tests?

AndrasHartmann:
Seems that this was not a good idea after all. Causes the test fail in testing/testOptKnock 
Created a pull request to revert these changes.

PR #103, opened 2016-01-21
updated findRxnsFromMets 

tpfau:
Updated findRxnsFromMets to accurately reflect the verboseFlag description (printing the reactions, irrespective of number of outputs) and added options for Property/Value pairs. 
In addition, I added a new property, allowing the selection of reactions which have to contain all of the provided metabolites.
The function is still compatible with any previous code (with the exception of actually providing output if verboseFlag is active irrespective of the number of requested outputs).


tpfau:
There still seems to be a bug.

PR #122, opened 2016-04-06
SBML level 3 without FBC

aebrahim:
I received a file written by the COBRA toolbox with SBML level 3 without FBC v2, which sounds like a bug to me. Looking at ```io/writeCbToSBMLfbc.m``` it seems like this type of file can get written pretty easily by the COBRA toolbox, so this patch removes that.

There are already so many different SBML formats for COBRA models, and we shouldn't have yet another one floating around.

I think it should just be:
 - SBML2 without fbc
 - SBML3 with fbc

Having sbml3 without fbc creates just another SBML variant that will be floating around until the tend of time. I'm also adding a warning into cobrapy whenever these types of models are detected.

### DIFF
--- a/io/writeCbToSBMLfbc.m
+++ b/io/writeCbToSBMLfbc.m
@@ -1,30 +1,21 @@
-function writeCbToSBMLfbc(model,fileName,fbc)
+function writeCbToSBMLfbc(model,fileName)
 %
 % Write a COBRA model to a SBML with FBC file
 %
 %INPUTS
 % model         COBRA model structure fileName      
 %
-%OPTIONAL INPUT
-%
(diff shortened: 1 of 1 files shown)


huldash:
Thank you for the pull request Ali. It makes sense to me. @longfeimao do you agree?

longfeimao:
"io/writeCbToSBMLfbc.m" is a "relic" produced in last September.  We have two new functions "readSBML" and "writeSBML" to read and write SBML-FBCv2 files. These new functions are now only available within the SBG group, and I hope they can be made public as soon as they are rigorously examined.

aebrahim:
Thank you for merging :)

PR #134, opened 2016-05-25
Updated verbosity of Fastcore Test Function and added fastCC test Function

tpfau:
Updated the verbosity of the FastCore testfunction. Its now silent.
Also, updated the fact,, that ibm_cplex is an option for both LP and MILP solvers, and added functionality that removes another log file often created by cplex to the solveCobraLP function.


huldash:
Thank you for all these improvements. testFASTCORE fails on my system because numel(A) = 5065, not 5064 like the reference solution. Is this a bug or is the solution solver dependent? Is the test too strict?

tpfau:
Hmm.
The solution of Fastcore is not necessarily unique, so yes, it is possible, that two different runs can yield different results.
Thus yes, maybe the test is to strict, and we could allow just testing whether the algorithm is providing a non empty result.

PR #135, opened 2016-05-25
updated testFASTCORE

tpfau:
Another update for testFASTCORE.
Unfortunatly, the only way to test this machine independent is to test, whether the results provide the specified properties (i.e. the resulting reaction set must allow all reactions from the core set to carry flux). This is now properly checked.


huldash:
Great, thanks. It works. I replaced the for loop with the fluxVariability function which does the same thing.

PR #136, opened 2016-05-27
ReconMap functionalities

bertonoronha:
This functionality enables the user to interact with the ReconMap (https://vmh.uni.lu/#mapnavigator)

* Flux distribution overlays

It is possible to upload a flux distribution which will be displayed in the ReconMap. Fluxes are normalized and given a specific thickness and colour.

* Subsystem overlays

Subsystem overlays are also possible, being possible to generate overlays for specific subsystems or all at the same time.


rmtfleming:
Alberto, please announce this on the cobra toolbox google group when the paper is on arxiv.

PR #149, opened 2016-08-25
Some modifications on FASTCC to lower CPU-time

Ojami:
Added lines to original LP7 of FASTCORE have been fully documented within the added script.


rmtfleming:
Edit fastcc.m, then commit that so that we can see clearly what the changes are. The lines should each be properly commented. Add one extra argument to use the previous version by default and the new version if specified.
https://github.com/opencobra/cobratoolbox/blob/master/modelGeneration/fluxConsistency/FASTCORE/fastcc.m

Ojami:
LP7 has been edited not the fastcc.m, so should I change LP7?

PR #150, opened 2016-08-26
Changes to better handle the response from MINERVA.

bertonoronha:
When submitting a new overlay remotely to ReconMap the MINERVA instance sends back a response. The changes will now transform that response to a matlab cell structure where the index 1 has a 0/1 value for wether the overlay was successfully sent, and the index 1 a success/error message.


huldash:
Hi Alberto,

Looks good but could you remove the asv file?

Thanks,
Hulda

PR #152, opened 2016-09-06
Update fastcc.m

Ojami:
The original fastcc script was modified to reduce the cpu time of the original code. Two solvers of 'glpk' and 'gurobi5' were used to compare the runtime results for five different COBRA models: Recon1, Recon2, iAF1260, iECW_1372 and iSF_1195). The results of two versions of the script are the same, and the CPU time of the proposed script was about an order of magnitude less than that of the fastcc (with the exception of large model such as Recon 2, which original fastcc took even more time).


rmtfleming:
@syarra please check this

laurentheirendt:
Cherry-picked commit is [here](https://github.com/laurentheirendt/cobratoolbox/commits/pr-152).

PR on hold until PR #317 merged (test required).

cc: @syarra

laurentheirendt:
Closing for now as discussed [here](https://github.com/laurentheirendt/cobratoolbox/commit/e19d827ea16b09dafcfcd2bb9012ed9b6d698830#commitcomment-21326177).

PR #160, opened 2016-10-12
Additional IF-conditions for improving robustness and ignoring .db files

laurentheirendt:
No conflicts with existing scripts reported.


rmtfleming:
Why does this not want to merge automatically?

PR #166, opened 2016-11-09
Fixed readSBML

simas232:
Fixed an issue regarding the incorrect inversion of the objective
function coefficient value (from 1 to -1), when importing FBCv2 model
to MATLAB. The problem is due to readSBML function, which ignores the
objective function coefficient value in modelSBML structure and only
relies on fbc_type.

Now readSBML function firstly checks, whether there is a defined
objective function coefficient value in modelSBML and only if it is
absent, fbc_type value is considered.


laurentheirendt:
Thanks @simas232. Before accepting the PR, may provide a minimal testing script? Thanks.

laurentheirendt:
Thank you, @simas232.

PR #175, opened 2016-11-23
Update readSBML.m

almut-heinken:
The objective sense is currently written into model.c. If the objective function in the SBML model is defined as maximized, model.c will have a value of -1 for the objective function, which results in a flux of zero when the objective function is maximized, and a negative flux when an irreversible reaction is minimized. The objective sense should instead be defined in model.osense.

### DIFF
--- a/io/utilities/readSBML.m
+++ b/io/utilities/readSBML.m
@@ -628,7 +628,8 @@ else    % in the case of fbc file
     if noObjective==0; % when there is an objective function
         indexObj=findRxnIDs(model,fbc_obj);
         % indexObj=find(strcmp(fbc_obj,model.rxns))
-        model.c(indexObj)=fbc_obj_value;
+        model.c(indexObj)=1;
+        model.osense=fbc_obj_value;
     end
     
     if all(cellfun('isempty',fbcMet.fbc_chemicalFormula))~=1  % if all formulas are empty


rmtfleming:
model.c can be of mixed sign, whether it is maximised or minimised depends on model.osense so I cant see why model.c should be "positive at all times".

almut-heinken:
Hi Ronan, if the objective is e..g, biomass, model.c is -1 for the BOF and model.osense is also -1, you can a flux of zero when optimizing the BOF (in a model that should produce biomass flux). If model.osense is 1, you get a negative biomass flux. So having a model.c of -1 results in objective values that aren't biologically meaningful.

rmtfleming:
And if one flips the signs of the biomass reaction and bounds?

almut-heinken:
If the sign of the negative biomass flux obtained was flipped, then you would get the correct biomass value. But that seems very confusing and non-intuitive.

io/utilities/readSBML.m
```diff
@@ -628,7 +628,8 @@
     if noObjective==0; % when there is an objective function
         indexObj=findRxnIDs(model,fbc_obj);
         % indexObj=find(strcmp(fbc_obj,model.rxns))
-        model.c(indexObj)=fbc_obj_value;
+        model.c(indexObj)=1;
+        model.osense=fbc_obj_value;
```
laurentheirendt: `model.osense = - sign(fbc_obj_value);`

cc: @rmtfleming

laurentheirendt: Fixed in PR #290.

PR #184, opened 2016-12-09
also warning when single input reaction ID is not found

karthikraman:
Currently, "named" warnings are given only for the reactions not found, when the input is a list. If a single reaction is input, either as a cell array of size 1, or as a single char array, no "named" warnings are given, indicating what reaction is not found. 

While this may be by design, sometimes, if there are a bunch of `changeRxnBound` commands inside a script, such warnings will be more useful.

Also, converted the `fprintf` for rxns not found, to a `warning`.


karthikraman:
This ends up warning twice for single cells. Need to fix. Sorry!

PR #187, opened 2016-12-09
Conflict fixed with main fork

lemmerelassal:
I was behind on some versions, had to reclone and fix this (thanks Sylvain!).


syarra:
@lemmerelassal Could you please respond to @rmtfleming's question?

PR #194, opened 2016-12-16
Updated Read, to add up Substrates and Products

tpfau:
SBML allows that a single molecule can be defined multiple times as a reactant and/or product. the current version of the readSBML only considers the last appearance of a species in the list of Reactants/Products. In particular, it will ignore a reactant definition if there is a product definition and in this instance only keep the product coefficient.
While this is no issue when importing COBRA generated models, it can be problematic when non Toolbox generated models are imported.
I would therefore suggest to add this modification which simply adds up all coefficients (negative for substrates).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,15 +10,13 @@ script:
 
     # change to the current directory
     - cd $CURRENT_DIR
-    
+
     # launch the tests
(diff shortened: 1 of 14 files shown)


rmtfleming:
In discussion with SBML folk at present...

PR #201, opened 2017-01-05
Modularise flux split code

huldash:
:art: :white_check_mark: Split part of the code into a new function that enables computation of flux splits in an arbitrary flux distribution, not necessarily one that minimises the euclidian norm of fluxes.


laurentheirendt:
Thanks @huldash and @syarra!

Issue #200, opened 2017-01-05
checkMassAndChargeBalance does not show unknown formulas as unbalanced


tpfau:
I'm just wondering, whether this is intended behaviour or not.
When performing checkMassAndChargeBalance, a missing formula for a metabolite will not lead to reactions containing this metabolite being flagged as unbalanced. Is this intended, if, it should be documented, if not (and I personally would support this) I think the function should be altered to indicate all reactions with metabolites with missing formulas as unbalanced. 
My personal stance would be: If a user is aware of placeholder metabolites in their model he should be knowing which reactions contain those, but not flagging them can easily lead to a user missing those. Especially as the function, while flagging the metabolites, does not indicate any issued with those reactions (except if turning the printlevel to 1, but those messages tend to be easily missed in a larger script).
I'm happy to implement a modification, but I would first want to know if the assumption, that reactions with metabolites that miss a formula should be flagged as unbalanced or not.

rmtfleming:
Hi Thomas,
this seems a sensible modification. I thought that it does flag the metabolites without formulae, but it should also do so for the reactions with one or more missing formula(e).
Regards,
Ronan

tpfau:
Hi Ronan, 
One more question:
As of this moment, checkBalance will return NaN if any metabolite is unknown. While this is a sensible approach to indicate that the method can't determine the balance, it might be useful to instead return the "unbalance" value ignoring the unknown metabolite. 
Obtaining this value would allow a user to estimate the atom composition of unknown compounds if only one compound is missing its formula.
While this is a different issue, I would suggest to at least add an option to the function to allow this behaviour, while keeping the default to return NaN to not mess with established code.

rmtfleming:
OK, please could you add some test functions for these tests of elemental balance? We are trying to increase the test coverage for v3.

tpfau:
I created a pulll request. Incidently I noticed, that the chargeBalance part of the function did not work at all. It was never ever set to anything. I fixed that along with the update.

PR #202, opened 2017-01-09
Reactions with unknown Reactions are now flagged as unbalanced 

tpfau:
This is a change for the issue discussed here:
https://github.com/opencobra/cobratoolbox/issues/200

It simply flags all reactions that contain unknown metabolites in the result from checkMassChargeBalance. Further it leads to massImbalance displaying the NaN value that those reactions should have.


tpfau:
Recreating a new pull request that is using the new file structure

laurentheirendt:
PR #253.

Thanks @tpfau .

PR #203, opened 2017-01-10
Improved SBML IO

tpfau:
The current SBML io is unable to restore e.g. rxnNotes. It is also prone to notes not having the specified format (<tag>CATEGORY: categorydata </tag>), since it only checks for the CATEGORY to be present anywhere in the string  at which point it might replace the proper category by the latter.
It was also unable to handle notes (or notes without a tag properly, these could only ever be read if they were written after a Confidence score (for whatever reason).

The updated version relies on some functionality introduced after 2013a, and will therefore check the version and use a corrected (at least with respect to additional Notes handling) version of the old code if the employed Matlab version is too old.

finally I added the NOTES category to the output comments again, as other tools that try to parse COBRA annotations rely on all fields having a specified structure as defined in the original COBRA paper (and shown above).

### DIFF
--- a/io/utilities/parseSBMLNotesField.m
+++ b/io/utilities/parseSBMLNotesField.m
@@ -7,6 +7,19 @@ function [genes,rule,subSystem,grRule,formula,confidenceScore,citation,comment,e
 % Markus Herrgard 8/7/06
 % Ines Thiele 1/27/10 Added new fields
 % Handle different notes fields
+% Thomas Pfau 1/10/17 Make distinction between Matlab versions
+
+MatlabVer = version('-release');
+[A,B] = regexp(MatlabVer,'[\d]+');
+MatlabYear = str2num(MatlabVer(A:B));
+%if we are prior to 2013 use the old version
+if MatlabYear < 2013
+    [genes,rule,subSystem,grRule,formula,confidenceScore,citation,comment,ecNumber,charge] = parseSBMLNotesField2012(notesField)
+    return
+end
+
+
+
 if isempty(regexp(notesField,'html:p', 'once'))
     tag = 'p';
 else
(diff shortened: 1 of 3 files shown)


laurentheirendt:
Hi @tpfau,

I cherry-picked the commits of your PR into a new branch in my fork: https://github.com/laurentheirendt/cobratoolbox/tree/pr-203

The tests in the folder testSBML however do not pass. May you have a look? I gave you temporary write-access on the branch `pr-203`.

Please let me know if you need my help. Thanks.

tpfau:
Hi @laurentheirendt,
There are a couple of issues (and my changes have little to do with them).
1. the testReadSBML function sets the solver to tomlab_cplex, so if that is not available the processing fails automatically. I'm not sure however, why there has to be a solver in the readSBML Model test (which should test, whether SBML are read properly and not whether the optimizers work). 
2. In the recent changes the matlab binaries for libsbml were removed. This makes use of TranslateSBML impossible (at least on my system). If I use the respective folder from an old toolbox it works fine again.

tpfau:
Hi @laurentheirendt 

Ignore my second comment (I only now noticed the change in submodules, which I did not expect). 
However, for the first, if I exchange the solver the test works fine. 
One thing though: If you run the build in an automated fashion, the SBML binaries will likely not get installed (from what I understood submodules init cannot be called automatically in a git clone as it would allow arbitrary code to be executed on a pulling machine without the user intending this.)

I added a small fix for part 1 of my response (by simply selecting the solver that was selected in the initCobraToolbox method), so this should now work on any system (since at least glpk is available).

tpfau:
Hi again,
Ok, I was working on the wrong branch. I'm checking it currently.

tpfau:
Ok, I found the prblem. I mixed up a string that I tested for. It passes on my machine now (with the branch on your repo). 
I also added the retrieveModels function in the test (and a change back to the test folder (maybe this could be made part of the retrieveModels script so that you can retrieve the models and still stay in your folder).
Could you give it a try if it works for you as well?

Best,
Thomas

laurentheirendt:
Closing in favour of PR #271 .

PR #204, opened 2017-01-11
Added autocompletion for filenames to a couple of COBRA read functions

tpfau:
the tabcomplete function from the Matlab FileExchange allows the addition of tab-completion options for matlab functions (which are stored in the toolbox/local/ folder of the MATLAB root directory).
This is a convenience update that adds tab-completion to a couple of read functions in the toolbox making it more convenient to read model files (especially as some have rather odd names).


laurentheirendt:
@tpfau, may I ask why you closed this PR? It looked good 😄  We would simply have to change slightly the structure.

tpfau:
I was working on it again, and there is a change in   Matlab R2016+ which makes this much easier.
In particular, this version depends on the user having writing rights on the matlab source directory (which is not a given), and would fail on newer systems which use the novel methodology. 
I'll reopen it as soon as I have the newer update set up (and I would suggest, that this stays a feature for Matlab 2016+)

laurentheirendt:
Nice 👍  - thanks for the explanation!

👀 forward to the new PR!

tpfau:
Ok,
I added functionSignatures to a few relevant functions (There are plenty left but I did not want to go through all the code. 
In general, it seems that with Matlab 2016, a function signature feature was added, that might at some point be used more extensively within the framework, and which might be worth adapting for future functionality (and at some point also updating the existing code accordingly)

rmtfleming:
@laurentheirendt  status?

laurentheirendt:
I am on it.

laurentheirendt:
@tpfau, I cherry-picked your commits to the branch `pr-204` in my fork (see [here](https://github.com/laurentheirendt/cobratoolbox/tree/pr-204)). 🎉 

I tried using the `tabcomplete` feature, but on a `UNIX` system, I get the following error message:
```
>> tabcomplete
Error using tabcomplete>getCurrentXmlDefs (line 176)
/usr/local/MATLAB/R2016b/toolbox/local/TC.xml was not found - cannot set tab-completions

Error in tabcomplete (line 106)
    [defs,tcXmlFilename] = getCurrentXmlDefs;
```
I guess that this has to do with the user having the right to write in the `MATLAB` source directory. Would it be possible to install this locally only, e.g. `~/`? May you have a look at this? Feel free to push to `pr-204` on my fork.

tpfau:
Hi laurent,
tabcomplete (as a tool) won't work with Matlab > 2016, but they introduced the functionSignatures.json info files in 2016, so you simply put a json file in the folder and get your tab-completion for files from there if the signature indicates filepath. 
Thats why the change to initCobra checks the matlab version and ignores the tabcomplete stuff if the version is > 2015 (I just noticed, that I did a mistake there and it should be <= 2015 not < ...

laurentheirendt:
@tpfau, I keep getting these errors:
```
>> submitFastGapFill(Error parsing JSON data; Boost reports "<unspecified file>(20): expected ',' or ']'"
```
Do you know why? This is thrown in `R2016b` on a `UNIX` machine.

tpfau:
Found it (I think) 
I forgot a comma in the json file, i.e. the list syntax was violated. Should be fixed now.

laurentheirendt:
Others were also throwing similar errors:
````
>> xls2model(Error parsing JSON data; Boost reports "<unspecified file>(11): expected ',' or '}'"
````

**Comment by tpfau** [comment]
Should be fixed. 
Sry for that

File: src/io/functionSignatures.json [hunk 1]
```diff
@@ -0,0 +1,20 @@
+{
+"readCbModel":
```
laurentheirendt: @tpfau , is there a reason for this duplicate `readCbModel` entry?

tpfau: From what I understood, a function can have multiple signatures, which could (in future) releases be used by Matlab to e.g. display different input argument combinations. 
While there is currently no reason for it, I thought to include it for potential future releases.
Admittedly, since functionSignatures is an undocumented feature, I'm not sure what mathworks will do with it in the future. But if you look at $MATLABROOT/toolbox/shared/io/functionSignatures.json there is also a duplicate function signature for detectImportOptions.
As I said, I'm not sure, what mathworks is going to do with it, but I doubt it will hurt.

laurentheirendt: ok - let's keep that one then for the future.

laurentheirendt:
@tpfau, I have `tabcomplete` working for `R2016+`, but, as you mentioned, there might be some issues for `R2015` and `R2014`. Shall we have this feature only for `R2016`?

tpfau:
For 2014/2015 (or earlier), there should not be an issue if the user has write permissions on their matlab folder. If they don't the autocompletion simply wont work, but it won't be causing any other issues.
Have a look at the modified initCobraToolbox. I'm checking whether the version is prior to 2016 and if, the tabcoompletion tool modifies the relevant File (if write access is possible), asking the user what he/she wants. If the user either doesn't want the tab-completion or doesn't have access, nothing happens for versions prior to 2016.

initCobraToolbox.m
```diff
@@ -171,7 +171,7 @@ function initCobraToolbox()
 %If the Matlab version is prior to Matlab 2016a, we have to use the TC.xml
 %structure to provide autocompletion, otherwise we can use the json
 %definition.
-if MatlabYear < 2015
+if MatlabYear < 2016
```
laurentheirendt: Shall we not write `> 2015` for subsequent versions?

tpfau: No.
This piece of code is only for earlier Matlab versions. For anything newer than 2016 it is completely useless.
Let me elaborate:
Before 2016: autocompletion was handled by the TC.xml file, which was loaded at Matlab startup. tabcomplete is a tool from fileexchange, that provides tools to modify this file and adjust it according to requests.
2016 and later(probably): The TC.xml file was removed. tab-completion is handled by functionSignatures.json files located in the folders of the respective files. 

Thus for 2016 or newer, this piece of code should not be called and would actually cause an error, as the file it looks for no longer exists. However for Matlab prior to 2016, the functionSignatures.json files are ignored, as they are a new addition, and therefore, to have tab completion for matlab versions before 2016, we need this bit of code.

laurentheirendt: Sorry, I got that wrong. I wanted to say `> 2015` to inform the user that `tabcomplete` works. 😄

tpfau: We could do an "else" statement after the setup for R2015 and below, that informs that autocomplete was added. BUt I don't think its necessary. its more an "expected behaviour" (my opinion)

laurentheirendt: I can't get `tabcomplete` to work on `R2014b` and `R2015b`. I am in favor of supporting `tabcomplete` only for `R2016+` and leaving the condition with `TC.xml` completely out. Do you agree?

tpfau: Whats the error you get? 
If its an error during the checks, than its because the function needs user input. 
In general, I don't mind not providing the tab.complete option for earlier versions, but I would prefer if it would be available.

laurentheirendt: On `R2014b` and `R2015b`, it just doesn't do anything (as if `tabcomplete` wasn't present).

Also, on `macOS`, same error:
````
Error parsing JSON data; Boost reports "<unspecified file>(11): expected ',' or '}'"
````

laurentheirendt:
Closing in favor of PR #316.

PR #209, opened 2017-01-18
Changes to Testing Suite

laurentheirendt:
Changes are as discussed, but mainly:

- Windows compatibility (`retrieveModels.m`)
- Removing `.sh` file to retrieve models
- Updated `testReadSBML` and `testWriteSBML` to increase coverage
- Minor changes to `.gitignore` and the `launchTests` script

_NB: In the future, the root folder of the CT might be global variable or being retrieved via a function._


test/models/retrieveModels.m
```diff
@@ -0,0 +1,47 @@
...
+modelArr = {
+    'iIT341.xml', 'http://bigg.ucsd.edu/static/models/iIT341.xml';
+    'Abiotrophia_defectiva_ATCC_49176.xml', 'https://webdav-r3lab.uni.lu/public/msp/AGORA/sbml/Abiotrophia_defectiva_ATCC_49176.xml';
+    'Ec_iJR904.xml', 'http://systemsbiology.ucsd.edu/sites/default/files/Attachments/Images/InSilicoOrganisms/Ecoli/Ecoli_SBML/Ec_iJR904.xml';
+    'Sc_iND750_flux1.xml', 'http://gcrg.ucsd.edu/sites/default/files/Attachments/Images/InSilicoOrganisms/yeast/Sc_iND750_flux1.xml';
+    };
+
+for i = 1:length(modelArr)
+    if exist(modelArr{i,1}, 'file') ~= 2
+        urlwrite(modelArr{i,2}, modelArr{i, 1});
```
syarra: Here you check the existence of the file. How do you ensure that the remote model hasn't been updated?

laurentheirendt: There is no `sync` for now. These models are published, and should not change over time. I however see 2 options:
- remove the files and download them again (takes time)
- have a cron on the CI that removes the file periodically

What do you think?

syarra: For now, we will not have such features but let's keep in mind that there is a tiny probability that these models may change in the future.

PR #237, opened 2017-02-02
Turn off patch report from codecov

laurentheirendt:
As discussed with @stevepeak. Thank you so much!


laurentheirendt:
Nice, it works - thanks @stevepeak!

Issue #21, opened 2014-05-19
Faulty metabolite renaming in checkCobraModelUnique


andrevo:
I am having issues running checkCobraModelUnique on my model - 'cell content indices must be greater than 0' at line 51, when renaming metabolites. By the looks of it, this can be traced back to line 49 - the function is looking through the model's reactions for a given metabolite, which doesn't seem right. Then, in line 51, it's trying to rename a reaction, while it should be trying to rename a metabolite.

rmtfleming:
If you're sure it is a problem with the code and not your model, please submit a pull request with your fix.

syarra:
@andrevo, any news?

andrevo:
I'm not 100% sure which model I was using when I experienced the issue, 
and if I even still have it on my computer, so I'm not sure if I can try 
to replicate the original problem. I'll try to find out and get back to you.

André

PR #247, opened 2017-02-07
Update checkCobraModelUnique.m

stefaniamagg:
Referring to issue #21 . Repace use of findRxnIDs and findMetIDs. Both functions return the first index of rxn/met found in model even if multiple entries are found in the model. Add test for checkCobraModelUnique.m.


syarra:
Thanks @stefaniamagg for changing this!

PR #248, opened 2017-02-07
Test changeRxnBounds.m

stefaniamagg:
Test function to check changeRxnBounds. Explores different input types, i.e.., multiple reactions, variable boundary types, and variable bound values.


syarra:
Thanks @stefaniamagg. Good job!

PR #249, opened 2017-02-07
Add test function for changeObjective.m

stefaniamagg:
Test function for changeObjective.m. Tests for default coefficient on single and multiple reactions, tests manually set coefficients on multiple rxns. Checks error when input rxn is not in model.


syarra:
Thanks @stefaniamagg for contributing!

PR #295, opened 2017-02-28
Fixed showprogress

laurentheirendt:
As discussed.


syarra:
Nice catch!
Thanks @laurentheirendt for fixing this!

PR #306, opened 2017-03-07
Removing redundant definitions of parseBoolean

laurentheirendt:
- Same function `parseBoolean.m` exists in `/src/tools/` and has been extended (see [here](https://github.com/opencobra/cobratoolbox/blob/master/src/tools/parseBoolean.m)). The version in `/src/tools/` contains the removed definitions.

- Test already exists.


syarra:
Nice catch! Let's chase all redundant files to get rid of them!

PR #309, opened 2017-03-09
Improved SBML reading (PR 181)

laurentheirendt:
Continuation of PR #181 .


src/io/utilities/parseSBMLNotesField2012.m
```diff
@@ -63,15 +63,15 @@
...
+        charge = str2double(regexprep(strrep(fieldStr,'CHARGE:',''),'^(\s)+',''));
     elseif (regexp(fieldStr,'AUTHORS'))
         if isempty(citation)
             citation = strcat(regexprep(strrep(fieldStr,'AUTHORS:',''),'^(\s)+',''));
         else
             citation = strcat(citation,';',regexprep(strrep(fieldStr,'AUTHORS:',''),'^(\s)+',''));
         end
     elseif (regexp(fieldStr,'^Confidence Level'))
-        confidenceScore = regexprep(strrep(fieldStr,'Confidence Level:',''),'^(\s)+','');
+        [matches, tmpTokens] = regexpi(fieldStr, 'Confidence[ _]Level: (\w+)', 'match', 'tokens');
```
akaviaLab: There's a few lines missing here, so matches and tokens aren't used.
    if (~isempty(matches))
        confidenceScore = str2double(tmpTokens{1}{1});
    end

laurentheirendt: @tpfau , may you look into this? Thanks.

tpfau: The comment is correct, but...
Is the confidenceScore a String value or a Double value? Up until now, it was present as a double in the models. 
The toolbox currently assumes, that this is a cell array (see e.g. addReaction, where a new reaction appends to the confidence score as a cell array). 
So for now, I would add :
if (~isempty(matches))
confidenceScore = tmpTokens{1}{1};
end
instead of the str2double version. But this should also go into our discussion about model fields (and again this is almost unused in any other toolbox function, so changing it should not be a big issue).

laurentheirendt: @akaviaLab , @rmtfleming: do you agree?

akaviaLab: Okay, if addReaction assumes it is a cell array, keep it as you suggested.
I would say this assumption doesn't make a lot of sense, and we should probably correct it when fixing model fields, but that's a future discussion.

laurentheirendt: @tpfau, may you make the change that you suggested? Thanks.

tpfau: Done

laurentheirendt: Thanks @tpfau!

PR #312, opened 2017-03-13
Improved robustness of sampling preprocessing (PR 310)

laurentheirendt:
Continuation of #310 .


external/mve11/mve_run.m
```diff
@@ -1,286 +0,0 @@
-function [x,E] = mve_run(A,b,x0)
```
laurentheirendt: @Bounciness

Bounciness: I was worried that it would get called by accident. The original zip file has two files with the name mve_run.m, one with functions in different files and one with all functions in 1 file. I updated the version with functions split across files, and deleted this one so that it would not get called by accident.

Bounciness: There was no deeper reason in deleting it, and it could be good to keep the original version of the code there. Maybe we could modify the file name to be mve_run_original or something?

laurentheirendt: @Bounciness, may you post the file here? I will then add it. Thanks.

laurentheirendt: @Bounciness, may you post the file here? I will then add it. Thanks.

Bounciness: @laurentheirendt I have attached the file, to be placed in external/mve11 (it did not let me post the .m file directly, so I have attached a zip file)
[mve.zip](https://github.com/opencobra/cobratoolbox/files/839103/mve.zip)


Although, would it be preferable to leave the external folder as is (i.e. as downloaded from their website), and place our modified code inside of the sampling folder? Currently, I just directly modified the functions inside external/mve11.

laurentheirendt: @Bounciness , yes, that would be much better. In `external/`, we have code that we do not maintain, but provide for convenience.

May you make the changes? I gave you write access on my fork. Simply push to the branch `pr-310` and it will show up here in the PR. The commands are:
```
git clone https://github.com/laurentheirendt/cobratoolbox.git
git checkout pr-310
# do your changes
git add [yourFiles]
git commit -m "[yourMessage]"
git push origin pr-310
```

Bounciness: @laurentheirendt OK, I've pushed the changes to the pr-310 branch on your fork. Thanks a lot for your help.

Bounciness: There are two commits on your fork because I realized I forgot something right after the first commit. The second one should be the one to use.

laurentheirendt:
Thanks @Bounciness .

PR #316, opened 2017-03-14
Added autocompletion for filenames (PR 204)

laurentheirendt:
Continuation of PR #204 .

**Notes**
- Support for `R2016b+` only.
- Tested on `R2016b` and `R2017a` on `Linux`, `macOS`, and `Windows`


src/io/utilities/functionSignatures.json
```diff
@@ -0,0 +1,18 @@
+{
```
syarra: I'm not sure I understand correctly this functionality. From the documentation of `tabComplete` I understand that it will generate the signature of the function automatically. Is this correct?
If so, I am wondering if it would be possible to put all the `json` files in a common directory. This is a little bit annoying to have all these `json` file in the `src` subfolders.

laurentheirendt: Actually, it would be nice to have all `json` files in a common folder. @tpfau, what do you think?

tpfau: @syarra the tabcomplete function from fileexchange has nothing to do with the json files.
tabcomplete is a tool for matlab prior to 2016a which adds function signatures (and thus autocompletion) to the TC.xml file which is loaded at startup by matlab and used to assign function signatures.

@laurentheirendt, @syarra : 
The json functionSignature mechanism is a Matlab-built-in mechanism. And that mechanism relies on the functionSignaturs.json being located in the folder of the function it provides a signature for. This is actually makes sense to avoid overloading signatures (i.e. using a wrong signature because two functions have the same name, which can easily happen in the old style where everything was saved in the TC.xml)
Thus, no, they unfortunately can't be moved to a common folder.
In fact, as discussed with @laurentheirendt I would either remove the whole tabcomplete fileexchange stuff from the pull again (only keeping the jsons), or, if tabcomplete stays in, provide a utilityfunction that essentially offers to add the tabcomplete functionality for Matlab versions prior to 2016a. But if we don't decide to provide this utility function, the tabcomplete folder can be completely removed.

laurentheirendt: @tpfau , are these `.json` files generated manually? If so, they will be challenging to maintain ...

tpfau: Yes, they are created manually.
True, maintaining them is a challenge, if the function signatures change (which should only happen in major revisions, where I would claim, that this effort is worthwile.

laurentheirendt:
As discussed with @tpfau and @syarra , we will freeze this PR for the following reasons (for now):
- too challenging to maintain
- very limited compatibility (only `R2016+`)

PR #318, opened 2017-03-15
ctype field needs to be a row vector in cplex.

tpfau:

- [X] Followed the guidelines in the [contributing](https://github.com/opencobra/cobratoolbox/blob/master/.github/CONTRIBUTING.md) document
- [X] Followed the [code styleguide](https://github.com/opencobra/cobratoolbox/blob/master/.github/CONTRIBUTING.md#code-styleguide)
- [X] Checked to ensure there aren't other open [Pull Requests](https://github.com/opencobra/cobratoolbox/pulls) for the same update/change
- [X] Written platform-independent code
- [X] Used `pwd` to get the current directory
- [X] Used `filesep` for paths (e.g., `['myPath' filesep 'myFile.m']`)

### Errors/fixes (with reference to eventual open issues)

e.g. OptKnock produced a row vector for the vartype, which was previously transposed in the cplex setup. The current code ensures, that transposition only happens if the ctype vector would otherwise have the wrong orientation.

###  Tests tailored to test the PR

No specific test for this PR.

### DIFF
--- a/src/solvers/solveCobraMILP.m
+++ b/src/solvers/solveCobraMILP.m
@@ -380,7 +380,7 @@ switch solver
         if size(vartype,1) > size(vartype,2)
             vartype = vartype';
         end
-        cplexlp.Model.ctype = vartype';
+        cplexlp.Model.ctype = vartype;
         cplexlp.Start.x = x0;
         cplexlp.Param.mip.tolerances.mipgap.Cur =  solverParams.relMipGapTol;
         cplexlp.Param.mip.tolerances.integrality.Cur =  solverParams.intTol;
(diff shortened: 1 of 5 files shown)


laurentheirendt:
I will add a quick test.

tpfau:
The problem is: you can't test it on 2016b, since Matlab segfaults with ibm_cplex on 2016b, so an explicit test will always fail on 2016b...

laurentheirendt:
Let's see ...

laurentheirendt:
Note: the `ibm_cplex` interface is tested for `R2014b` and `R2015b` (but not for `R2016b`).

syarra:
Good job @tpfau and @laurentheirendt! :1st_place_medal:

PR #348, opened 2017-03-28
:bug: Fixing problem with textprogressbar

tpfau:

- [X] Followed the guidelines in the [contributing](https://github.com/opencobra/cobratoolbox/blob/master/.github/CONTRIBUTING.md) document
- [X] Followed the [code styleguide](https://github.com/opencobra/cobratoolbox/blob/master/.github/CONTRIBUTING.md#code-styleguide)
- [X] Checked to ensure there aren't other open [Pull Requests](https://github.com/opencobra/cobratoolbox/pulls) for the same update/change
- [X] Written platform-independent code
- [X] Used `pwd` to get the current directory
- [X] Used `filesep` for paths (e.g., `['myPath' filesep 'myFile.m']`)

### Errors/fixes (with reference to eventual open issues)
Text injected during a textprogressbar was active lead to a very odd output.
Added an option to textprogressbar that handles output while a textprogressbar is in process, which can be used in the future.
#### Previous output
....
94%     [.....................................   ]Perform model consistency check
reduceModel.m: Model is consistent
Reduced model: 87 rxns 68 metabolites
Create warmup points
Creating warmup points ...
100%    [........................................]
Run sampler for a total of 2000000 steps
ACHR sampling in progress ...
File #	Point #	Step #	Time	#Time left
1	1	0	    0.00	     NaN
0%      [                   0%      [                    0%      [                    1%      [                     1%      [                     1%      [                     1%      [                     1%      [                                        ]
....

#### Output of PR-merged version

94%     [.....................................   ]Perform model consistency check
reduceModel.m: Model is consistent
Reduced model: 87 rxns 68 metabolites
Create warmup points
Creating warmup points ...
100%    [........................................]
Run sampler for a total of 200000 steps
ACHR sampling in progress ...
File #	Point #	Step #	Time	#Time left
1	1	0	    0.00	     NaN
...
1	901	180000	    1.22	    0.14
1	926	185000	    1.25	    0.10
1	951	190000	    1.29	    0.07
1	976	195000	    1.32	    0.03
100%    [........................................]

###  Documentation updates

[*Include here information on how the documentation has been updated*]

- [X] Documented your code based on the [Documentation styleguide](https://github.com/opencobra/cobratoolbox/blob/master/.github/CONTRIBUTING.md#documentation-and-comments-styleguide).
- [X] Documented the `Input` and `Output` arguments
- [X] Followed the guidelines to automatically generate the documentation

###  Tests tailored to test the PR

None, should be tested during the sampler Testing. 

- [X] Tested your PR locally prior to submission
- [X] Checked that all tests pass
- [X] Tested the code on Linux

### DIFF
--- a/external/textprogressbar/textprogressbar.m
+++ b/external/textprogressbar/textprogressbar.m
@@ -22,7 +22,6 @@ function textprogressbar(c)
 
 %% Initialization
 global WAITBAR_HANDLE;           %   Carriage return pesistent variable
-persistent WAITBAR_PROGRESS;
 persistent WAITBAR_STRING;
 % Vizualization parameters
 strPercentageLength = 10;   %   Length of percentage string (must be >5)
@@ -39,6 +38,7 @@ elseif isempty(WAITBAR_HANDLE) && ischar(c)
 elseif ~isempty(WAITBAR_HANDLE) && ischar(c)
     % Progress bar  - termination
     WAITBAR_HANDLE = [];
+    WAITBAR_STRING = [];
     fprintf([c '\n']);
 elseif isnumeric(c)
     % Progress bar - normal progress
--- a/external/textprogressbar/textprogressbar.m
+++ b/external/textprogressbar/textprogressbar.m
@@ -67,6 +67,8 @@ elseif isnumeric(c)
 elseif islogical(c)
     if c %remove the current progress
         fprintf(WAITBAR_HANDLE);
+        fprintf(repmat(' ',1,numel(WAITBAR_STRING)-1));
+        fprintf(WAITBAR_HANDLE);
     else
         fprintf(WAITBAR_STRING);
     end
--- a/external/textprogressbar/textprogressbar.m
+++ b/external/textprogressbar/textprogressbar.m
@@ -17,16 +17,18 @@ function textprogressbar(c)
 %
 %      - renamed strCR to WAITBAR_HANDLE
 %      - changed persistent type of strCR to global type
+%      - Changes To the progressbar text to be consistent (to allow
+%        intermittent output)
 
 %% Initialization
 global WAITBAR_HANDLE;           %   Carriage return pesistent variable
-
+persistent WAITBAR_PROGRESS;
+persistent WAITBAR_STRING;
 % Vizualization parameters
 strPercentageLength = 10;   %   Length of percentage string (must be >5)
 strDotsMaximum      = 40;   %   The total number of dots in a progress bar
 
 %% Main
-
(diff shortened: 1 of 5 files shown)


external/textprogressbar/textprogressbar.m
```diff
@@ -17,16 +17,18 @@ function textprogressbar(c)
 %
 %      - renamed strCR to WAITBAR_HANDLE
 %      - changed persistent type of strCR to global type
+%      - Changes To the progressbar text to be consistent (to allow
+%        intermittent output)
 
 %% Initialization
 global WAITBAR_HANDLE;           %   Carriage return pesistent variable
-
+persistent WAITBAR_PROGRESS;
```
laurentheirendt: Why do you declare these `persistent` variables?

external/textprogressbar/textprogressbar.m
```diff
@@ -40,29 +42,37 @@ function textprogressbar(c)
...
 
     % Update carriage return
-    WAITBAR_HANDLE = repmat('\b',1,length(strOut)-1);
+    WAITBAR_HANDLE = repmat('\b',1,length(WAITBAR_STRING)-1);
 
-    if c == 100
+    if WAITBAR_PROGRESS == 100
         fprintf('\n');
     end
+elseif islogical(c)
```
laurentheirendt: I am wondering if we shall not move `textprogressbar.m` to `tools/` - our version is highly customized already. @syarra ?

external/textprogressbar/textprogressbar.m
tpfau: WAITBAR_PROGRESS, doesn't necessarily need to be set persistent, but either this, or the WAITBAR_STRING needs to be persistent, to be able to recover the old status of the progress bar, when removing/adding text.
I initially intended to recalculate the Str based on waitbar progress, but noticed, that just using the string is actually more efficient (and over that forgot to revert the WAITBAR_PROGRESS variable).

laurentheirendt: I am just wondering if `persistent` is enough, or if we should have `global` ...

tpfau: Persistent is entirely enough. This is a field, that is only ever called in this function, and therefore global would be unnecessary. From my understanding, persistent variables can be considered as private fields of a function handle (i.e. the function is treated as an object and the persistent variable is a field of the function object). The assumption in this context is, that there is only ever one textprogressbar at the same time (we might actually consider to change this to an object at some point), and as such that object uses the given String.

laurentheirendt:
Closing PR temporarily.

PR #351, opened 2017-03-28
Fix for code grade - omit gitignore files

laurentheirendt:
As discussed.


test/testAll.m
```diff
@@ -46,6 +46,15 @@
 profile on;
 
 if ~isempty(strfind(getenv('HOME'), 'jenkins'))
+
+    % ignore list of files
+    ignoreFiles = {'./src/fluxomics/c13solver/IsotopomerModel.txt',
```
syarra: Could you read from the `.gitignore` file instead of creating this list of ignored files?

laurentheirendt: This overcomplicates ... I suggest to leave that for now.

syarra: As you wish!

PR #371, opened 2017-04-04
New init script and new changeCobraSolver

laurentheirendt:
**Major changes**
- new `initCobraToolbox` script
- new `changeCobraSolver`
- updated `SOLVER.md`
- simplified tests


.github/SOLVERS.md
```diff
@@ -1,107 +1,192 @@
...
     ````
-
 
 ## IBM ILOG CPLEX
 
 1) Download the `CPLEX` installation binary from [here](https://www-01.ibm.com/software/websphere/products/optimization/cplex-studio-community-edition/)
 
-2) Add `execute` permission to binary
-
+2) <img src="https://prince.lcsb.uni.lu/jenkins/userContent/apple.png" height="20px" alt="macOS">  <img src="https://prince.lcsb.uni.lu/jenkins/userContent/linux.png" height="20px" alt="Linux"> Add `execute` permission to binary
```
tpfau: Why only the community version of cplex, which will be useless for most medium size models (max 1000 Variables)?
Maybe a hint to the IBM Academic initiative might be useful)

src/solvers/changeCobraSolver.m
```diff
@@ -1,365 +1,385 @@
...
+end
+
+if nargin < 3
+    printLevel = 1;
+end
+
+% Attempt to set the user provided solver for all optimization problem types
+if (strcmp(solverType, 'ALL'))
+    for i = 1:length(OPT_PROB_TYPES)
+        changeCobraSolver(solverName, OPT_PROB_TYPES{i}, 0);
```
tpfau: This surpresses the output, shouldn't this be called with the actual provided printlevel?
Also, this doesn't give a return value (i.e. this will likely fail), is this intentional?

src/solvers/changeCobraSolver.m
```diff
@@ -1,365 +1,385 @@
...
+% Output:
+%     solverOK: true if executableName exists, false otherwise.
+%
+    solverOK = false;
+    if unix
+        if ~isunix && printLevel > 0
+            error('%s interface not implemented for non unix OS', solverName);
+        end
+    end
+    [status, cmdout] = system(['which ' executableName]);
```
tpfau: which a a unix specific command for all I know. Is this intentionally only working on a unix system?

src/solvers/changeCobraSolverParams.m
```diff
@@ -56,9 +56,9 @@
 allowedQPparams = {'minNorm', 'printLevel'};
 allowedMILPparams = {'intTol','relMipGapTol','timeLimit','logFile','printLevel'};
 
-% Only LP is currently included
+% Only LP, QP and MILP are currently included
```
tpfau: Just as a general question,
Wouldn't it make sense to change these parameters from global variables to persistent variables in this function? 
This would force usage of this function for parameter changes, and thus avoid tracking issues.
Alternatively an actual config file could be used, which could actually be accessed by paralell workers (but that would make it much slower probably)

src/solvers/configEnvVars.m
```diff
@@ -0,0 +1,56 @@
...
+        solverPaths{1,2} = {'/Applications/IBM/ILOG/CPLEX_Studio1262', '/Applications/IBM/ILOG/CPLEX_Studio1263', '/Applications/IBM/ILOG/CPLEX_Studio127', ...
+                            '/opt/ibm/ILOG/CPLEX_Studio1262', '/opt/ibm/ILOG/CPLEX_Studio1263', '/opt/ibm/ILOG/CPLEX_Studio127', ...
+                            'C:\Program Files\IBM\ILOG\CPLEX_Studio1262', 'C:\Program Files\IBM\ILOG\CPLEX_Studio1263', 'C:\Program Files\IBM\ILOG\CPLEX_Studio127'};
+        solverPaths{2,1} = 'GUROBI_PATH';
+        solverPaths{2,2} = {'/Library/gurobi600', '/Library/gurobi650', '/Library/gurobi702', '/opt/gurobi650', '/opt/gurobi70', 'C:\gurobi600', 'C:\gurobi650', 'C:\gurobi70'};
+        solverPaths{3,1} = 'TOMLAB_PATH';
+        solverPaths{3,2} = {'/opt/tomlab', 'C:\tomlab', '/Applications/tomlab'};
+        solverPaths{4,1} = 'MOSEK_PATH';
+        solverPaths{4,2} = {'/opt/mosek/7/', '/opt/mosek/8/', '/Applications/mosek/7', '/Applications/mosek/8', 'C:\Program Files\Mosek\7', 'C:\Program Files\Mosek\8'};
+
```
tpfau: Just out of curiosity.
If we use multiple options for locations of specific solvers (and e.g. Cplex has many more than those mentioned here), wouldn't it be an idea to use the OS to look for those files? 
E.g. 
if the ILOG_CPLEX_PATH environment variable is empty, we could use "locate cplexqp.m" on a unix system or "where cplexqp.m"  on a windows machine to determine the path (if any) and even give an option to the user to select the path.
just a thought.

test/models/retrieveModels.m
```diff
@@ -21,35 +38,55 @@
...
+        end
     end
 end
 
-% Ec_iAF1260_flux1.xml
+% download Ec_iAF1260_flux1.xml
 if exist('Ec_iAF1260_flux1.xml', 'file') ~= 2
     urlwrite('http://systemsbiology.ucsd.edu/sites/default/files/Attachments/Images/InSilicoOrganisms/Ecoli/Ecoli_SBML/msb4100155-s6.zip', 'msb4100155-s6.zip');
     system('unzip -qq msb4100155-s6.zip');
     system('rm Ec_iAF1260_flux2.txt');
```
tpfau: rm and mv along with unzip are all unix commands. This will likely Fail on a Windows system.

test/verifiedTests/testCardOpt/testSparseLP.m
```diff
@@ -6,18 +6,13 @@
...
-% define global paths
-global path_TOMLAB
-global path_GUROBI
 
 % save the current path
 currentDir = pwd;
 
 % initialize the test
-initTest(fileparts(which(mfilename)));
+fileDir = fileparts(which('testSparseLP'));
```
tpfau: wouldnt mfilename instead of the actual fileName make sense here (avoids problems if the file ever gets renamed)

test/verifiedTests/testCorrespondingRowsCols/testCorrespondingRowsCols.m
```diff
@@ -11,7 +11,8 @@
 currentDir = pwd;
 
 % initialize the test
-initTest(fileparts(which(mfilename)));
+fileDir = fileparts(which('testCorrespondingRowsCols'));
+cd(fileDir);
```
tpfau: same as above

test/verifiedTests/testDeletionStudy/testDeletionStudy.m
```diff
@@ -1,27 +1,30 @@
...
-global path_GUROBI
+global CBTDIR
 
 % save the current path
 currentDir = pwd;
 
 % initialize the test
-initTest(fileparts(which(mfilename)));
+fileDir = fileparts(which('testDeletionStudy'));
+cd(fileDir);
```
tpfau: see above

test/verifiedTests/testDynamicFBA/testDynamicFBA.m
```diff
@@ -6,16 +6,16 @@
...
-global path_TOMLAB
+global CBTDIR
 
 % save the current path
 currentDir = pwd;
 
 % initialize the test
-initTest(fileparts(which(mfilename)));
+fileDir = fileparts(which('testDynamicFBA'));
+cd(fileDir);
```
tpfau: see above

test/verifiedTests/testElementalBalance/testElementalBalance.m
```diff
@@ -10,7 +10,8 @@
 currentDir = pwd;
 
 % initialize the test
-initTest(fileparts(which(mfilename)));
+fileDir = fileparts(which('testElementalBalance'));
+cd(fileDir);
```
tpfau: see above (and at a couple of additional places)

test/verifiedTests/testPrint/testPrintConstraints.m
```diff
@@ -24,8 +27,8 @@
 maxInf = +Inf;
 printConstraints(model, minInf, maxInf);
 diary off
-text1 = fileread('refData_printConstraints.txt');
-text2 = fileread('printConstraints.txt');
+text1 = readMixedData('refData_printConstraints.txt');
+text2 = readMixedData('printConstraints.txt');
 assert(isequal(text1, text2));
```
tpfau: why the change?

test/verifiedTests/testPrint/testPrintRxnFormula.m
```diff
@@ -37,8 +41,8 @@
 formulas = printRxnFormula(model, model.rxns, true, true, true, 1, true, true);
 diary off
 
-text1 = fileread('refData_printRxnFormulaGPR.txt');
-text2 = fileread('printRxnFormula.txt');
+text1 = importdata('refData_printRxnFormulaGPR.txt');
+text2 = importdata('printRxnFormula.txt');
```
tpfau: Why the changes from fileread to importdata? I know the main difference between them is that linebreaks get ignored but would this have any influence?

test/models/retrieveModels.m
laurentheirendt: This has been tested on Windows. These commands also work on Windows.

test/verifiedTests/testCardOpt/testSparseLP.m
laurentheirendt: The simple answer is no. 

`R2014b` and `R2015b` do not support this when the tests are run in `runtests()`. This is the only way this can work on `R2014b` and `R2015b` when the tests are called in `runtests()`. Reason is that `mfilename` is the to the path of `ScriptTestCaseProvider` in `runtests`.

test/models/retrieveModels.m
tpfau: Windows 7 32bit machine: 

> 'mv' is not recognized as an internal or external command operable program or batch file 

Thats what I get here.

test/verifiedTests/testCorrespondingRowsCols/testCorrespondingRowsCols.m
laurentheirendt: same comment as above

test/verifiedTests/testDeletionStudy/testDeletionStudy.m
laurentheirendt: same comment as above

test/verifiedTests/testDynamicFBA/testDynamicFBA.m
laurentheirendt: same comment as above

test/verifiedTests/testElementalBalance/testElementalBalance.m
laurentheirendt: same comment as above

test/verifiedTests/testPrint/testPrintConstraints.m
laurentheirendt: Compatibility on Windows ...

test/verifiedTests/testPrint/testPrintRxnFormula.m
laurentheirendt: Compatibility on Windows ...

test/verifiedTests/testCardOpt/testSparseLP.m
tpfau: ok. 
wasn't aware of that.

test/verifiedTests/testPrint/testPrintRxnFormula.m
tpfau: oh well ^^

test/models/retrieveModels.m
tpfau: Ok, unzip actually works.

src/solvers/configEnvVars.m
laurentheirendt: The thought is good - the implementation is too system dependent - maybe later.

The path is set to the `path` as a user would have set by following the instructions in `SOLVERS.md`. The user has to set the environment variable when installing the solver - this is purely for convenience if the default paths have been used in the past when installing a solver, but the environment variables have not been set (e.g. this is the case for existing users).

src/solvers/changeCobraSolverParams.m
laurentheirendt: `global` is preferred here to avoid a config file in a parallel setting.

tpfau: But global variables are not propagated to a parallel pool neither, so it doesn't really make any difference for parallel work.

src/solvers/changeCobraSolver.m
laurentheirendt: Yes, indeed, this should be within the `if` condition. The function `checkSolverInstallationExecutable` is only intended for unix-specific solver interfaces (at least for now). Good catch :+1:

test/models/retrieveModels.m
laurentheirendt: If the `Git Bash shell tools` are not installed, as explained in the prerequisites, these commands work (https://github.com/opencobra/cobratoolbox/blob/develop/.github/REQUIREMENTS.md). However, I will implement this to be fail safe.

test/verifiedTests/testCardOpt/testSparseLP.m
laurentheirendt: Me neither :smile: ...

src/solvers/changeCobraSolverParams.m
laurentheirendt: At least from my testing with `changeCobraSolver`, the global variable setting worked, but let me test this again.

.github/SOLVERS.md
laurentheirendt: On that website, there is a link to the community edition and the academic initiative. I agree though that this should be more visible. I will add a direct link to the academic initiative.

src/solvers/changeCobraSolver.m
tpfau: But doesn't this throw an error, if you run it with all and an output value?

And wouldn't it be sufficient to call it with a 0 printlevel in the init Script to avoid this?

.github/SOLVERS.md
laurentheirendt: Done.

src/solvers/changeCobraSolver.m
laurentheirendt: Thanks for your comment @tpfau

src/solvers/changeCobraSolver.m
laurentheirendt: Fixed. Thanks for your comment @tpfau

test/models/retrieveModels.m
laurentheirendt: Done.

test/verifiedTests/testCardOpt/testSparseLP.m
laurentheirendt: Done.

test/verifiedTests/testCorrespondingRowsCols/testCorrespondingRowsCols.m
laurentheirendt: Done.

test/verifiedTests/testDeletionStudy/testDeletionStudy.m
laurentheirendt: Done.

test/verifiedTests/testDynamicFBA/testDynamicFBA.m
laurentheirendt: Done.

test/verifiedTests/testElementalBalance/testElementalBalance.m
laurentheirendt: Done.

test/verifiedTests/testPrint/testPrintConstraints.m
laurentheirendt: Done.

test/verifiedTests/testPrint/testPrintRxnFormula.m
laurentheirendt: Done.

src/solvers/changeCobraSolver.m
laurentheirendt: Done.

src/solvers/changeCobraSolverParams.m
laurentheirendt: From my testing, the `global` variables work on each worker. For instance:
```Matlab
initCobraToolbox
global CBT_LP_SOLVER
parpool(2);
parfor i = 1:2
   CBT_LP_SOLVER
end
```
Note: The `global` variable can also be set after the `parpool` and is propagated to all parallel workers.

PR #375, opened 2017-04-07
Read objective reactions with symbols correctly and support >1 objective reactions

shjchan:
*(Note: You may replace [ ] with [X] to check the box)*

**I hereby confirm that I have:**

- [X] Tested my code on my own machine
- [X] Followed the guidelines in the [contributing](https://github.com/opencobra/cobratoolbox/blob/master/.github/CONTRIBUTING.md) document
- [X] Followed the [code styleguide](https://github.com/opencobra/cobratoolbox/blob/master/.github/CONTRIBUTING.md#code-styleguide)
- [X] Checked to ensure there aren't other open [Pull Requests](https://github.com/opencobra/cobratoolbox/pulls) for the same update/change
- [X] Written platform-independent code
- [X] Used `pwd` to get the current directory
- [X] Used `filesep` for paths (e.g., `['myPath' filesep 'myFile.m']`)
- [X] Made sure that computational results are reproducible

#### Previous output

[*Include here information of the current version of The COBRA Toolbox*]

#### Output of PR-merged version

[*Include here information of modified version of The COBRA Toolbox after your proposed PR is accepted*]

###  Documentation updates

[*Include here information on how the documentation has been updated*]

**I hereby confirm that I have:**

- [ ] Documented your code based on the [Documentation styleguide](https://github.com/opencobra/cobratoolbox/blob/master/.github/CONTRIBUTING.md#documentation-and-comments-styleguide).
- [ ] Documented the `Input` and `Output` arguments
- [ ] Followed the guidelines to automatically generate the documentation

###  Tests tailored to test the PR

[*Include here information on how new tests and/or how existing tests have been adapted*]
%test symbols in objective reactions
Test codes:
```
for jTest = 1:2
    if jTest == 1
        %test objective reactions with symbols
        model = createModel({'EX_a(e)';'EX_b(e)'},{'Test A';'Test B'},...
            {'a[e] <=>'; 'b[e] <=>'});
        model.c = [1;0];
    elseif jTest == 2
        model = createModel({'EX_a';'EX_b'},{'Test A';'Test B'},...
            {'a[e] <=>'; 'b[e] <=>'});
        %test more than one objective reactions
        model.c = [1; -2];
    end
    model.lb = model.lb(:);
    model.ub = model.ub(:);
    %add the fields outputted by readSBML
    [model.modelVersion.SBML_level,model.modelVersion.SBML_version,...
        model.modelVersion.fbc_version] = deal(3,1,2);
[...]

PR #378, opened 2017-04-13
Read objective reactions with symbols correctly and support >1 objective reactions (PR 375)

laurentheirendt:
Continuation of PR #375 (cherry-picked commits).

Test given in the description of PR #375 should be implemented in the test suite.

cc: @tpfau


src/io/utilities/readSBML.m
```diff
@@ -636,10 +640,10 @@
     model.lb = fbc_lb;
     model.ub = fbc_ub;
     if noObjective==0; % when there is an objective function
-        indexObj=findRxnIDs(model,fbc_obj);
+        indexObj=findRxnIDs(model,cleanUpFormatting(fbc_obj));
         % indexObj=find(strcmp(fbc_obj,model.rxns))
-        model.c(indexObj)=1;
-        model.osense = - sign(fbc_obj_value);
+        model.c(indexObj) = fbc_obj_value(:);
+        model.osense = fbc_obj_sense;
```
laurentheirendt: This needs to be changed back to ` model.osense = - sign(fbc_obj_value);` before accepting the PR.

src/io/utilities/readSBML.m
```diff
@@ -636,10 +640,10 @@
     model.lb = fbc_lb;
     model.ub = fbc_ub;
     if noObjective==0; % when there is an objective function
-        indexObj=findRxnIDs(model,fbc_obj);
+        indexObj=findRxnIDs(model,cleanUpFormatting(fbc_obj));
         % indexObj=find(strcmp(fbc_obj,model.rxns))
-        model.c(indexObj)=1;
-        model.osense = - sign(fbc_obj_value);
+        model.c(indexObj) = fbc_obj_value(:);
```
laurentheirendt: This must be changed back to `model.c(indexObj)=1;` before accepting the PR.

tpfau: I actually think in this instance, the fbc_obj_value(:) would be correct, as this would set it to the actual objective value. the current code assumes no objective coefficient.

src/io/utilities/readSBML.m
tpfau: agreed, but only for the osense line. the model.c line is correct (multiple reactions).

shjchan: Is osense supposed to be a single number '-1' for 'max' and '1' for 'min'? That is what I assumed from the solver functions, e.g. solveCobraLP.
This is implemented in the suggested update by defining the addition fbc_obj_sense.

In the current master version, 
```
fbc_obj_value = modelSBML.(fbc_list{f})(1).fbc_type(1:3);
...
model.osense = - sign(fbc_obj_value);
```
The fbc_obj_value is either 'min' or 'max'. "- sign(fbc_obj_value)" is always [-1 -1 -1] in either case. (I tried loading some models with fbc_version using R2015b. Or am I missing anything here?).

If using `model.osense = - sign(fbc_obj_value);`
with the updated definition of fbc_obj_value from line 306, osense will become a vector with 1 or -1 corresponding to each non-zero element in the objective vector.

tpfau: @shjchan you are right. I missed that.

shjchan: @laurentheirendt @tpfau 
Thank you. So can we agree on the changes? If so, are the following what I should do? 
(1) add the test file described in #375, and 
(2) create a new pull request (to the develop branch) cherry-picking the two commits here and the new commit adding the test file, with reference to these two pull requests #375 and #378.

Am I right? Thanks for helping.

tpfau: @shjchan,
I think it would be best if you add the test code to the existing testReadSBML (test/verifiedTest/testSBML/testReadSBML.m). but yes, this should be sufficient.

laurentheirendt: @shjchan, thanks for willing to help! :tada:

You can create a new PR, or you can push to the branch `pr-375` on my fork (I invited temporarily to collaborate). You simply have to do:
````sh
git clone https://github.com/laurentheirendt/cobratoolbox.git
git checkout pr-375
# do your changes
git add [yourFiles]
git commit -m "[yourMessage]"
git push origin pr-375
````

shjchan: @laurentheirendt, thanks Lauren. Have done that.

laurentheirendt:
Well done @shjchan and @tpfau!  🎉

PR #414, opened 2017-05-05
:bug: wrong argument provided to addReaction (was bool instead of empty)

tpfau:
The function provides a bool instead of an empty or cell array input. 


laurentheirendt:
Nice catch @tpfau! 🍰

PR #418, opened 2017-05-08
Changed to solveCobraLP to actually allow different solvers

tpfau:
Changed to solveCobraLP, the original code used a direct call to gurobi, which fails on any system that doesn't have gurobi installed, but goes unnoticed in the test systems (as they have gurobi).

However this necessitated removal of 2 output values (as they are gurobi specific outputs - but they are never used in the COBRAToolbox.)


laurentheirendt:
Nice @tpfau :cactus:

PR #419, opened 2017-05-09
IO Tutorial and IO Update

tpfau:
IOTutorial added and the IO is updated to be much more flexible when loading models.
Also added a checkModel() function - needs to be updated at some point to actually represent the required/optional fields, currently only a subset is used.

read and writeCbModel are adapted to accept parameter value pairs and to allow Excel and matlab IO (i.e. you can load models from mat files). I.e. all model IO can no go through these functions (no more xls2model etc).
Set a definition for Excel format models and their IO (could be updated in the future)
The latter functionality depends on the model struct and currently only those with required fields present will be loaded. This again depends on the definition of those fields and is currently saved in getDefinedFieldProperties (the function will have to be adapted to the new field definitions in the future).
Moved the mps io from writeCbModel to writeLPProblem (the old convertCobraLP2mps is changed to that function) and updated the respective tests. 
Also updated the tests for printRxnFormula/printFluxVector to reflect the change in printRxnFormula (irreversible reactions are no longer determined by "rev" but by lower bounds.


tpfau:
Closing in favor of

rmtfleming:
Please also include your tutorial in .m and .pdf formats derived from the .mlx so web users can see the content as .mlx files appear "binary" to git.

rmtfleming:
Hi Thomas,
please put the verification scripts into 
modelGeneration/modelVerification
Regards,
Ronan

tpfau:
closing in favor of [420](https://github.com/opencobra/cobratoolbox/pull/420)

PR #420, opened 2017-05-09
WIP: Io tutorial and update

tpfau:
Updated, mergable pr for original [PR419](https://github.com/opencobra/cobratoolbox/pull/419)
IOTutorial added and the IO is updated to be much more flexible when loading models.
Also added a checkModel() function - needs to be updated at some point to actually represent the required/optional fields, currently only a subset is used.

read and writeCbModel are adapted to accept parameter value pairs and to allow Excel and matlab IO (i.e. you can load models from mat files). I.e. all model IO can no go through these functions (no more xls2model etc necessary. The function will remain but its not required).
Set a definition for Excel format models and their IO (could be updated in the future)
The latter functionality depends on the model struct and currently only those with required fields present will be loaded. This again depends on the definition of those fields and is currently saved in getDefinedFieldProperties (the function will have to be adapted to the new field definitions in the future).
Moved the mps io from writeCbModel to writeLPProblem (the old convertCobraLP2mps is changed to that function) and updated the respective tests.
Also updated the tests for printRxnFormula/printFluxVector to reflect the change in printRxnFormula (irreversible reactions are no longer determined by "rev" but by lower bounds).


tpfau:
Closing temporarily for updates

laurentheirendt:
@tpfau, do you need my help here?

tpfau:
@laurent:
This should be ready to merge. Are there any comments/additional corrections I should do?

laurentheirendt:
@tpfau, do you mind doing a rebase here as well? Thanks! 🌵

src/base/io/writeCbModel.m
```diff
@@ -1,51 +1,98 @@
-function outmodel = writeCbModel(model,format,fileName,compSymbolList,compNameList,sbmlLevel,sbmlVersion, solverParams)
 % Writes out COBRA models in various formats
+% The "writeCbModel" function relies on another function
+% "io/utilities/writeSBML.m" to convert a COBRA-Matlab structure into
```
laurentheirendt: Something is not right here with the header of the function. This will cause problems when generating the documentation.

src/base/io/writeCbModel.m
```diff
@@ -1,51 +1,98 @@
...
 %    `io/utilities/writeSBML.m` to convert a COBRA-Matlab structure into
 %    a libSBML-Matlab structure and then call `libSBML` to export a
 %    FBCv2 file. The current version of the `writeSBML.m` does not require the
 %    SBML toolbox (http://sbml.org/Software/SBMLToolbox).
 
-if ~exist('compSymbolList','var') || isempty(compSymbolList)
-    compSymbolList = {'c','m','v','x','e','t','g','r','n','p','l','y'};
-    compNameList = {'Cytoplasm','Mitochondrion','Vacuole','Peroxisome','Extracellular','Pool','Golgi','Endoplasmic_reticulum','Nucleus','Periplasm','Lysosome','Glycosome'};
-end
+%For backward compatability, we are checking whether the old signature is
```
laurentheirendt: Having a comment on the first line will break the documentation generation. Can you put this comment as an in-line comment?

src/base/io/writeCbModel.m
```diff
@@ -102,308 +149,173 @@
     case {'text', 'txt'}
         fid = fopen(fileName,'w');
         fprintf(fid,'Rxn name\t');
-        if (isfield(model,'rxnNames'))
-            fprintf(fid,'Rxn description\t');
-        end
+        %         if (isfield(model,'rxnNames'))
```
laurentheirendt: Do we need these lines?

src/base/io/writeCbModel.m
```diff
@@ -102,308 +149,173 @@
...
             fprintf(fid,'Gene-reaction association\t');
         end
-        fprintf(fid,'Reversible\tLB\tUB\tObjective\n');
+        fprintf(fid,'LB\tUB\tObjective\n');
         for i = 1:nRxns
             fprintf(fid,'%s\t',model.rxns{i});
-            if (isfield(model,'rxnNames'))
-                fprintf(fid,'%s\t',model.rxnNames{i});
-            end
+            %             if (isfield(model,'rxnNames'))
```
laurentheirendt: Do we need these lines?

src/base/io/writeCbModel.m
```diff
@@ -102,308 +149,173 @@
...
     case 'sbml'
         % sbmlModel = convertCobraToSBML(model,sbmlLevel,sbmlVersion,compSymbolList,compNameList);
-        outmodel = writeSBML(model,fileName,compSymbolList,compNameList)
-%         if exist('fileName','var')&&~isempty(fileName)
-%             OutputSBML(sbmlModel,fileName);
-%         else
-%             OutputSBML(sbmlModel);
-%         end
+        outmodel = writeSBML(model,fileName,input.compSymbols,input.compNames);
+        %         if exist('fileName','var')&&~isempty(fileName)
```
laurentheirendt: Do we need these lines?

src/base/io/writeLPProblem.m
```diff
@@ -0,0 +1,246 @@
...
+            [Contain]=BuildMPS(Ale, ble, Aeq, beq, osense*c, lb, ub, PbName);
+            OK=SaveMPS(fileName, Contain);
+        else
+            [solution] = BuildMPS(Ale, ble, Aeq, beq, osense*c, lb, ub, PbName,'MPSfilename',MPSfilename,'EleNames',EleNames,'EqtNames',EqtNames,'VarNames',VarNames, 'Integer',intIndex,'Binary',binaryIndex);
+            OK = ~isempty(solution);
+        end
+        %display([' > The .MPS file <', MPSfilename, '> has been written to ', pwd]);
+end
+% 
+% E=false(mlt,1);
```
laurentheirendt: Do we need this code?

src/base/print/printRxnFormula.m
```diff
@@ -11,14 +11,17 @@
 % INPUTS:
 %    model:             COBRA model structure
 %
-% OPTIONAL INPUTS:
+% OPTIONAL INPUTS
 %    rxnAbbrList:       Abbrs of reactions whose formulas are to be printed
+%
+% OPTIONAL INPUTS as 'ParameterName',Value pairs:
```
laurentheirendt: This will break the documentation generation ... The header can only be `OPTIONAL INPUTS:`

test/verifiedTests/testMPS/testMPS.m
```diff
@@ -20,20 +20,21 @@
...
 assert(out == 1);
 
 % run using a legacy interface (creates a file LP.mps)
 solution = solveCobraLP(model, 'solver', 'mps');
 assert(isempty(solution));
 
 % run using writeCbModel
-writeCbModel(model, 'mps', 'LP2.mps', [], [], [], [], []);
+writeLPProblem(model, 'problemName','mps','fileName','LP2.mps','writeMatrix',0);
+%writeCbModel(model, 'mps', 'LP2.mps', [], [], [], [], []);
```
laurentheirendt: I guess that this line is obsolete ...

src/base/io/writeCbModel.m
tpfau: That's from the old header. I'll remove it.

src/base/io/writeCbModel.m
tpfau: Changed

src/base/io/writeCbModel.m
tpfau: If you mean the commented ones: no. Removed

src/base/io/writeCbModel.m
tpfau: as above

src/base/io/writeCbModel.m
tpfau: no, they were already deprecated before.

src/base/io/writeLPProblem.m
tpfau: No. (Just left it in for reference, Removed it now

src/base/print/printRxnFormula.m
tpfau: Well.... We need a distinction between "optional" and Parameter/Value pairs. 
Optional is an argument with a fixed position. 
Param/Value pairs don't have a fixed position. 
And we do already have functions with Param/Value pairs, so this will need to be updated in the Documentation generation.

test/verifiedTests/testMPS/testMPS.m
tpfau: you mean the commented one. yes. 
Removed it.

tpfau:
I noticed, that I will need the model Fields Corrections to be in the Toolbox in order to set up the tutorial properly. 
The reason is as follows:
I would modify the mat loading mechanism to automatically do a conversion of old fields (and an addition of osense/csense, which are missing in most models), so that (most) models pass the checkModel test.

src/base/print/printRxnFormula.m
laurentheirendt: Let's see then - this is quite a change then in the documentation generation scripts.

tpfau: True, But essentially, it woudl just be another Option wouldn't it? So a third case in addition to 
INPUT
OPTIONAL INPUT
OUTPUT
OPTIONAL OUTPUT
and new e.g.:
OPTIONAL PARAMETER VALUE PAIRS

laurentheirendt: yes ... @syarra knows more on that

tpfau:
As for file loading:
readCbModel currently assumes that it will only lookup the current folder. So what I can do is copy the files to the at the beginning of the tutorial. But Other than that I don't see how this could be addressed.

laurentheirendt:
`initCobraToolbox` adds the path, so the model should be found ... Can you not define the absolute path?

tpfau:
The problem is:
I'm currently doing the following: If I only get a name (no extension), I'm looking for matching files on the current directory, and if multiple exist, I offer a popup. I can't do this for the whole path.
So I'll copy some stuff over to the tutorial path at the start of the tutorial. This should work.

tpfau:
OK, tutorial added and last bug fixed. Should be ready to merge now

laurentheirendt:
@tpfau, great! Do we still need [this](https://github.com/tpfau/cobratoolbox/blob/IOTutorialAndUpdate/src/base/io/COBRA_structure_fields.xlsx) file?

@jacekwachowiak, may you format the documentation here?

tpfau:
Are those updates ok wrt to documentation?. About the file: No, but the ModelFields.md will still get a final update at some point.

rmtfleming:
The specification of model fields should be done in a .m file, used programmatically for various functions in the toolbox, and any .md generated automatically from it.

tpfau:
please have a look at generateFieldDescriptionFile, and getDefinedFieldProperties.m. These two files define and generate the COBRAModelFields.md file.

laurentheirendt:
Thanks @tpfau for this. Good start 🎉 

I would recommend however to call `generateFieldDescriptionFile()` automatically in order to generate the documentation instead of manually.

cc: @syarra

PR #424, opened 2017-05-11
Model field corrections

tpfau:
Removed Duplicate Fields and single use Fields. 
Adapted some fields, reducing the List of fields present in the Toolbox.


laurentheirendt:
@tpfau, do you need my help here?

tpfau:
Just done the rebase. Should now be ready to merge (except if you have comments/requests).

test/verifiedTests/testTopology/testFindExtremePathway.m
```diff
@@ -46,14 +42,14 @@
...
     if solverOK == 1
-        v = findExtremePathway(model);
-        assert(all(abs(model.S * v) < tol))
-
-        % testing findExtremePathway with different arguments
-        if isfield(model, 'revRxns')
-            model = rmfield(model, 'revRxns');
-        end
-
+%         v = findExtremePathway(model);
```
laurentheirendt: Does this test not work anymore?

src/base/io/utilities/outputNetworkCytoscape.m
```diff
@@ -154,11 +154,11 @@
...
 fclose(fidEdgeType);
 if (isfield(model,'subSystems'))
-    if (isfield(model,'subSystemsMet'))
-        for i = 1:length(model.mets)
-            fprintf(fidSubSys,'%s = %s\n',model.mets{i},model.subSystemsMet{i});
-        end
-    end
+%     if (isfield(model,'subSystemsMet'))
+%         for i = 1:length(model.mets)
+%             fprintf(fidSubSys,'%s = %s\n',model.mets{i},model.subSystemsMet{i});
```
laurentheirendt: Don't you want to add a `printLevel` here?

src/base/io/utilities/convertOldStyleModel.m
```diff
@@ -0,0 +1,109 @@
+function model = convertOldStyleModel(model)
```
laurentheirendt: Do we have a way of testing this function? It seems important ...

tpfau: Not really.
I could put in a function if you want, though I think it woud be pretty artificial, as I have to take an existing model with "old" fields and move it to a new model, and check whether the fields indeed got moved/merged.

src/base/io/utilities/outputNetworkCytoscape.m
tpfau: I removed it mainly because this is essentially only a printout where a field was used, that doesn't exist any more (and imo never existed in the first place)

test/verifiedTests/testTopology/testFindExtremePathway.m
tpfau: it is essentially obsolete, as it is the same as the later tests, since revRxns and reversibilities were removed as model fields.

src/base/io/utilities/outputNetworkCytoscape.m
laurentheirendt: Ok, so if the field doesn't exist anymore, then I would suggest to remove the code for it.

src/base/io/utilities/convertOldStyleModel.m
tpfau: ok. I'll remove it.

src/base/io/utilities/outputNetworkCytoscape.m
tpfau: ok, I'll remove it

src/base/io/utilities/convertOldStyleModel.m
laurentheirendt: I guess it is important as it ensures backward compatibility ... You can also rename the fields temporarily from an already 'new' model, and then test whether they get changed. 

We can also put this off for later.

test/verifiedTests/testTopology/testFindExtremePathway.m
laurentheirendt: I can see why the test on `revRxns` is obsolete, but why is the test on ` v = findExtremePathway(model);`?

As `revRxns` is obsolete, I would recommend removing the test code ...

tpfau: Ahh ok... My fault.
Yes, it doesn't work any more (since the removal of revRxns). at least if no lb/ub are defined. 
I'll move it further down in the function. after lb/ub are added, at which point it will work again.

tpfau:
Anything Else needed here? 
I would need this to go in to properly set up the IO Tutorial and again update the readCbModel function (i.e. that it tries to convert models found in mat files to the new structure -> including things like osense/csense etc if missing).

laurentheirendt:
Very nice @tpfau 👍 🎉

PR #432, opened 2017-05-15
Add data for tests

arichelle:


laurentheirendt:
@arichelle, may you include the `.mat` files that are needed to run your code in [this function](https://github.com/opencobra/cobratoolbox/blob/develop/test/models/retrieveModels.m)? The aim is to not distribute models in `.mat` directly within this repository, but to download them from a published reference as needed.

Closing the PR now.

PR #436, opened 2017-05-17
creation of a new commit for pFBA new version, without use of fastFVA

arichelle:
I had trouble to update the pFBA script and publish it in the branch "add-extraction-methods", therefore i created i new branch just for the addition of pFBA script.


rmtfleming:
I have been trying to add generic functions to the cobra toolbox such that things like fluxVariability.m can call fastFVA as an option, if it is available. This separates the purpose of the function from the algorithmic implementation of the function, so I prefer to see fluxVariability.m used like you have as it is more generic, but it would be even better if fluxVariability could call fastFVA!

PR #437, opened 2017-05-17
pFBA update

arichelle:


test/verifiedTests/testpFBA/testpFBA.m
```diff
@@ -47,6 +47,12 @@
...
         % run pFBA
         fprintf('\n*** Test basic pFBA calculations ***\n\n');
+        fprintf('\n** Optimal solution - minimize flux not associated with gene: glucose\n');
+        [t_objGenes.GeneClasses_glc1 t_objRxns.RxnClasses_glc1 t_objModel.modelIrrev_glc1] = pFBA(model_glc, 'geneoption', 2);
+
+        fprintf('\n** Optimal solution - minimize flux not associated with gene: lactate\n');
+        [t_objGenes.GeneClasses_lac1 t_objRxns.RxnClasses_lac1 t_objModel.modelIrrev_lac1] = pFBA(model_lac, 'geneoption', 2);
+        
         fprintf('\n** Optimal solution - minimize gene-associated flux: glucose\n');
         [t_objGenes.GeneClasses_glc1 t_objRxns.RxnClasses_glc1 t_objModel.modelIrrev_glc1] = pFBA(model_glc, 'geneoption', 1);
```
tpfau: This shoudl also actually test the results if possible (i.e. have an assertion somewhere).

tpfau:
A comment about the failed tests:
There is a small bug in findGenesFromReactions (fixed here #438 ).
@arichelle could you check if it is intentional, that the reactions use do not have any gene associations?

arichelle:
Yes it is intentional to give an option for which the minimization only that into account the reactions that do not have gene association. Doing so, the algorithm try to get rid of these and favor gene-associated reactions within the optima to counterbalance the minimization.

@tpfau with the fix #438, does the pFBA algorithm run properly ?

test/verifiedTests/testpFBA/testpFBA.m
arichelle: Ok will update 'testpFBAData.mat' and the results part

PR #438, opened 2017-05-18
initialized geneList

tpfau:
geneList could be empty, which would raise an error, if only non-gene associated reactions are checked.
(This happened here: #437 )


laurentheirendt:
Thank @tpfau 👍

PR #447, opened 2017-05-18
update of scripts related to FASTCORE

arichelle:


src/dataIntegration/transcriptomics/FASTCORE/fastcore.m
```diff
@@ -1,119 +1,155 @@
...
 %       - Ronan Fleming, commenting of code and inputs/outputs
+%       - Anne Richelle, code adaptation to fit with createTissueSpecificModel
+
+
+    if ~exist('printLevel','var')
+        %For Compatability with the original fastcore syntax
+        printlevel = 1;
+    end
+
+    %Define the set of core reactions
```
tpfau: I would still argue, that expressionRxns and threshold should be preformed BEFORE going into fastcore.
This would be preprocessing which is necessary for any of the tissueCreation Algorithms and not part of the individual algorithm. 
So I would prefer not to have this in the fastcore function

mariapirespacheco:
Hi
As one of the author of FASTCORE, I do not recommend using
arbitary/heuristic thresholds with FASTCORE or with any algorithms. We
published FASTCORMICS (
https://bmcgenomics.biomedcentral.com/articles/10.1186/s12864-015-1984-4)
in order to avoid setting these kind of thresholds. The setting of a
threshold has a huge impact on the output model.
Maria

src/dataIntegration/transcriptomics/FASTCORE/fastcore.m
rmtfleming: I am sure that with a generous collegial approach, yee can come to some compromise that

mariapirespacheco:
Methods that propose a discretization method should be run with this
method. It is true for FASTCORE but also for mCADRE and others. You can
propose an option to run the script with other thresholds but it should not
be done by default.
The reason people come up with discretization methods is to avoid things
like taking the upper 25th...


And a warning should be given that heursitic threshold have a huge impact
on the output model.
The same threshold cannot be used for the every algorithm as every
algorithm is build on different assumptions. FASTCORE expect that the core
set has a high confidence level therefore the threshold should be
stringent...


Maria

arichelle:
@mariapirespacheco completely agree with you about the impact of using a heuristic threshold but currently, to my knowledge, it does not exist one universal methodology to define the cutoff for gene activity (which is one of the big weakness of extraction methods based on transcriptomic data). Each of the method that will be implemented deal differently with this cutoff and we will try to give to the user the most flexibility as possible to modulate it depending on the method used and also inform them about the way to modulate it. As you said, a small note will need to be introduced in the description of the global function call for extraction methods to inform the influence of using the default setting for threshold on the model extracted.

Moreover, concerning fastcormics, feel free to propose an integration of this method as a new solver option for createTissueSpecificModel ;)

mariapirespacheco:
@arichielle I am glad that you agree on the impact of heuristic threshold.
There is no universal methods because the algorithms are different.
Searching for an universal method does not make sense.
As there is no universal method, I suggest that the one proposed by the
authors is used by default. The authors of each algorithm knows better how
to feed its algorithms as they benchmarked it. Any user can of course use
other threshold but I strongly advise anyone to test if these modification
changes the performance.  A user that runs an algorithm expect to have the
performance described in the paper, which might not be the case if you do
not use the recommended settings. And I am not sure that an non experienced
 Cobra user might notice that he actually running a modified version.


I have to discuss these option with the other authors. I mean it would be
rude to add work of other people without noticing them :P.
Have a nice WE

PR #448, opened 2017-05-18
FIX - bounds trouble in converToIrreversible.m

arichelle:
If original reaction has a positive lb, backwards reaction should have nonnegative upper bound

### DIFF
--- a/src/reconstruction/refinement/convertToIrreversible.m
+++ b/src/reconstruction/refinement/convertToIrreversible.m
@@ -94,7 +94,12 @@ for i = 1:nRxns
         modelIrrev.rxns{cnt} = [model.rxns{i} '_b'];
         modelIrrev.rev(cnt) = true;
         modelIrrev.lb(cnt) = 0;
-        modelIrrev.ub(cnt) = - model.lb(i);
+        %if original reaction has a positive lb, backwards reaction should have nonnegative upper bound.
+        if model.lb(i)<0
+            modelIrrev.ub(cnt) = - model.lb(i);
+        else
+            modelIrrev.ub(cnt) =  model.lb(i);
+        end
         modelIrrev.c(cnt) = 0;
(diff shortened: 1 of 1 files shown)


src/reconstruction/refinement/convertToIrreversible.m
```diff
@@ -1,142 +1,147 @@
...
+        cnt = cnt + 1;
+        matchRev(cnt) = cnt - 1;
+        matchRev(cnt-1) = cnt;
+        modelIrrev.rxns{cnt-1} = [model.rxns{i} '_f'];
+        modelIrrev.S(:, cnt) = - model.S(:, i);
+        modelIrrev.rxns{cnt} = [model.rxns{i} '_b'];
+        modelIrrev.rev(cnt) = true;
+        modelIrrev.lb(cnt) = 0;
+        %if original reaction has a positive lb, backwards reaction should have nonnegative upper bound.
+        if model.lb(i)<0
```
tpfau: if the original reaction has a positive lower bound, the backward reaction should have both bounds set to 0, as it is not reversible.
In essence: this should not check for rev at all but should check for lb < 0

tpfau: essentially: 
Forward:
```
modelIrrev.lb(cnt) = max(model.lb(i),0);
modelIrrev.ub(cnt) = max(0,model.ub(i));
```
backward:
```
modelIrrev.lb(cnt) = max(-model.ub(i),0);
modelIrrev.ub(cnt) = max(0,-model.ub(i));
```
and anything where ub == lb == 0 can be dropped.

arichelle: you're right, I've update the script

tpfau:
looks fine

laurentheirendt:
Thanks @arichelle 👍

PR #449, opened 2017-05-18
solveCobra MILP returns a solution when time limit is reached

arichelle:


laurentheirendt:
Nice! Thanks @arichelle 🎉 

@tpfau: you are right, we can maybe down the road add the `timelimit` option for other solvers.

PR #452, opened 2017-05-19
Integration of the extraction methods in createTissueSpecificModel.m

arichelle:
Here a draft of createTissueSpecificModel. I've, for the moment, removed the options of processing the expression data, to focus on the integration of the methods and how to deal with the parameters that could be defined by the user for each extraction methods. 

Once, we agree on this step, we can open the discussion about the way to provide a processing data option for the user

### DIFF
--- a/src/dataIntegration/transcriptomics/createTissueSpecificModel.m
+++ b/src/dataIntegration/transcriptomics/createTissueSpecificModel.m
@@ -22,8 +22,6 @@ function [tissueModel,Rxns] = createTissueSpecificModel(model, expressionData,ex
 %	funcModel               1 - Build a functional model having only reactions
 %                           that can carry a flux (using a consistency check), 0 - skip this
 %                           step (Default = 0)
-%	options                 1- user defines the parameters for running the extraction
-%                           methods 0- extraction methods use default settings
 %   param                   structure containing the specific parameters reauired for each
 %                           extraction method need only to be provided is options = 1 (see
 %                           PARAMETER section below)
@@ -144,9 +142,6 @@ if ~exist('funcModel','var') || isempty(funcModel)
     funcModel = 0;
 end
 
-if ~exist('options','var') || isempty(options)
-    options = 0;
-end
 if ~exist('param','var') || isempty(param)
     param = {};
 end
@@ -175,7 +170,7 @@ end
 
 switch solver
     case 'iMAT'
-        if options==1
+        if ~isempty(param)
             tissueModel = iMAT(model, expressionRxns, param.threshold_lb, param.threshold_ub, param.tol, param.core, param.logfile, param.runtime); 
         else
             % if options are not defined use default settings
@@ -183,7 +178,7 @@ switch solver
         end
         
     case 'GIMME'
-        if options==1
+        if ~isempty(param)
             tissueModel = GIMME(model, expressionRxns, param.threshold, param.obj_frac); 
         else
             % if options are not defined use default settings
@@ -191,7 +186,7 @@ switch solver
         end
         
     case 'INIT'
-        if options==1
+        if ~isempty(param)
             tissueModel = INIT(model, param.weights, param.tol, param.runtime, param.logfile); 
         else
             % if options are not defined use default settings
@@ -206,7 +201,7 @@ switch solver
         end
         
     case 'MBA'
-        if options==1
+        if ~isempty(param)
             tissueModel = MBA(model, param.medium_set, param.high_set, param.tol, param.core);       
         else
             % if options are not defined use default settings
@@ -222,7 +217,7 @@ switch solver
         end
 
     case 'mCADRE'
-        if options==1
+        if ~isempty(param)
             tissueModel = mCADRE(model, param.ubiquityScore, param.confidenceScores, param.protectedRxns, param.checkFunctionality, param.eta, param.tol);
         else
             % if options are not defined use default settings
@@ -243,7 +238,7 @@ switch solver
         end
     
     case 'fastCore'
-        if options==1
+        if ~isempty(param)
             tissueModel = fastcore(model, param.core, param.epsilon, param.printlevel);
         else
             % if options are not defined use default settings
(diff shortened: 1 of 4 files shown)


tpfau:
Just as a question:
Should this function be backward compatible? Then we would need to cover cases where old input is provided, or do we accept that it will not be backward compatible?

arichelle:
@tpfau I think that it could be interesting to offer the possibility to the user to process their data as done before. The ideas you were proposing about the call -1/0/1 depending on the absence/unknow/presence was cool and part of the code written before can be used for that

src/dataIntegration/transcriptomics/extractGPRs.m
```diff
@@ -0,0 +1,120 @@
...
+    parsedGPR = [];
+    corrRxn = [];
+    cnt = 1;
+
+    for i = 1:length(model.rxns)           
+        if length(model.rules{i}) > 1
+            % Parsing each reactions GPR containing "OR" rule
+
+            [parsing{1,1},parsing{2,1}] = strtok(model.rules{i},'|');           
+            for j = 2:1000
```
tpfau: while this j is probably sufficient, shouldn't it depend on the actual number of tokens?

src/dataIntegration/transcriptomics/findUsedGenesLevels.m
```diff
@@ -0,0 +1,44 @@
...
+%
+%   gene_expr           vector of expression values associated to each
+%                       gened_id
+%
+% S. Opdam & A. Richelle May 2017
+
+
+gene_id = {};
+gene_expr=[];
+for i = 1:numel(model.genes)
```
tpfau: gene_id = model.genes ?

src/dataIntegration/transcriptomics/findUsedGenesLevels.m
```diff
@@ -0,0 +1,44 @@
...
+        
+    cur_ID = gene_id{i};
+	dataID=find(ismember(expressionData.gene,cur_ID )==1);
+	if isempty (dataID)
+    	gene_expr(i)=-1;
+    elseif length(dataID)==1
+    	gene_expr(i)=expressionData.value(dataID);
+    elseif length(dataID)>1
+    	disp(['Double for ',num2str(cur_ID)])
+    	gene_expr(i)=sum(expressionData.value(dataID));
```
tpfau: Is this intentional, that you sum up here and not take the average? 
Chips with multiple probes for the same gene will have very high values for those genes if the expressions are just summed up and I'm not sure, whether thats correct.

src/dataIntegration/transcriptomics/createTissueSpecificModel.m
```diff
@@ -1,701 +1,292 @@
...
+%
+%OPTIONAL INPUTS
+%	exRxnRemove             Names of exchange reactions to remove
+%                           (Default = [])
+%	solver                  Use either 'GIMME','iMAT','INIT','GIMME','mCADRE','fastCore'
+%                           (Default = 'GIMME')
+%	funcModel               1 - Build a functional model having only reactions
+%                           that can carry a flux (using a consistency check), 0 - skip this
+%                           step (Default = 0)
+%	options                 1- user defines the parameters for running the extraction
```
tpfau: I think options can be omitted. if params is provided, the defaults should be overwritten. This is kind of a replication You have to indicate that you want the defaults to be overwritten AND provide the overriding values. In myy opinion just providing the values should be sufficient.

src/dataIntegration/transcriptomics/findUsedGenesLevels.m
arichelle: Agree, i will change by taking the mean

src/dataIntegration/transcriptomics/createTissueSpecificModel.m
arichelle: agree

src/dataIntegration/transcriptomics/extractGPRs.m
```diff
@@ -0,0 +1,120 @@
...
+                [parsing{j,1},parsing{j+1,1}] = strtok(parsing{j,1},'|');
```
arichelle: will try to improve that ASAP

src/dataIntegration/transcriptomics/createTissueSpecificModel.m
```diff
@@ -1,701 +1,287 @@
...
-        if sumExp == 0
-            Transcripts.UnExpressed = [Transcripts.UnExpressed;locus];
-        elseif sumExp == length(loc)
-            Transcripts.Expressed = [Transcripts.Expressed;locus];
-        end
-    end
-end
-Transcripts.Locus = mem_tran;
+function [tissueModel,Rxns] = createTissueSpecificModel(model, expressionData,exRxnRemove,solver,funcModel,options,param)
+
```
tpfau: If possible, please use an input parser for this method. 
e.g. I can imagine the most common call would be that someone wants to just set the solver and param, which now would lead to a call that sets two things just to their default. 
if you use an input-parser and parameter/value pairs, you can just set those parameters you want, and don't have to care about the rest.
maybe even merge the solver into the param input. (i.e. param.solver indicates which solver to use, and extract all useable fields afterwards).

tpfau:
a struct field makes code much less readable, since fields for that struct can be declared anywhere, and it is not directly clear which fields have been set.
It is also, for an unexperienced user, much more difficult to understand how to use the function.

arichelle:
As Ronan, I more used to have an optional structure field but I agree with Thomas point of view... Let me know what you think I should do at the end ;)

src/dataIntegration/transcriptomics/extractGPRs.m
arichelle: OK I've introduced your parser and modified the codes that use parsedGPR variable. I'll push the new version soon.

arichelle:
This PR can be closed, new version of createTissueSpecificModel available in PR #535

laurentheirendt:
Thanks @arichelle!

PR #453, opened 2017-05-19
scripts to test a list of metabolic tasks

arichelle:
This PR contains the scripts used to test a list of metabolic task. These codes are inspired from the checkTasks.m from the RAVEN toolbox. I've adapted them to be compatible with the cobratoolbox but I've also modify the way to define the task itself as now it only rely on the management of constraints on flux related to a list of input substrates and output products and not anymore on the relaxation of pseudo-stationary assumption.

These scripts could be very useful to test if the extracted model keep the same list of metabolic functionalities as the original one. We are currently working on this kind of analysis with @nel3 . Any comments/suggestions for improvement are welcome.

### DIFF
--- a/docs/source/modules/MetabolicTask.rst
+++ b/docs/source/modules/MetabolicTask.rst
@@ -1,0 +1,9 @@
+
+.. _MetabolicTask:
+
+
+MetabolicTask
+-----
+.. automodule:: src.dataIntegration.transcriptomics.MetabolicTask
+   :members:
+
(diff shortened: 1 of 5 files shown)


tpfau:
We need to add the [Raven License ](https://github.com/SysBioChalmers/RAVEN/blob/master/LICENSE) file if we take code from the RAVEN toolbox. The same applies for INIT and mCADRE (for the latter we need the license file from their github page

laurentheirendt:
@jacekwachowiak, may you format the documentation here? Thanks.

rmtfleming:
The idea of

PR #462, opened 2017-05-30
File io updates

tpfau:
This is a switch to a new SBML IO using libsbml 5.15 instead of 5.13.
It also includes a rework of the readSBML and writeSBML functions. 


src/base/io/utilities/SBML/parseSBMLNotesField.m.orig
```diff
@@ -0,0 +1,120 @@
+<<<<<<< 5933086c821a593ef371fa8475e061683c3d7764:src/base/io/utilities/parseSBMLNotesField.m
```
laurentheirendt: @tpfau, may you remove this `.orig` file? Thanks.

src/base/io/utilities/SBML/parseSBMLNotesField2012.m.orig
```diff
@@ -0,0 +1,102 @@
+<<<<<<< 5933086c821a593ef371fa8475e061683c3d7764:src/base/io/utilities/parseSBMLNotesField2012.m
```
laurentheirendt: @tpfau, may you also remove this `.orig` file?

test/verifiedTests/testSBML/testReadSBML.m
```diff
@@ -52,7 +32,7 @@
 
 
 % load the models
-modelArr = {'Abiotrophia_defectiva_ATCC_49176.xml', 'STM_v1.0.xml', 'iIT341.xml', 'Ec_iAF1260_flux1.xml'};
+modelArr = { 'STM_v1.0.xml', 'iIT341.xml'};
```
laurentheirendt: I would suggest however to leave `Abiotrophia_defectiva_ATCC_49176.xml'` as a test model.

test/verifiedTests/testSBML/testReadSBML.m
```diff
@@ -78,10 +58,10 @@
 
 
             % define the maximum objective values calculated from pre-converted .mat files
-            modelFBAf_max = [0.149475406282249; 0.477833660760744; 0.692812693473487; 0.736700938865275];
+            modelFBAf_max = [0.477833660760744; 0.692812693473487];
 
             % define the minimum objective values
-            modelFBAf_min = [0.0; 0.0; 0.0; 0.0];
+            modelFBAf_min = [0.0; 0.0];
```
laurentheirendt: These fields will then have to be updated...

test/verifiedTests/testSBML/testReadSBML.m
```diff
@@ -122,23 +102,12 @@
 
     model.lb = model.lb(:);
     model.ub = model.ub(:);
-
+    model.comps = {'e'};
+    model.compNames = {'ExtraCellular'};
     % add the fields outputted by readSBML
     [model.modelVersion.SBML_level, model.modelVersion.SBML_version, model.modelVersion.fbc_version] = deal(3, 1, 2);
```
laurentheirendt: What is the reason for removing the `metField` variable?

test/verifiedTests/testSBML/testWriteSBML.m
```diff
@@ -36,15 +45,7 @@
 else
     warning(['The file', fullFileNamePath, ' does not exist and could not be deleted.']);
 end
-
-load('Ec_iJR904.mat')
-model = convertOldStyleModel(model);
```
laurentheirendt: What is the reason for removing the test on `convertOldStyleModel`?

test/verifiedTests/testSBML/testWriteSBML.m
```diff
@@ -36,15 +45,7 @@
...
-% use a different deck to convert a COBRA model to an SBML file
-modelSBML = convertCobraToSBML(model);
```
laurentheirendt: What is the reason for removing the test on `convertCobraToSBML`?

src/base/io/utilities/writeSBML.m
```diff
@@ -933,21 +800,11 @@
...
-str = strrep(str,',','_COMMA_');
-str = strrep(str,'.','_PERIOD_');
-str = strrep(str,'''','_APOS_');
-str = regexprep(str,'\(e\)$','_e');
-str = strrep(str,'&','&amp;');
-str = strrep(str,'<','&lt;');
-str = strrep(str,'>','&gt;');
-str = strrep(str,'"','&quot;');
-end
+% %% Format For SBML
```
laurentheirendt: Do we still need these lines here?

src/base/io/utilities/readSBML.m
```diff
@@ -1,787 +1,494 @@
...
+%extract the fbc_associations fbc_association field or return an empty
+%string
+if isempty(fbc_gprAssoc) || isempty(fbc_gprAssoc.fbc_association)
+    rule = '';
+    return
+end
+try
+    rule = fbc_gprAssoc.fbc_association.fbc_association;
+catch
+    disp('blubb')
```
laurentheirendt: What does this message mean?

src/base/io/utilities/readSBML.m
```diff
@@ -1,787 +1,494 @@
...
 end
+end
+
+function field = extractfield(startstruct,fieldname)
 
+if ~isfield(startstruct,fieldname)
+    error('Field %s does not exist',fieldname)
+end
+field = {startstruct.(fieldname)};
+%if all(cellfun(@isnumeric , field))
```
laurentheirendt: Do we still need these lines?

src/base/io/utilities/readSBML.m
```diff
@@ -1,787 +1,494 @@
...
-        if iscell(genes{i})
-            [tmp,geneInd] = ismember(genes{i},allGenes);
+%% Set up the objective. The default is maximisation.
+model.osense = -1;
+if isfield(modelSBML,'fbc_objective')
+    %We only support the first one we find
+    if ~isempty(modelSBML.fbc_objective)
+        osenseStr = modelSBML.fbc_objective(1).fbc_type;
+        if strcmp(osenseStr , 'maximize')
+            model.osense = -1;
```
laurentheirendt: This condition is not necessary, as `model.osense` has already been initialized to `-1`.

tpfau: fixed

src/base/io/utilities/SBML/parseSBMLNotesField.m.orig
tpfau: done

src/base/io/utilities/SBML/parseSBMLNotesField2012.m.orig
tpfau: done

src/base/io/utilities/readSBML.m
tpfau: Leftover from debugging that I forgot to remove.

src/base/io/utilities/readSBML.m
tpfau: no

src/base/io/utilities/writeSBML.m
tpfau: no

test/verifiedTests/testSBML/testReadSBML.m
tpfau: I put it in again.

test/verifiedTests/testSBML/testReadSBML.m
tpfau: done

test/verifiedTests/testSBML/testReadSBML.m
tpfau: readSBML removes any empty fields. The only fields guaranteed to be present in a model returned by readSBML are (if there are any reactions/mets in the SBML file):
S
lb
ub
c
osense
csense
b
rxns
mets

Everything else could potenitally be removed.

test/verifiedTests/testSBML/testWriteSBML.m
tpfau: This was not a test for convertOldStyle model but rather something to convert the ecoli model (which is very old). 
But: This test was invalid anyways as it compared an SBML model structure with a COBRA model structure. isSame actually returned false all the time, and it was just that both structs don't have any fields in common that led to numDiff being empty. so essentially, the check for ~any(numDiff) can make sense if one is sure, that both are proper models with potentially different field IDs, but is a bad idea, if that's not sure.

test/verifiedTests/testSBML/testWriteSBML.m
tpfau: I would alter that function to just call writeSBML (which it kind of does). The inputs of that function were ignored anyways (and it was just kept for backward compatability). Even the old writeSBML was unable to create any SBML Level 2 models and I would strongly discourage giving that option. 
I altered the function accordingly.

PR #469, opened 2017-05-31
Tutorial - Numerical characteristics

laurentheirendt:
- Added tutorial for numerical characteristics
- Added `external/spyc` and accompanying functions for colourful spy plot of stoichiometric matrix
- Added `checkScaling()` function to provide a solver recommendation based on the scaling `S`
- CI integration of tutorial and `checkScaling()`
- Test for `checkScaling()`


src/modelAnalysis/subspaces/checkScaling.m
```diff
@@ -0,0 +1,308 @@
...
+%
+% OPTIONAL INPUTS:
+%
+%    estLevel:                      level of estimation: `crude`, `medium`, `fine` (default)
+%    printLevel:                    verbose level (default: 1). Level 0 is quiet.
+%
+% OUTPUTS:
+%
+%    solverRecommendation:          estimation of precision (string, `double` or `quad`, default: `double`)
+%    scalingProperties:             structure with properties of scaling
```
tpfau: Would it be an idea to indicate which currently supported solvers could be used in either case?

src/modelAnalysis/subspaces/checkScaling.m
```diff
@@ -0,0 +1,308 @@
...
+    end
+
+    if abs(minS) > 0 && abs(ratioS) > 0
+        scalingProperties.ratioS = ratioS;
+        if printLevel > 0
+            fprintf(' Ratio of stoichiometric coefficients:         %1.2e\n', ratioS);
+        end
+    end
+    if abs(ratioS) > 0
+        scalingProperties.ratioS_orderOfMag = abs(floor(log10(abs(ratioS))));
```
tpfau: Is this really true?
if you have e.g. minS = -10000, and maxS = 5000, than ratioS will be 5000/-10000 = -0.5, while you could have coefficients of 0.00001 somewhere, which would be an enourmous scaling issue of 10 orders of magnitude. 

The same probably applies below.

src/modelAnalysis/subspaces/checkScaling.m
laurentheirendt: Yes, that could be an idea. I will implement that based on the table in `initCobraToolbox`.

src/modelAnalysis/subspaces/checkScaling.m
laurentheirendt: The ratio of the maximum and minimum stoichiometric coefficients is given for information only; the solver recommendation itself is based on the row and column scaling coefficients that are calculated by `gmscale`.

Do you think that it might be confusing for the end user to, however, display the ratio of maximum and minimum stoichiometric coefficients?

tpfau: I would just be a little confused because I think, that the "min/max" order of magnitude difference is mostly useless wrt scaling/numerical issues, while the order of magnitudes spanned could be an indicator for numerical issues. I would just rather expect the latter and not expect the former.

laurentheirendt: If I understand correctly, instead of:
````Matlab
minS = full(min(min(S)));
maxS = full(max(max(S)));
if abs(minS) > 0
    ratioS = maxS / minS;
else
    ratioS = maxS;
end
````
it would be better to use the absolute values of the stoichiometric coefficients of `S`:
````Matlab
minS = full(min(min(abs(S))));
maxS = full(max(max(abs(S))));
if abs(minS) > 0
    ratioS = maxS / minS;
else
    ratioS = maxS;
end
````

**Comment by laurentheirendt** [comment]
Closing temporarily to make further changes.

File: src/modelAnalysis/subspaces/checkScaling.m (hunk 2 above)
Comment (tpfau) [inline src/modelAnalysis/subspaces/checkScaling.m]: Just noticed this:
minS should be 
`minS = full(min(min(abs(S(S~=0)))));`
Otherwise you will for sure get 0 (no chance of not having it).

File: src/modelAnalysis/subspaces/checkScaling.m (hunk 1 above)
Comment (laurentheirendt) [inline src/modelAnalysis/subspaces/checkScaling.m]: I will implement this, but only later.

File: src/modelAnalysis/subspaces/checkScaling.m (hunk 2 above)
Comment (laurentheirendt) [inline src/modelAnalysis/subspaces/checkScaling.m]: But that is the point: I want to get `0` as the stoichiometric coefficient. Let's imagine you have the coefficients: `-10, 0, 10, 100`. Then I want to get the minimum absolute value coefficient (that is `0` in this case). The order of magnitude difference is then `2`. Otherwise, with the minimum absolute *non-zero* coefficient, I would get `10`, and the order of magnitude difference would be `1`.

Comment (tpfau) [inline src/modelAnalysis/subspaces/checkScaling.m]: Ok, true. But:

In this case I would suggest:
```Matlab
minSNZ = full(min(min(abs(S(S~=0)))));
minS = full(min(min(abs(S))));
maxS = full(max(max(abs(S))));
if minSNZ < 1
    minS = minSNZ
end
if abs(minS) > 0
    ratioS = maxS / minS;
else
    ratioS = maxS;
end
```
Otherwise if you have -10, 0.01, 10, 100, your ratio would be 100 while it should be 10000.



`

laurentheirendt: Not sure I understand. Right now:
```Matlab
>> S = [-10, 0.01, 10, 100];
>> minS = full(min(min(abs(S))));
>> maxS = full(max(max(abs(S))));
>> ratioS = maxS / minS
```
which returns:
```
ratioS =

       10000
```

tpfau: ok, bad example.... 
use: -10, 0.01, 10, 100, 0  (and any S matrix will have 0 elements).

laurentheirendt: Any sparse `S` would however not have elements that are `0` (unless something is wrongly defined) ... but I see your point.

tpfau: calling min on a sparse matrix, should still give you 0, as all non defined elements are 0.

laurentheirendt: In fact, I think it is easier then how you suggested initially, but then we can also remove the check `if abs(minS) > 0`:
```Matlab
minS = full(min(min(abs(S(S ~= 0)))));
maxS = full(max(max(abs(S))));
ratioS = maxS / minS;
```
Do you agree?

tpfau: In general: yes. 
And I think we can make that assumption. I can't imagine any (relevant) stoichiometric matrix that is full.

laurentheirendt: Some test matrices are full. Let me change that then.

laurentheirendt:
Closing temporarily to make further changes.

PR #471, opened 2017-05-31
Code for up- and down-regulation prediction

arichelle:
Scripts for prediction of up- and down-regulated gene-reaction pairs under two different external conditions. 


laurentheirendt:
Very nice @arichelle! May you however put the keyword `USAGE:` before `INPUT:`?

May you also try to push only **once**, i.e. to run `contribute` and select menu `[3]` only when you are done changing multiple files? Otherwise, with an open PR,  a new build is triggered for each commit, and the CI-server does not follow. Thanks 👍

arichelle:
Will do that ASAP

PR #478, opened 2017-05-31
adding gams function to find mustL set

snmendoz:
[*This GAMS function solves a MIP problem n-times until find all reactions that belongs to the MustL Set. This function is called from matlab, and reads inputs generated from matlab*]


laurentheirendt:
Thanks @snmendoz 🚗

PR #494, opened 2017-06-01
XLS IO Update

tpfau:
Update of the XLS IO to reflect the current structure.
This pr addresses and fixes issue #146


rmtfleming:
Please ensure that this is backward compatible with:
http://www.nature.com/nprot/journal/v5/n1/abs/nprot.2009.203.html
(Great if it is!)

tpfau:
It's not, and neither is the current xls2model function. But at least this version sticks to the definitions we made, and the sheet provided in the protocol paper will never be translatable directly, since it contains "Required/Optional" lines and those changed significantly.
So anyone using it can easily read the new documentation.

rmtfleming:
OK, please add an error message to that effect if xls2model is used in an old way inconsistent with its current incarnation. I am sure we will otherwise get forum questions about this.

src/base/io/utilities/xls2model.m
```diff
@@ -151,32 +185,31 @@
     %Default -1000
     lowerBoundList(isnan(lowerBoundList)) = -1000;
 else
-    lowerBoundList = -defaultbound*ones(length(rxnAbrList),1);
+    lowerBoundList = 1000*ones(length(rxnAbrList),1);
```
laurentheirendt: Why has `defaultbound` been removed here?

laurentheirendt: Why has `defaultbound` been removed here?

tpfau: Because this changed at some point and since I was basing a lot of my stuff on my FileIOUpdate branch, I simply overwrote the whole file, corrected now, along with an additional fix for IDs which are numbers.

laurentheirendt:
Thanks @tpfau ⚽️

PR #495, opened 2017-06-01
Model Creation Tutorial and update to Model Manipulation Tutorial

tpfau:
Addition of a model creation tutorial and update to the model manipulation tutorial to reflect changes in addReaction useability along with a set of small fixes to several files wrt rev and field names


src/modelAnalysis/exploration/printFluxBounds.m
```diff
@@ -0,0 +1,14 @@
+function printFluxBounds(model)
+%PRINTFLUXBOUNDS prints the reactionID and upper/lower flux bounds.
```
laurentheirendt: May you format the documentation here?

src/reconstruction/refinement/addReaction.m
```diff
@@ -1,4 +1,4 @@
-function [model, rxnIDexists] = addReaction(model, rxnName, metaboliteList, stoichCoeffList, revFlag, lowerBound, upperBound, objCoeff, subSystem, grRule, geneNameList, systNameList, checkDuplicate, printLevel)
+function [model, rxnIDexists] = addReaction(model, rxnID, varargin) %function [model, rxnIDexists] = addReaction(model, rxnName, metaboliteList, stoichCoeffList, revFlag, lowerBound, upperBound, objCoeff, subSystem, grRule, geneNameList, systNameList, checkDuplicate, printLevel)
```
laurentheirendt: Do we still need the old header here? Or could that be moved to the bottom of the function header?

src/reconstruction/refinement/addReaction.m
```diff
@@ -258,108 +275,143 @@
...
+        model = changeGeneAssociation(model,rxnID,grRule,geneNameList,systNameList);
     end
 end
 
 % Figure out if the new reaction already exists
 rxnInModel=false;
+duplicatePos = [];
 if (nNewMets > 0) && isempty(find(newMetsCoefs == 0, 1))
-    Stmp = [model.S;sparse(nNewMets,nRxns)];
+    Stmp = model.S;
```
laurentheirendt: Do we need a special condition for `Stmp = model.S;`? It seems as if `Stmp` is also set  in the `else` block.

There is actually no need for this `if` condition.

laurentheirendt:
May you rename `tutorials/ModelCreation` and `tutorials/ModelManipulation` to `tutorials/modelCreation` and `tutorials/modelManipulation`?

laurentheirendt:
Nice tutorials! In the `.pdf`, the lines are not properly broken. May you add `...` to break the lines in `MATLAB` code? Thanks.

src/reconstruction/refinement/addReaction.m
tpfau: No

src/reconstruction/refinement/addReaction.m
tpfau: I think it was simply a sanity check. you should not look for duplicates, if there are new Metabolites in the reaction.

src/reconstruction/refinement/addReaction.m
tpfau: we don't need it, and it can just be removed.

tutorials/modelCreation/ModelCreation.m
```diff
@@ -0,0 +1,110 @@
+%% Creating a Model 
+% To create a new model, there is a simple function ( createModel ):
+
```
laurentheirendt: May you run `initCobraToolbox` here?

tutorials/launchTutorials.m
```diff
@@ -38,7 +38,7 @@
     changeCobraSolver('glpk')
     sparseLP_example;
     uniformSampling;
-    tutorial_IO;
+    tutorial_IO;
```
laurentheirendt: May you also include `ModelCreation` and `ModelManipulation` here? If you can, it would be nice if you actually could rename the tutorials as `tutorial_modelCreation` and `tutorial_modelManipulation`.

tpfau:
updated the respective functions/renamed the tutorials and added them to launchtutorials

laurentheirendt:
Nice! Thanks @tpfau 🌵

PR #497, opened 2017-06-01
rumba files with proper headers

arichelle:
Creating a new PR as the existing rumba branch didn't recognize the file updates, will close the other PR concerning up and down regulation (PR #471)


src/modelAnalysis/rumba/pValDistForModelOverlap.m
```diff
@@ -0,0 +1,58 @@
...
+
+% Add in changed zero flux rxns %%% make sure this doesn't conflict with
+% later steps
+model1.rxns(end+1:end+length(ZeroRxns2))=ZeroRxns2;
+model2.rxns(end+1:end+length(ZeroRxns1))=ZeroRxns1;
+model1.points(end+1:end+length(ZeroRxns2),:)=0;
+model2.points(end+1:end+length(ZeroRxns1),:)=0;
+
+% make a list of reactions that are shared
+rxnsInCommon = intersect(model1.rxns,model2.rxns);
```
tpfau: shouldn't this automatically be all reactions, we just added all missings between model1 and model2, or do I miss something?

src/modelAnalysis/rumba/pValDistForModelOverlap.m
```diff
@@ -0,0 +1,58 @@
...
+model2.points = [];% clear the model points from memory
+Modpts = model1.points(findRxnIDs(model1,model1.rxns),1:numPnts);
+model1.points = [];% clear the model points from memory
+
+% get indicies for the shared rxns
+ModInd = findRxnIDs(model1,rxnsInCommon);
+CntrlInd = findRxnIDs(model2,rxnsInCommon);
+
+% Compute much the average median flux value for each reaction
+MedianFlux_m = median(abs(Modpts(ModInd,:)),2);
```
tpfau: Is it intentional, that the absolute value is taken here? If there is a large deviation from model1 to model2, where a lot of fluxes get reversed, this will lead to potentially a very small median difference if -50 100 -30 and 50 -100 30  are considered to be the same. for the median calculation.

src/modelAnalysis/rumba/rumba.m
```diff
@@ -0,0 +1,294 @@
+function [RUMBA_outputs, UpRegulated, DownRegulated, MetConnectivity1, MetConnectivity2] = rumba(model1, model2, completeModel, sampling, maxMetConn, RxnsOfInterest, GenesOfInterest, NormalizePointsOption, PValCuttoff, MaxNumPoints, LoopRxnsToIgnore, verboseTag)
+% RUMBA predicts which reactions significantly change their
```
tpfau: There seem to be a lot of optional parameters here. (checks for nargin <4).
A: These should be marked as optional inputs (and their defaults mentioned)
B: For more than 3 optional parameters I strongly suggest using a inputParser and allowing parameter/value pairs. It's otherwise very easy to get the wrong input at the wrong position without noticing it, if the input values are similar.

src/modelAnalysis/rumba/rumba.m
```diff
@@ -0,0 +1,294 @@
...
+% are in the sampled models
+model1 = addMissingReactions(model1,completeModel);
+model2 = addMissingReactions(model2,completeModel);
+
+% Normalize the sampled points by scaling by the net network flux (1 = by
+% net flux, 2 = by growth rate
+    [model1,model2] = normalizePoints(model1,model2,NormalizePointsOption,LoopRxnsToIgnore);
+
+% Rename genes to avoid incompatibilities
+model1.genes = regexprep(model1.genes,'[_\-]','');% remove underscores and hyphens
```
tpfau: see above about name conversion, and why are undescores problematic?

src/modelAnalysis/rumba/classifyRxns.m
```diff
@@ -0,0 +1,158 @@
...
+    end
+end
+
+ConnectedMet(ismember(ConnectedMet,Mets2Ignore)) = [];
+
+sampledModelPoint = sampledModel.points;
+sampledModel.points = [];
+NumPts = length(sampledModelPoint(1,:)); % number of samples
+
+if verboseTag
```
arichelle: yes, i will do that

src/modelAnalysis/rumba/pValDistForModelOverlap.m
arichelle: Yes, you right, but was initiallly put to be sure that the addMissingReactions subroutine has done properly its job... i will add a warning at this step if the intersect is not equal to model1/2.rxns ;)

src/modelAnalysis/rumba/pValDistForModelOverlap.m
arichelle: You right, I will think about and chat with Nate

src/modelAnalysis/rumba/rumba.m
arichelle: OK

src/modelAnalysis/rumba/pValDistForModelOverlap.m
arichelle: I discussed with Nate about that and actually it is intentional to take the absolute value as we consider gene-reactions pair and no just reaction up or down regulation. The idea is that the amount of enzymes required (then the gene expression) is dependent on the absolute value of the flux and not on it's directionality.

src/modelAnalysis/rumba/pValDistForModelOverlap.m
arichelle: warning added

src/modelAnalysis/rumba/pValDistForModelOverlap.m
tpfau: True, looking at it like this makes sense.

arichelle:
I think most of the comment have been adressed, do I need to perform any other modification?

src/modelAnalysis/rumba/rumba.m
tpfau: Could you use a inputParser here, I can easily imagine someone wanting to set the verbose output for a default input, at which point it would be necessary to provide all 8 other optional arguments with their correct default values just to get a verbose output. If you want help, just let me know.

arichelle: @tpfau, yes if you could add the inputparser will be great because I never used and it is far from straightforward for me how to do it efficiently... the 3 required inputs are model1, model2 and complete model,  all the other are optional
thanks

tpfau: Hi,
I implemented a inputParser for the method on my rumba branch ( commits: 771ff81 and  94917ea ), which could be cherry-picked onto your branch.
BUT: since your current branch is behind develop (i.e. not rebased recently), There is an issue with verifyModel on your current branch. I use the syntax of the updated one on the development branch, but that will mean that the current function will not work on your current branch unless you rebase.

rmtfleming:
@tpfau 
Please refrain from requesting or implementing any further input parsers. We have a large user base who are not familiar with this approach, the merits of which are not clear, except perhaps for terminal functions in a pipeline, like plotting. Until further notice, no further pr's with new input parsers will be merged.

PR #498, opened 2017-06-01
adding tutorial for optForce and fixing errors for gams functions

snmendoz:
[*I added the tutorial for running the functions in design/optForceGAMS. In this tutorial I show an example of Optforce with E coli. I noticed that this pull request cannot be merged automatically. This is probably because when I updated my cobra folder, 2 functions were missing (optForceWithGAMS.m and testOptForceSol). So, I added these functions to the folder manually and then I updated optForceWithGAMS.m (I added a few lines of code). I am uploading here the new optForceWithGAMS.m. Please erase the optForceWithGAMS.m and testOptForceSol I uploaded before and keep the ones I am uploading now*]


tutorials/optForceGAMS/optForceTutorialGAMS.m
```diff
@@ -0,0 +1,204 @@
...
+% increase the production of succinate? We will approach this problem in
+% this tutorial and we will see how each of the three steps are solved.
+
+% In this tutorial we will run steps 2 and 3 in GAMS and we will use MATLAB
+% to call GAMS. Therefore, GAMS must be intalled in the system before
+% running this turorial. After GAMS installation, we have to add tha path
+% where GAMS in located
+initCobraToolbox
+
+gamsPath = 'C:\GAMS\win64\24.8';
```
tpfau: Please don't use absolute FileNames and don't use operating system dependent code. 
GAMS could be installed elsewhere, I suggest using either a global variable, or to assume, that GAMS is set up and check whether it is (and if its not report an appropriate error).

src/design/optForceGAMS/optForceWithGAMS.m
```diff
@@ -0,0 +1,840 @@
...
+%export inputs to GAMS
+inputFolder = 'InputsOptForce';
+exportInputsOptForceToGAMS(model, {targetRxn}, mustU, mustL, minFluxesW, maxFluxesW, minFluxesM, maxFluxesM, k, nSets,...
+    constrOpt, excludedURxns, excludedLRxns, excludedKRxns, inputFolder)
+
+% if the user wants to generate a report, print results.
+if printReport; fprintf(freport, '\n------RESULTS------:\n'); end;
+
+%run optForce in GAMS.
+if verbose
```
tpfau: InputsOptForce/ looks like unix style file separation (is this intentional or should this be filesep instead of /) ?

tutorials/optForceGAMS/optForceTutorialGAMS.m
snmendoz: Ok. I will change it. After modifying the function, I have to generate a new pull request?

tpfau: No. As soon as you push something, your pull request gets updated.
This actually means: please do not push every individual commit you do, but only push once you have done everything you want to change. Otherwise, the testing system will start new test runs for each individual commit (and tends to break down) :)

laurentheirendt: Actually, the CI system would not break down, but would have a hard time catching up 😄

src/design/optForceGAMS/optForceWithGAMS.m
snmendoz: I am a little bit confusing. In line 522 I see "InputOptForce", not "InputOptForce/". Am I missing something? I tried to make the code independent of the operation system, but I don't know if I succeded.

tutorials/optForceGAMS/optForceTutorialGAMS.m
snmendoz: Ok. I'm working with MATLAB.devTools, so I think these will be the next steps
>contribute
>choose option 2 (open the feature)
>choose my feature (add-tutorialgams)
>modify the variable gamsPath to make it system independent
>contribute
>choose option 3 (update the feature)
>choose my feature (add-tutorialgams)

With these steps my pull request will be updated?

tpfau: yes

src/design/optForceGAMS/optForceWithGAMS.m
tpfau: Sry, the commen went to the wrong place.
I meant line 531 with --myroot=InputsOptForce/ . the / here could well be intentional and have nothing to do with OS things. I was just wondering.

snmendoz:
Thanks Laurent. I will answer below.

Q: We try to not distribute models in .matthrough this repository directly. Instead, may you add the line in retrieveModels.m? This works well if the model is already published somewhere.

A: The model is published in an article (https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2048574/), but if you see the paper you will notice that the model is written in the paper itself. I mean, I had to copy and paste the model to an excel file and then read the excel with matlab to convert the model to a matlab format. In fact, I the model is not even in the supplementary information, so I think there is no direct link to the model. I saw in retrieveModels.m that I have to add a web direction (http://....), but I think that there is no direct web link to the model. What should I do?

Q: Also, do we need the report files such as tutorials/optForceGAMS/TestOptForce/OutputsFindMustL/report-01-Jun-2017-19h-23m.txt?

A: That kind of reports could be very useful for users, because they are generated every time you run optForce. So, for example, the user can know which inputs were used two months ago to run optForce and which outputs were generated using that inputs. If the user do not want to show this report, he can enter the input option printReport = 0. 

I ran the tutorial for OptForce I during execution some folders (TestOptForce and TestOptForce2) and files where generated. I decided to include this information because this way, the user can see an example of the files that can be generated by the functions

I you want, maybe we just remove the examples (TestOptForce and TestOptForce2) that are generated when running the tutorial 

Q: It seems as if you have multiple times the file GtoM.gdx- is this always the same file? May you only include it once? 

A: GtoM.gdx is not always the same file. I use a package (GDXMRW) to write and read information which could be read by both MATLAB and GAMS. For example, when I run the function findMustLWithGams.m it calls the GAMS function findMustL.gms, which in turn, generates a file called GtoM.gdx. This file contains the solution to that particular bilevel optimization problem (must set of first order for downregulations). When I call another function, for example, findMustUUWithGams.m, a file called GtoM.gdx, is also generated but it contains a different kind of information inside (must sets of second order for upregulations)

Maybe, I could just change the name of  GtoM.gdx for each function, so a file with a different name will be generated for each function.

Q: Is this a binary file that is platform dependent? If so, then I suggest to put this file into COBRA.binary. 

A: Whay do you mean platform dependent? It should work for windows, linux and mac (always that you have matlab and gams installed)

Q: What about the optForce.lst files - may we only have 1 file? Are they automatically generated?

A: optForce.lst is autogenerated by gams. I decided to run the tutorial with all the options equal to "true" and that means store inputs and outputs for future revision. This way the cobra user can see examples for the inputs and outputs generated by my functions.

I you want, maybe we just remove the examples (TestOptForce and TestOptForce2) that are generated when running the tutorial

Q: Would you mind formatting the documentation as explained in the documentation guide? Thank you.

A: For documentation you mean the commentaries in the functions before the code? For example like this link: https://opencobra.github.io/cobratoolbox/latest/modules/design.html

If so, I will add usage and optional inputs sections, because I already add inputs and outputs. Some else I would be aware of?

Q: As for the conflict, don't worry - we will try to fix that one.

Ok, thanks.

src/design/optForceGAMS/optForceWithGAMS.m
snmendoz: This is intended. GAMS only read "/" not "\", independent of the system (I am wonking with windows)

snmendoz:
Laurent, why my function decrease the coverage according to Codecov Report?

tpfau:
Hi,
Files, which are generated during a testrun should (in general), not be included in a test (You might want to add these kind of files in a tutorial, if you want to provide users with an example output, but in tests, they should not be included. To some extend, they might be a reason, why the test coverage drops, as they are considered additional code which is, (obviously) not executed.  
In addition, tests should clean up after themselves if they are successfull, i.e. if a test function is run, and succeeds, any file generated during the test should be removed again. This should not happen, if there is an error (i.e. if there is an assertion, that is not fullfilled), but otherwise, these files are unnecessary and would just clutter a machine the test was run on. 
I hope this answers some of the questions.

tutorials/optForceGAMS/optForceTutorialGAMS.m
```diff
@@ -33,9 +33,13 @@
...
+% We check if gams was correctly installed and if GAMS folder was added
+% to MATLAB path
+fullGamsPath = which('gams');
+if isempty(fullGamsPath)
+   error('GAMS is not installed in your system or GAMS directory has not been added to MATLAB path') 
+end
+% Now, we can start with the modelling
 
 %% Modelling 
 % First we load the model. This model comprises only 90 reactions, which
```
tpfau: Model.rev is deprecated (use model.lb < 0 instead). Could you adjust the code accordingly, otherwise you will get errors.

tpfau:
About the model file:
Is it absolutely necessary to use this particular model? It, in general, looks very similar to the ecoli_core model, which is already downloaded automatically. Would it make sense to use the ecoli core model instead of the model you currently use for the tutorial?

src/design/optForceGAMS/optForceWithGAMS.m
```diff
@@ -0,0 +1,840 @@
...
+%                            Description: Minimum fluxes for each reaction
+%                            in the model for mutant strain
+%                            Example: minFluxesW=[-90; -56];
+%
+% maxFluxesM (obligatory):   Type: double array of size n_rxnsx1
+%                            Description: Maxmum fluxes for each reaction
+%                            in the model for mutant strain
+%                            Example: maxFluxesW=[92; -86];
+%
+% k(optional):              Type: double
```
tpfau: for more than a few optional arguments, It is highly recommended to use an input parser and allow parameters as Parameter/Value pairs. Otherwise code with many input arguments is prone to typos, which worst case lead to wrong input being used unnoticed.
Also: There is a 
OPTIONAL INPUT: 
flag to indicate optional input for the function that is used in the documentation generation.  (in general please follow the documentation [guideline](https://opencobra.github.io/cobratoolbox/docs/documentationGuide.html)

src/design/optForceGAMS/testOptForceSol.m
```diff
@@ -0,0 +1,116 @@
+function [maxGrowthRate, minTarget, maxTarget] = testOptForceSol(model, targetRxn, solution, relax, tol)
+%% DESCRIPTION
```
tpfau: This file (along with any files necessary for input to the function) should be moved to tests/verifiedTests/testOptForceWithGAMS
Otherwise it will not be recognized as a test function, and thus not be run in the test suite (which leads to the code coverage report)

tutorials/optForceGAMS/optForceTutorialGAMS.m
snmendoz: Ok. I will change the code accordingly.

src/design/optForceGAMS/optForceWithGAMS.m
snmendoz: I was looking for a guide like that! Thanks! I will look at it in detail. And I will change my codes in order to follow the guideline

src/design/optForceGAMS/testOptForceSol.m
snmendoz: Sorry for the misunderstanding. testOptForceSol is not a test function. This function analize if the solution found by optForce guarantee a minimum production of the target metabolite. I think I will change the name of the function to avoid misunderstandings

laurentheirendt: Thanks @snmendoz 👍

snmendoz:
Thank @tpfau ! for your answers. 

I included the output files for my tutorial because I think that it will be easier for the user to know how to use the functions if he has an example of the output files. I will add some comments in the tutorial to explicit the files that are generated during the execution of the tutorial. In despite of that, I think the best is not to include the ouputs in my pull request. Ronan asked me to clone my fork and work in just one branch. I will clone my fork again, I will modify functions according to your commentaries and I will make a pull request with my updated functions from another branch. 

About your proposal to use the ecoli core model, I can answer the following: This is the first time that someone implements optForce, so Ronan asked me to make sure that the algorithm is working properly. The only way to do that is to replicate the same example used for the group of Costas D. Maranas (author of OptForce) to run optForce, with the same inputs, and that includes using the model provided by them (AntCore). I think that the ecoli core model is very similar to AntCore model but If we use the ecoli core model I cannot ensure that we will get the same solutions found by the Costas D. Maranas group because even though the models are very similar, they are not the same model.

src/design/optForceGAMS/optForceWithGAMS.m
snmendoz: @tpfau @laurentheirendt 

I am changing the documentation of my functions according to the guideline. I have some functions that generate files during their execution. These files are very important for the users because results are saved in these files. In fact, I have some functions whose only purpose is to generate files and no output variables are returned by these functions. I couldn't find any rule in the guideline to describe such output files. I think I should describe such output files because they are important for the user but should I describe these files in the documentation? I so, should I put the description of these files in the OUTPUT box?

tpfau: What I would do is the following:
Textual descriptions of output files (small excerpts of file header and similar things) go into the OUTPUT and are provided in the tutorial. The actual file should be generated on the machine of a user and not be distributed via github. The reason is quite simple: if in a tutorial I get all outputs provided and I have no clue about matlab, I might think that something is working while it is actually not working. E.g. if my gams would not be working properly, I can easily imagine users going on with the tutorial if the output file is there and can be further processed without them actually having performed the steps. And later wondering, why it worked in the tutorial but not in their actual code.

rmtfleming:
how about a subfolder of the tutorial folder called 'anticipatedResults'? Would that remove the ambiguity? The actual results would be saved to some folder specified by the user, which should not be a subfolder of the toolbox. It is better to separate code from data and results. Here we are dealing with a grey area for tutorial material.

src/design/optForceGAMS/optForceWithGAMS.m
snmendoz: I agree. In my new pull request (https://codecov.io/gh/opencobra/cobratoolbox/pull/523?src=pr&el=desc) I removed all the files generated during the execution of functions. 

About the description of the output files, In the new PR you will see that I put the description of outputs files inside of each function, in the OUTPUT section. Additionally, In the tutorials, I added comments to highlight the lines in which the files are generated. Is that what you meant, isn't it?

tpfau: Yes, that is a good way to do it (in my opinion).

snmendoz:
Yes. We are dealing with a grey area. Let me know about your final decision about folders, so I can change the code of my tutorials accordingly. I think the idea of a new folder called "anticipatedResults" is actually very good. You can see the updated functions to run optforce here: https://codecov.io/gh/opencobra/cobratoolbox/pull/523?src=pr&el=desc

laurentheirendt:
@snmendoz, in my opinion, we should *not* include any auto-generated files in the repository. Instead, we should make sure that in the tutorial there is a section entitled *Anticipated Results* in which the output is described.

Many times, we have tests or tutorials that generate an output, but the (data) output is never saved to the repository. The only exceptions to that are the output of each command in the tutorial (text output in the `.pdf` and `.html` files) and reference data in tests.

May you remove all auto-generated files? We can also include a list of them in the `.gitignore` file at the root of the repository to avoid a user committing them after running the tutorial.

As for the `GtoM.gdx` files, it would be good if all files were named with a unique name. We aim at having no duplicate files or files with the same name in the repository. May you do that?

Including the `AntCore.mat` model in the repository of The COBRA Toolbox is really not desirable. Would you mind hosting the model in a own Github repository, e.g. www.github.com/snmendoz/AntCoreModel? You could also include there the script that you used to generate the model and a `README.md` file with a link to the publication. We could then use that URL in `retrieveModels.m`.

cc: @syarra

snmendoz:
@laurentheirendt 

I removed all the auto-generated files in the new PR #503. Should I remove the files in this PR too?

I've been describing outputs files in the sections "procedure". I will describe them in the section "Anticipated Results" as you suggest. 

Is that ok that the files are generated inside "cobratoolbox/tutorials/optforce" or should be generated outside "cobratoolbox/"

There is a file that is generated during optforce running. This file contains the report of the running and its name depends on the date. For example, If I run OptForce today, the name will be "report-12-june-2017...". I can imagine that it is impossible to add this file to .gitignore. Should I avoid putting the date in the file name?

I will rename all repeated files (e.g GtoM)

I will include AntCore (and all related files) to my git repository.

laurentheirendt:
@snmendoz, thanks for your swift response. I am not sure, what PR is that? May we actually somehow combine these PRs? I am a little bit lost - which files are new and which ones are not?

I recommend creating a new branch (using `contribute` and selecting `[1]`), and then only submit 1 PR with the code. Then, create a new branch, and submit a PR with the tutorial. Would that work for you? This would ease a lot the task of merging and solving possible conflicts. As of now, there will likely be many conflicts, as many files are repeated in the different PRs that are not exactly the same.

If you generate all files in a folder called (for example) `tutorials/optForce/generatedFiles`, we can exclude the contents of that folder in the `.gitignore` file.

snmendoz:
@laurentheirendt

Sorry. The new files are in the PR #523, not 503. In this PR you will find all the files for running optforce as well as the tutorials for optforce. In this PR all the autogenerates files were deleted. Thomas Pfau suggested some modifications to my functions and I modified the functions accordingly. 

The old ones (PR 498)

laurentheirendt:
So, if I understand correctly, can we close then this PR?

snmendoz:
Yes, I think that is easier for all of us. In that case, are the commits of PR #498 counted as a contribution?

laurentheirendt:
Is there anything in this PR #498 that is not already included in PR #523?

If we close this PR and there is nothing new compared to PR #523, your 2 commits of this PR will not be counted as a contribution.

laurentheirendt:
Closing as discussed. Thanks @snmendoz 👍

PR #501, opened 2017-06-05
Develop

dayena:
Splitting a specific list of reversible reactions to irreversible, rather than converting the entire model to irreversible. 


laurentheirendt:
The PR has been opened in order to merge `develop` into `master`. Not generally something a regular user would do - probably only a mistake. @dayena?

PR #503, opened 2017-06-05
Splitting a specific list of reversible reactions into irreversible.

dayena:


tpfau:
Hi, 
Please add this functionality to "convertToIrreversible" adding an optional parameter for the reactions (which is the whole model by default). At the moment, this is unfortunately a lot of code duplication, which will lead to unnecessary maintenance (when one function needs to be changed, both have to be adapted).

PR #507, opened 2017-06-06
Deprecate rev

tpfau:
Deprecating the .rev model field replacing it by a check for the lower bound < 0 whereever necessary. 
Needed to update some tests, due to contradicting lb and rev fields.

- [X] Tested my code on my own machine (2016b)


rmtfleming:
Thanks Thomas, depreciating model.rev will help everyone a lot, even if they do not realise it yet.

laurentheirendt:
Thanks @tpfau  ⚽️

PR #511, opened 2017-06-06
option added to convertToIrreversible.m 

dayena:
An option has been added to convertToIrreversible.m to allow the user to only split a selected list of reversible reactions, rather than the entire model reactions. Additional functions have been added to find reactions or metabolites from a given compartment. 


laurentheirendt:
As discussed.

PR #514, opened 2017-06-06
Addition of option to convertToIrreversible.m

dayena:
Adding the an option to convertToIrreversible to allow the user to ony convert a specific list of reversible reactions. Addition of two functions to explore the model by finding all reactions and metabolites in a specific compartment.

### DIFF
--- a/src/reconstruction/refinement/convertToIrreversible.m
+++ b/src/reconstruction/refinement/convertToIrreversible.m
@@ -1,13 +1,17 @@
-function [modelIrrev, matchRev, rev2irrev, irrev2rev] = convertToIrreversible(model)
-% Converts model to irreversible format
+function [modelIrrev, matchRev, rev2irrev, irrev2rev] = convertToIrreversible(model, sRxns)
+% Converts model to irreversible format, either for the entire model or for
+% a defined list of reversible reactions.
 %
 % USAGE:
 %
-%    [modelIrrev, matchRev, rev2irrev, irrev2rev] = convertToIrreversible(model)
+%    [modelIrrev, matchRev, rev2irrev, irrev2rev] = convertToIrreversible(model, sRxns)
 %
 % INPUT:
 %    model:         COBRA model structure
 %
+% OPTIONAL INPUTS:
+%    sRxns          List of specific reversible reactions to convert to
+%                   irreversible (Default = model.rxns)
 % OUTPUTS:
 %    modelIrrev:    Model in irreversible format
 %    matchRev:      Matching of forward and backward reactions of a reversible reaction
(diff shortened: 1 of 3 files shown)


src/reconstruction/refinement/convertToIrreversible.m
```diff
@@ -25,88 +29,184 @@
...
+%       false (forward) and model.rev = true (reversible).
 
-modelIrrev.S = spalloc(size(model.S, 1), 0, 2 * nnz(model.S)); % declare variables
+modelIrrev.S = spalloc(size(model.S,1),0,2*nnz(model.S)); %declare variables
 modelIrrev.rxns = [];
-modelIrrev.lb = zeros(2 * length(model.rxns), 1);
-modelIrrev.ub = zeros(2 * length(model.rxns), 1);
-modelIrrev.c = zeros(2 * length(model.rxns), 1);
-matchRev = zeros(2 * length(model.rxns), 1);
+modelIrrev.rev = zeros(2*length(model.rxns),1);
```
laurentheirendt: @dayena, `model.rev` has been deprecated in PR #507. May you adapt your code accordingly? Thank you.

cc: @tpfau

dayena: Hi Laurent! I made the necessary changes. Thank you!

laurentheirendt:
cc: @tpfau

tpfau:
Wouldn't it make more sense, to default to model.rxns for sRxns? I.e. when no reactions are specified, all reactions are used. That way, the code duplication could be avoided (and it would be the same code for both versions).

dayena:
Hi Thomas! 

I thought that this is already what is being done by adding the option of sRxns. If it is not empty, then the entire model.rxns is being used instead? 

Best wishes, 

Diana

tpfau:
Currently: You have something like:
```
if ~exist('sRxns','var')
 sRxns = {};
end

if ~isempty(sRxns)
do a lot of code for those reactions which are in sRxns
else
do a lot of code for all reactions.
end
```
What I'm suggesting is:
```
if ~exist('sRxns','var')
 sRxns = model.rxns;
end
do a lot of code for those reactions which are in sRxns
```
The reason is quite simple: Currently, if the function ever needs to be changed, corrections/alterations have to be done in 2 places (as otherwise the functionality is out of sync). Also, I would expect that the results, of convertToIrreversible(model,model.rxns) and convertToIrreversible(model), to be the same. but if I run it on e.g. ecoli_core, e.g. the sizes of the reactions are not the same. 
That's why I suggested to only have one set of code. If you want to avoid an "_r" attached to the reactions, I would suggest to introduce another flag.

dayena:
Hello Thomas, 

I see what you mean... if you could help me that would be great! Actually, all I wanted to do with the additional option of sRxns (for specific reactions) was to only split those reversible reactions into two irreversible ones and nothing else. I want to keep the bounds 0 to 1000 and I want to flip model.S. The reason why convertToIrreversible(model,model.rxns) doesn't give the same number of reactions as  convertToIrreversible(model) is because in the first case, we assume that all reactions are reversible (when in reality they are not! so such an input would not make sense). Please feel free to make any additional changes you think might further improve this code... All I want is to have an option to only and only split those reactions of interest, modifying nothing else in the model, except for those specific reactions... 

Best wishes, 

Diana

tpfau:
Ok, I'll have a look at it.
I already noticed, that there is a bug in the function leading to complete backward reactions not being marked properly. 
I'll address that along with your code. 
Just to make sure that this is how you want it:
The function will, when done, copy all rxn related fields to the irreversible model and only have one reaction for pure backward reactions (i.e. those are not duplicated). Any non reversible reaction (i.e. lb >= 0)  will not be modified.

dayena:
Hi Thomas, 

1) "copy all rxn related fields to the irreversible model": yes!
2) "only have one reaction for pure backward reactions (i.e. those are not duplicated)": no! pure backward reactions should not be modified at all, I want them to stay as in the original input model. 
3) "Any non reversible reaction (i.e. lb >= 0) will not be modified" : no! any irreversible and reversible reaction that is not in sRxns must not be modified. 

Thank you! From the changes I made, I am already getting what I want with the correct output, but if you have any ideas on how to shorten this code, please feel free to do so.

Best wishes, 

Diana

tpfau:
The second option is contradictory to the intent of convertToIrreversible, since an irreversible model should not have any negative fluxes. 
E.g. (but thats maybe only my opinion), the optimal value of a flux minimisation on an irreversible model should be the same as minimisation of sum of absolute flux values. 
Just out of curiosity: Why do you need backward reactions to stay backward? 
I can implement a flag, that ignores Backward Reactions, but this is kind of contradictory to the idea of making a model irreversible (again, this is maybe just my opinion).

rmtfleming:
In an irreversible model, a flux can only be zero or positive.

tpfau:
Please have a look at #527 which should introduce the options you want (for your desired result the call would be: convertToIrreversible(model,'sRxns',reactions,'leaveInverse',true)
reactions would be the set of reactions you want to split and leaveInverse would indicate that any purely backward reaction would be ignored in the process.
I based the pr on your branch, so this one can be closed.

laurentheirendt:
Closing in favor of PR #527. Thanks @tpfau.

Issue #515, opened 2017-06-06
Duplicate fastcore.m function


laurentheirendt:
@jacekwachowiak just discovered that in PR #447, `fastcore.m` has been added although an existing function `fastcore.m` was already present.

- [original version](https://github.com/opencobra/cobratoolbox/blob/develop/src/dataIntegration/transcriptomics/FASTCORE/fastcore.m)
- [new version](https://github.com/opencobra/cobratoolbox/blob/develop/src/dataIntegration/transcriptomics/fastcore.m)

@arichelle, may you have a look at that? Thanks.

cc: @rmtfleming, @tpfau, @syarra

arichelle:
I see, the original one that i've submitted included in its core the definition of "expressionRxns" and "threshold" which are now defined in createTissueSpecificModel  then the lines 45 to 52 in the original have been removed for the new version ... it is the only difference between the two from the code part (just saw that the header function in the new one are not properly formatted...)

laurentheirendt:
@arichelle, would you mind taking care of this? Thanks 👍

arichelle:
You want that I update the header of the new version ?

rmtfleming:
The original file should remain and a diff done to add functionality to it if necessary, otherwise delete the new duplicate one. Thanks.

arichelle:
OK will do that

arichelle:
I've update the files but src/dataIntegration/transcriptomics/fastcore.m need still to be removed as it seems like I cannot do it myself using devTools

laurentheirendt:
Fixed by PR #517. Thanks @arichelle!

PR #517, opened 2017-06-07
factore update header

arichelle:
Update of fastCore.m to be compatible with createTissueSpecificModel, update header and remove duplicate


arichelle:
I've try to remove the duplicate (src/dataIntegration/transcriptomics/fastcore.m) using devTools but it does not want to commit the deletion...

laurentheirendt:
Fixes issue #515.

laurentheirendt:
@arichelle, thanks. I removed the file now. What exactly was the issue with the devTools?

Would you also mind formatting the documentation here? @jacekwachowiak had done such a nice job already. A documentation guide is [here](https://github.com/opencobra/cobratoolbox/blob/master/docs/source/guides/documentationGuide.md). Thanks.

src/dataIntegration/transcriptomics/FASTCORE/fastcore.m
```diff
@@ -1,65 +1,51 @@
...
+    if nargin < 3 || isempty(epsilon)
+        epsilon=1e-4;
     end
 
+    coreSetRxn = core;    
     model_orig = model;
 
     %Find irreversible reactions
-    irrevRxns = find(model.lb>=0); %% could be called irrevRxns
+    irrevRxns = find(model.rev==0);
```
laurentheirendt: @arichelle, I think that @tpfau deprecated `model.rev` recently in PR #507. Also, in PR #513, @tpfau already had fixed this specific line. 

May you revert back to what he had changed? Thanks.

arichelle: Sorry didn't catch that

Issue #343, opened 2017-03-27
some functions (like extractSubNetwork) miss model fields


akaviaLab:
*(Note: You may replace [ ] with [X] to check the box)*

**I hereby confirm that I have:**

- [X ] Checked that a similar issue hasn't already been filed

extractSubNetwork (and possibly other functions) have a problem where the model fields are coded in the function.
if (isfield(model,'geneNames'))
    subModel.geneNameRules = model.geneNameRules(selRxns);
    subModel.geneNames = model.geneNames(selGenes);
end
if (isfield(model,'subSystems'))
    subModel.subSystems = model.subSystems(selRxns);
end
if (isfield(model,'rxnNames'))
    subModel.rxnNames = model.rxnNames(selRxns);
end
if (isfield(model,'rules'))
    subModel.rules = model.rules(selRxns);
end

The problem is that in case of models that have additional fields (such as RECON 2.04), the functions won't change them.
removeRxns, lines 63-73 defines the fields to process, but we could expand the defintion to a global struct or function, or maybe part of every model

Then we could intersect with current model, and have code that would look something like:
modelOut.(rfields{i}) = model.(rfields{i})(selectRxns, :) for removing/selecting reactions and so forth.

If we can agree on a way of doing this, I'm willing to go over the code, find cases, and make sure they all use the updated version. I keep tripping on this when I use new functions, and then miss certain fields or something.

tpfau:
This is part of an issue, we hope to address soonish (essentially, that additional model fields cannot clearly be associated with specific required fields). 
@syarra @laurentheirendt @rmtfleming 
I like the idea of a model field defining fields, which could be used to properly update fields and is flexible wrt future changes.

laurentheirendt:
Thanks for solving this issue here, @tpfau! 👍

PR #520, opened 2017-06-07
Update to functions modifying field sizes

tpfau:
This is an update to functions modifying field sizes. Associated field sizes are now updated automatically (if they fit to some rules, either same siz, or if multiple base fields with the same size exist, fitting to e.g. rxnXYZ for rxns, or being defined in the modelFields).
This is addressing issue #343


src/base/io/utilities/creategrRulesField.m
```diff
@@ -16,7 +16,7 @@
 % .. Authors: - Thomas Pfau May 2017
 
 
-
+modelWithField = model;
```
laurentheirendt: Do we need this deep copy here?

src/reconstruction/fastGapFill/AuxillaryFiles/createXMatrix2.m
```diff
@@ -224,5 +224,6 @@
     end
     showprogress(i/length(compounds));
 end
-
-ExchangeRxnMatrix.mets = ExchangeRxnMatrix.mets';
+%Currently we are likely to have a largely blown up S matrix.
+ExchangeRxnMatrix.S = ExchangeRxnMatrix.S(1:numel(ExchangeRxnMatrix.mets),1:numel(ExchangeRxnMatrix.rxns));
+%ExchangeRxnMatrix.mets = ExchangeRxnMatrix.mets';
```
laurentheirendt: Do we still need this line?

src/reconstruction/fastGapFill/AuxillaryFiles/createXMatrix2.m
```diff
@@ -224,5 +224,6 @@
     end
     showprogress(i/length(compounds));
 end
-
-ExchangeRxnMatrix.mets = ExchangeRxnMatrix.mets';
+%Currently we are likely to have a largely blown up S matrix.
+ExchangeRxnMatrix.S = ExchangeRxnMatrix.S(1:numel(ExchangeRxnMatrix.mets),1:numel(ExchangeRxnMatrix.rxns));
```
laurentheirendt: Why exactly is this done here?

src/reconstruction/refinement/addReaction.m
```diff
@@ -317,7 +317,8 @@
     else
         try
         model.S = [Stmp Scolumn];
-        model = updateReactionFields(model);
+        model = updateRelevantModelFields(model,'rxns');
+        %model = updateReactionFields(model);
```
laurentheirendt: Do we still need this line here?

src/reconstruction/refinement/addReactionGEM.m
```diff
@@ -69,6 +69,8 @@
 end
 
 nMet = length(model.mets)+1;
+orignMets = numel(model.mets);
```
laurentheirendt: Do we still need `nMet`, or why could `origMets` not be defined as a function of `nMet`?

src/reconstruction/refinement/changeGeneAssociation.m
```diff
@@ -78,12 +80,12 @@
                 grRule = regexprep(grRule,[genes{i} ')'],[newGene ')']);
                 genes{i} = newGene;
             else
-                warning(['Gene name ' genes{i} ' not in translation list']);
+                fprintf(['Gene name ' genes{i} ' not in translation list']);
```
laurentheirendt: May you add here a line break in `fprintf`?

src/reconstruction/refinement/changeGeneAssociation.m
```diff
@@ -78,12 +80,12 @@
...
                 genes{i} = newGene;
             else
-                warning(['Gene name ' genes{i} ' not in translation list']);
+                fprintf(['Gene name ' genes{i} ' not in translation list']);
             end
         end
         geneID = find(strcmp(model.genes,genes{i}));
         if (isempty(geneID))
-            warning(['New gene ' genes{i} ' added to model']);
+            fprintf(['New gene ' genes{i} ' added to model']);
```
laurentheirendt: May you add here a line break in `fprintf`?

src/reconstruction/refinement/mergeTwoModels.m
```diff
@@ -26,205 +28,202 @@
...
+%Making the comps fields unique
+if(isfield(modelNew, 'comps'))
+    [ucomps,ia,ic] = unique(modelNew.comps);
+    toKeep = true(size(modelNew.comps));
+    toKeep(ia) = false;
+    modelNew = removeRelevantModelFields(modelNew,~toKeep,'comps');
+end
+
+
+function modelNew = mergeFields(modelNew,model1,model2,type)
```
laurentheirendt: May you document as well the inputs/outputs here?

src/reconstruction/refinement/removeUnusedGenes.m
```diff
@@ -11,21 +11,18 @@
 %
 % OUTPUT:
 %		 model:    COBRA model structure with updated gene field reflecting only
-%							 content present in the model
+%                  content present in the model
```
laurentheirendt: The formatting does not see right here. Is the output argument `modelNew`?

test/testAll.m
```diff
@@ -22,12 +21,13 @@
...
-    addpath(genpath(pwd));
+    cd([CBTDIR, filesep, '..', filesep]);    
+else
+    CBTDIR = fileparts(which('initCobraToolbox.m'));
+    cd(CBTDIR);
 end
-
+%include the root folder and all subfolders.
+addpath(genpath(pwd));
 CBTDIR = fileparts(which('initCobraToolbox.m'));
```
laurentheirendt: Is there still a need to redefine `CBTDIR` here?

src/reconstruction/fastGapFill/AuxillaryFiles/createXMatrix2.m
tpfau: Because the ExchangeMatrix S is initialized as a VERY large matrix. 
If you look at lines 55ff, this is a space allocation, and  actually initializes the size of the matrix.
This leads to ExchangeRxnMatrix (which is created as a COBRA model struct having a size that is inconsistent with the actual number of reactions/metabolites).

src/base/io/utilities/creategrRulesField.m
tpfau: No, I'll change this to just have model.

src/reconstruction/fastGapFill/AuxillaryFiles/createXMatrix2.m
tpfau: no

src/reconstruction/refinement/addReaction.m
tpfau: No, actually we don't need the nested function any more neither

src/reconstruction/refinement/addReactionGEM.m
tpfau: orignMets is the number of mets at the start. nMets is modified during the function, so we need 2 separate values because to properly update the remaining met fields, we need to know what the original size was.

src/reconstruction/refinement/changeGeneAssociation.m
tpfau: done

src/reconstruction/refinement/changeGeneAssociation.m
tpfau: done, thanks for catching those.

src/reconstruction/refinement/mergeTwoModels.m
tpfau: done

src/reconstruction/refinement/removeUnusedGenes.m
tpfau: yes modelNew is the output. There were some tabs left in the file. removed them and updated the docu

test/testAll.m
tpfau: yes, since CBTDIR otherwise is the /test dir.

src/reconstruction/refinement/addReactionGEM.m
laurentheirendt: Isn't `length(model.mets)` always equal to `numel(model.mets)`? Couldn't we define `orignMets = nMet + 1`?

I am just curious why once `length` and once `numel` is used for the same purpose.

test/testAll.m
laurentheirendt: I am not sure that it is a good idea to redefine `CBTDIR` with an addition of `/test`. Whenever `/test` is needed, could we not have `[CBTDIR filesep 'test']`? 

Simplifies maintenance when `CBTDIR` always points to the root.

src/reconstruction/fastGapFill/AuxillaryFiles/createXMatrix2.m
laurentheirendt: Is there actually a pre-allocation memory check done as well? Or is the code running until the memory is full?

src/reconstruction/refinement/addReactionGEM.m
tpfau: length/numel is the same for a vector. and admittedly length is preferable. Its just an old habit of mine to use numel, as it does not hide the other dimensions. length gives you the largest dimension (essentially max(size(x))) while numel gives yout the multiplication (essentially prod(size(x))).
As for orignMets:
You could add multiple metabolites in one call of addReactionGEM (having multiple metabolites in the reaction not yet in the model), so while yes, there is a direct link between orignMets and nMet (namely the number of added metabolites), this would require an additional call. 
As mentioned, nMet is modified later on to know the current number of metabolites in the model, and we need to update the fields for all of those. And its faster if we only do this once and not multiple times.

test/testAll.m
tpfau: agreed, but the assignment of CBTDIR to the test folder was not my code ^^ 
I'll change it anyways. (along with folder changes)

src/reconstruction/refinement/addReactionGEM.m
tpfau: Ahh ok, now I get it. yes, you are right, we could simply say "nMet - 1".. updating it.

src/reconstruction/fastGapFill/AuxillaryFiles/createXMatrix2.m
tpfau: I have no idea. 
its just initialized using spalloc amd I assume, its initialized in a way that allows the code to run in any instance. I was just seeing that there was a size discrepancy which leads to problematic code, and inconsistent models being passed into functions (that admittedly originally didn't care about these things, as they only relied on mets and rxns as size indications....)

laurentheirendt: Let's keep that in mind for later ...

test/testAll.m
laurentheirendt: Thanks @tpfau 😄

laurentheirendt:
Thanks @tpfau 🎉

PR #523, opened 2017-06-07
adding updated optForce functions

snmendoz:
[*I updated my tutorials following the guideline provided by Ronan F. I modified the documentation section in all my functions according to the guideline provided by Thomas P. and Laurent H. I removed deprecated lines according to Thomas P. comments. I removed autogenerated files as suggested by Laurent. I renamed the function testOptForceSol to analizeOptForceSol*]


src/design/optForce/optForce.m
```diff
@@ -0,0 +1,1000 @@
...
+    error('OptForce: No model specified');
+else
+    if ~isfield(model,'S'), error('OptForce: Missing field S in model');  end
+    if ~isfield(model,'rxns'), error('OptForce: Missing field rxns in model');  end
+    if ~isfield(model,'mets'), error('OptForce: Missing field mets in model');  end
+    if ~isfield(model,'lb'), error('OptForce: Missing field lb in model');  end
+    if ~isfield(model,'ub'), error('OptForce: Missing field ub in model');  end
+    if ~isfield(model,'c'), error('OptForce: Missing field c in model'); end
+    if ~isfield(model,'b'), error('OptForce: Missing field b in model'); end
+end
```
tpfau: You could use the verifyModel function with the set of required fields as option:
verifyModel('requiredFields',{'S','rxns','mets','lb','ub','c','b'}). This would at the same time check, whether these fields (and any other fields) are consistent with the definitions.

src/design/optForce/optForce.m
```diff
@@ -0,0 +1,1000 @@
...
+    targetRxn, mustU, mustL, minFluxesW, maxFluxesW, minFluxesM, maxFluxesM, k,...
+    nSets, constrOpt, excludedRxns, runID, outputFolder, outputFileName,... 
+    printExcel, printText, printReport, keepInputs, verbose)
+% This function runs the third step of optForce that is to solve a
+% bilevel mixed integer linear programming problem to find sets of
+% interventions that lead to an increased production of a particular target
+%
+%
+% USAGE: 
+%
```
tpfau: If possible, with this many optional inputs, I would really recommend an [input parser](https://nl.mathworks.com/help/matlab/ref/inputparser-class.html), otherwise this is extremely prone to typos. e.g. when a user wants a verbose output he needs to give inputs for everything else.

src/design/optForce/analizeOptForceSol.m
```diff
@@ -0,0 +1,131 @@
+function [maxGrowthRate, minTarget, maxTarget] = analizeOptForceSol(model, targetRxn, solution, relax, tol)
```
tpfau: Just a hint:
When renaming files:
use git mv originalFileName newFileName
That way, the files get tracked and its much easier to check earlier changes.

snmendoz: I am using MATLAB.devTools as suggested by Ronan. Exist that in MATLAB.devTools?

src/design/optForce/optForce.m
snmendoz: Yes, I understand your point. I will try to change it.

src/design/optForce/analizeOptForceSol.m
laurentheirendt: @tpfau, the `src/design/optForce/analizeOptForceSol.m` file is a new file.

For your information, the MATLAB.devTools recognise when a move of a file is made (e.g. just happened on PR #525).

tpfau: Will you have to "publish" after moving but before editing the file?
Or how does it detect the file move? 
because analizeOptForceSol is the original testOptForceSol file.

laurentheirendt: Yes, generally you will have to run `contribute` [3] for a simple file move and then only do modifications in the file and re-run `contribute` [3] to publish the changes. This is standard `git` behaviour though. 

You would have 2 commits (1 move and 1 change commit) then, which is easier to track for reviewers.

src/design/optForce/optForce.m
snmendoz: @tpfau . The function verifyModel didn't work. 

I ran:
`>> verifyModel(model,'simpleCheck','requiredFields',{'S','lb','ub','c'})`

and I got:

Error using cell/ismember (line 34)
Input A of class cell and input B of class char must be cell arrays of strings, unless one is a string.

Error in verifyModel (line 51)
if any(ismember(varargin,'requiredFields'))

**I think the error is that the third input of varargin ({'S','lb','ub','c'}) is not a char.**

src/design/optForce/optForce.m
snmendoz: @tpfau I changed my codes and now I use varargin to handle inputs

snmendoz: @laurentheirendt @tpfau Please let me know which other changes I should make to merge this branch

laurentheirendt: @snmendoz, we have to merge PR #498 first. I only solved that conflict yesterday afternoon.

In order to speed things up, try to make smaller PRs. For instance, PR #498 has close to 1 million lines(!) to be added. Also, if you could, try to publish to the same feature (branch) when modifying the same files, without duplicating code in another PR or submitting another PR with similar files when a PR is already open. It makes it easier for us to review the contribution. Thanks.

*(Note to myself: provide these as guidelines on how to efficiently use the MATLAB.devTools.)*

src/design/optForce/optForce.m
tpfau: Thanks for spotting this. 
I updated the function in a new PR, changing to Parameter/value pairs, so the call, once its merged and your code rebased would now be:
`>> verifyModel(model,'simpleCheck',true,'requiredFields',{'S','lb','ub','c'})`

laurentheirendt:
@snmendoz, I rebased this branch now. You can now try running `verifyModel` with the version that @tpfau fixed in PR #529.

src/design/optForce/optForce.m
tpfau: Btw:
Thanks a lot for that change (I think it will make usage of the function much easier).

laurentheirendt:
@snmendoz, if you remove `tutorials/optForceGAMS/AntCore.mat` as discussed in PR #498, and put include the model URL in `retrieveModels.m`, we are ready to merge this PR here. 🎉

syarra:
@snmendoz, I rebased your commits and fixed the conflicts. I am merging right away.

PR #525, opened 2017-06-08
Test for metaboTools tutorial I

maikeaurich:


laurentheirendt:
Very nice @maikeaurich !

However, timing on the CI is 3651 seconds. 

Let's see how we can reduce that time.

test/verifiedTests/testMetaboToolsTutorialI/testMetaboTools_tutorialI.m
```diff
@@ -0,0 +1,371 @@
...
+    solverLPOK = changeCobraSolver(solverPkgs{k}, 'LP', 0);
+    solver = solverPkgs{k};
+    if solverLPOK
+        
+        
+        
+        %% Step *0* Define variables
+        
+        outputPath = pwd;
+        solver = solverPkgs{k};
```
tpfau: why do you change the solver twice here?

test/verifiedTests/testMetaboToolsTutorialI/testMetaboTools_tutorialI.m
```diff
@@ -0,0 +1,371 @@
...
+        ovB =0.9228;
+        assert(FBAA.f - ovA < toltests); % tol must be defined previously, e.g. tol = 1e-6;
+        assert(FBAB.f - ovB < toltests); % t
+                
+        
+        %% Step *5* setSemiQuantConstraints
+        % This function applies the constraints to the models. It takes two
+        % condition specific models into consideration.
+        
+        toltests = 1.0000e-04; % reduce for FBA results.
```
tpfau: this is actually an increase of the tolerance :)

test/verifiedTests/testMetaboToolsTutorialI/testMetaboTools_tutorialI.m
```diff
@@ -0,0 +1,371 @@
...
+            7367
+            8399
+            23545
+            54363
+            66002
+            129807
+            221823
+            ];
+        
+        [model_B_GE] = integrateGeneExpressionData(model_B_BM, dataGenes);
```
tpfau: Just out of curiosity: 
Is it intentional, that this only works for Recon1 and only if all model.genes fields match [0-9]+\.[0-9]+ ? 
Wouldn't it be better to adjust integrateGeneExpressionData to actually be able to handle other models as well?

maikeaurich: No, it is not intentional, just my limited horizon ;). I agree, it would be better.

test/verifiedTests/testMetaboToolsTutorialI/testMetaboTools_tutorialI.m
maikeaurich: I reduced the risk that the test fails!

test/verifiedTests/testMetaboToolsTutorialI/testMetaboTools_tutorialI.m
tpfau: could you cherry-pick my commit on https://github.com/tpfau/cobratoolbox/commit/13b5128df01ca0aa2773ddf22426464b1156824a that actually allows using non-numeric gene sets to be used in integrateGeneExpressionData ?
It still works with the old numeric stuff from Recon1.

laurentheirendt: Done.

syarra:
Temporarily closing as taking too much time to run.

PR #529, opened 2017-06-09
Update to verifyModel - Param/Value pairs

tpfau:
Needed to change the function to use parameter/value pairs, as otherwise requiredFields is incompatible with other options (ismember cannot handle anything but strings). 

The bug was spotted thanks to @snmendoz. 


tpfau:
Further corrections to add output information.

laurentheirendt:
As discussed. Thanks @tpfau for fixing this! ⚽️ 

@snmendoz, I will rebase some of your PRs to take this into account, notably PR #523.

PR #533, opened 2017-06-09
update minimizeModelFlux

arichelle:
Modification of minimizeModelFlux.m to set minNorm=1 +  moving the function in modelAnalysis folder


arichelle:
The minimizeModelFlux.m present in external/RAVEN still need to be deleted... as told to Laurent, I've trouble performing script deletion using devTools.

laurentheirendt:
@arichelle, may you check if my latest commit is what you intended? Thanks.

*FYI: I really can't reproduce (yet) that other error in the devTools.*

arichelle:
Yes, it is the good script deletion! 
Thank you @laurentheirendt

PR #534, opened 2017-06-09
fixing bug for handling inputs in function convertToIrreversible

snmendoz:
[*This function is used by function OptKnock. The function assumed that the fields .grRules and .rules has to be present in the model if the field .genes is present. iJO1366 from BIGG database has the fields .genes and .rules and hasn't the field .grRules. When I run OptKnock with this model (for the tutorial), the running chash. Therefore, I separated the handling of each field (.genes, .grRules, etc.)*]


laurentheirendt:
@tpfau: PR #527

tpfau:
This is covered in the new version of convertToIrreversible.
ONe question though:
Does OptForce (like OptKnock) rely on the order of reactions in the irreversible model to be forward reaction directly followed by backward? If so, please use `convertToIrreversible(model,'orderReactions', true)`.

PR #535, opened 2017-06-09
New version of createTissueSpecific

arichelle:
New version without default processing of data and update of some of the method calls to be in agreement with this new version

Note : I think it could be interesting to propose some script to facilitate the preprocessing by the user (e.g. parser for GPR rules, etc...), will add that in a new PR asap.


src/dataIntegration/transcriptomics/FASTCORE/fastcore.m
```diff
@@ -1,5 +1,5 @@
 function tissueModel = fastcore(model, core, epsilon, printlevel)
-% Use the FASTCORE algorithm ('Vlassis et al, 2014') to extract a context
+% Use the FASTCORE algorithm (Vlassis et al, 2014*) to extract a context
```
laurentheirendt: It seems like you are having an old version here, no?

src/dataIntegration/transcriptomics/FASTCORE/fastcore.m
```diff
@@ -45,7 +45,7 @@
     model_orig = model;
 
     %Find irreversible reactions
-    irrevRxns = find(model.rev>=0);
+    irrevRxns = find(model.rev==0);
```
laurentheirendt: I guess that this is an old version ...?

arichelle: Arf, I always forget to keep that :/ really sorry

laurentheirendt: May you also keep the header of the function that you already formatted? Thanks 😄

laurentheirendt:
@arichelle, the original `INIT.m` version has been moved to the `src` folder as discussed.

Is the version of `INIT.m` in this PR a new one?

arichelle:
The version of INIT is the same but i've modified some options to fit with the new version of createTissueSpecific

tpfau:
Hi  @arichelle, thanks a lot for this. :tada: 
The only thing that will need some change is the documentation (as we don't currently have additional parameters being parsed automatically). 
The easiest (for now) would be to just change it to:
"NOTE: "
because thats a known field for the documentation parser.

src/dataIntegration/transcriptomics/FASTCORE/fastcore.m
tpfau: Btw: this shouldn't use rev but lb >= 0

arichelle:
I think I've performed all the requested changes, let me know if I need to perform any other modification

tpfau:
Should be fine I think.

laurentheirendt:
Thanks @arichelle and @tpfau 🌵

Issue #545, opened 2017-06-12
Using box constraints (+-1000) versus using infinity constraints (+-inf) for unknown bounds can lead to different results


Sergio-Garcia:
For example, consider model_M with a large scalar M for unknown bounds:
``` Matlab
model_INF = model_M; 
model_INF.lb(model_M.lb ==-M) = -inf;
model_INF.ub(model_M.ub ==M) = inf;
```

The outcome of `findBlockedReaction` for model_M is different than that of model_INF (in my case the later having more blocked reactions). The reason behind this is that when the LP solver maximizes/minimizes those reactions with unknown bounds, which also happen to be unbounded, it returns a status of unbounded which the `fluxVariablity` function treats as 0. In fact, the following code solves the problem (i.e. the blocked reactions are the same in both models), but the issue can lead erroneous calculations if someone is unaware of this:
``` Matlab
model_INF = model_M;
[minFlux,maxFlux] = fluxVariability(model_M,0);
model_INF.lb(model_M.lb ==-M & minFlux ~=-M) = -inf;
model_INF.ub(model_M.ub ==M & maxFlux ~=M) = inf;
```

**I hereby confirm that I have:**

- [y] Tried to solve the issue on my own
- [y] Retried to run my code with the latest version of The COBRA Toolbox
- [y] Checked that a similar issue has not already been opened

rmtfleming:
Thanks for contributing this Sergio.

PR #547, opened 2017-06-13
Check for unboundedness/non optimal solutions

tpfau:
Addressing #545 
FVA now checks whether the solution status is 2 (i.e. unbounded). If the solution is unbounded, its obvious that maxFlux/minFlux are inf / -inf 

### DIFF
--- a/src/base/solvers/solveCobraLP.m
+++ b/src/base/solvers/solveCobraLP.m
@@ -1234,32 +1234,31 @@ switch solver
                 x = ILOGcplex.Solution.x;
                 w = ILOGcplex.Solution.reducedcost;
                 y = ILOGcplex.Solution.dual;                                            
-            elseif origstat == 4
+            elseif origStat == 4
                 %This is likely unbounded, but could be infeasible 
                 %Lets check, by solving an additional LP with a bounded
                 %objective.
-                ILOGcplex.Model.obj   = osense*c;
-                ILOGcplex.Model.lb    = lb;
-                ILOGcplex.Model.ub    = ub;
-                ILOGcplex.Model.A     = [LPproblem.A;osense*c];                
-                ILOGcplex.Model.lhs   = [b_L;-100* max(abs([lb(abs(lb) < inf);ub(abs(ub)<inf)]))];
-                ILOGcplex.Model.rhs   = [b_U;100* max(abs([lb(abs(lb) < inf);ub(abs(ub)<inf)]))];
+                %Store the original solution
+                Solution = ILOGcplex.Solution;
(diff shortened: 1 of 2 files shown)


src/base/solvers/solveCobraLP.m
```diff
@@ -1228,13 +1228,59 @@
...
                 x = ILOGcplex.Solution.x;
                 w = ILOGcplex.Solution.reducedcost;
-                y = ILOGcplex.Solution.dual;
+                y = ILOGcplex.Solution.dual;                                            
+            elseif origStat == 4
+                %This is likely unbounded, but could be infeasible 
+                %Lets check, by solving an additional LP with a bounded
+                %objective.
+                %Store the original solution
+                Solution = ILOGcplex.Solution;
```
syarra: I suggest we print out a message to the user that the status is `4` and that the problem is likely unbounded rather than solving another optimisation problem. We can suggest how to check for unboundedness or for infeasibility.

tpfau: To be honest:
This is not an option, or rather, we would than have to add another option to do this or not to do this.
The problem arises in small networks, where the pre-solver already determines, that this is unbounded/infeasible, which CPLEX doesn't distinguish. So we actually never "solved" the problem, but only ran a presolver on it.
If we give a warning, than e.g. fva can't distinguish whether its unbounded or infeasible, and we would either have to test this again there and handle different options at all code places, where those make a difference. 
This return value is also very rare for any complex problem, so its not a computational time problem (in my opinion).

rmtfleming: Primal unbounded = dual infeasible. The primal problem (non-zero flux) is still feasible.

tpfau: Yes, but the pre-solver doesn't indicate which of it it is, so we need to actually check it.

rmtfleming: If this is within FVA, we know it is feasible if at least one previous solve has been feasible.

tpfau: I know, but this is only on one particular solver and the return value was 4 (indicating something we don't know, since we only have -1 / 0 / 1 / 2 / 3 as defined return values). So, in this instance (where we have ibm_cplex and a solution status of 4, we know this is unbounded). But in other situations we might not know the reason for this undefined status 4, and someone might simply check for infeasibility and if its not infeasible (i.e. not 0), do something, which then might be wrong. 
I would prefer not to check every single instance for every individual solver but do this in solveCobraLP, and give a correct status.

syarra: I understand your point.

PR #550, opened 2017-06-15
LP support for C*v<=d and LP-QP call obj=0 fix, for nnz(c)>1

rmtfleming:

- [x ] Selected `develop` as a target branch (top left drop-down menu)


src/modelAnalysis/FBA/optimizeCbModel.m
```diff
@@ -368,7 +429,7 @@
 
         if isfield(solution,'dual')
             if ~isempty(solution.dual)
-                solution.dual=solution.dual(1:m,1);
+                solution.dual=solution.dual(1:nMets,1);
```
tpfau: Shouldn't the dual also contain the duals for the additional constraints?

src/modelAnalysis/coupling/constrainRxnListAboveBound.m
```diff
@@ -0,0 +1,54 @@
...
+
+[nMet,nRxn]=size(model.S);
+
+% Identify the indices of these reactions in the new model
+rxnInd = findRxnIDs(model, rxnList);
+
+modelConstrained=model;
+C=sparse(1,nRxn);
+if ~exist('c', 'var')
+    modelConstrained.C(1,rxnInd)=1;
```
tpfau: Could you update this so that subsequent calls to this function update the model adding rows to the Constraint matrix? 
I know, you can provide a C matrix, but for users it would be much more convenient, to just add all their constraints one after another).

src/modelAnalysis/coupling/constrainRxnListAboveBound.m
```diff
@@ -0,0 +1,54 @@
+function modelConstrained=constrainRxnListAboveBound(model,rxnList,c,d,csense)
+%
+% Constrains a (weighted) sum of absolute fluxes to be above a lower bound.
+%
```
tpfau: Isn't whats added just a sum of fluxes (i.e. Sum v_i < k) and not absolute ie Sum |v_i|  < k?

src/modelAnalysis/FBA/optimizeCbModel.m
```diff
@@ -194,40 +199,85 @@
...
+        if nIneq~=mltd
+            error('For the constraints C*v <= d, the number of rows of C and d are inconsisent')
+        end
+        if nltd~=1
+            error('For the constraints C*v <= d, d must have only one column')
+        end
+        LPproblem.A = [model.S;model.C];
+        % Fill in the RHS vector if not provided
+        if ~isfield(model,'b')
+            warning('LP problem has no defined b in S*v=b. b should be defined, for now we assume b=0')
```
rmtfleming: addRatioReaction should be moved to this coupling folder then we ultimately need a tutorial explaining how to use them. Each should make C*v <=> d constraints that optimizeCbModel recognises.

rmtfleming: csense means constraint sense, not constraint sense of the mass balances alone, so I left it as is. Thanks for the other suggestions.

tpfau:
One argument against C and additional fields:
At the moment, you can't determine what a certain constraint is intended for, as it doesn't have a description (which is most often useful). `addRatioConstraint` gives the metabolite a name that gives this indication, and so could a general `addConstraint` function. 
In addition, these constraints can't properly be exported to other formats than a mat file or they end up there as additional metabolites and io becomes inconsistent. 
For me the thing is: We can currently handle these constraints in the format that is present and we can give them sensible descriptions/explanations (mets/metNames/metNotes). Furthermore, if metabolites are used, all current functions can properly handle them. 
If you introduce these new fields, you will have to update all functions that build their own LPproblem objects from a model structure or functions will be out of sync. 
Keeping to the "metabolite" concept (even if its a pseudo-metabolite), avoids all these issues.
If you want to be able to distinguish the type of 'metabolite' I could imagine a "metPseudoMetabolite" boolean matrix, that can be checked, if something like this should be ignored.

rmtfleming:
C*v {<=,=,>=} d 
contains the meaning inherently, without recourse to pseudometabolites.
All such inequality constraints, from whatever function, should converge on appending constraints in the form above so we can start homogenising the approach to such constraints. If people do not encode the building of these constraints then they will not be used, which is the default behaviour.

laurentheirendt:
Thanks everyone.

tpfau:
So, now we have a fluxvariability function that calculates the maximum objective using these constraints and ignores them when doing the variability -> Wrong results!
The same applies for all other functions that use both optimizeCbModel and solveCobraLP.

PR #552, opened 2017-06-15
Multiple small corrections

tpfau:
CreateMultiSpeciesModel now restricts the multispecies model to those fields present in all individual species models.
convertOldStyleModel now converts all empty numeric entries in cell arrays (default init vlue often present in old models) to ''
buildPairWiseModel now uses readCbModel instead of load (thereby automatically converting old style models)
Added a function to restrict a set of models to a given set of FieldNames (droppping all other fieldNames).


laurentheirendt:
Thanks @tpfau for fixing this! 🎉

PR #553, opened 2017-06-15
Added an option to specify which model to load from a mat file

tpfau:
Added an option to readCbModel allowing to specify a model name to load from a mat file. 


laurentheirendt:
That is a nice option @tpfau! Thanks for implementing this! 🌵

PR #565, opened 2017-06-17
Adding fast rxn/gene deletion based on Fast-SL

karthikraman:
I have integrated some hacks from our paper [Fast-SL](https://academic.oup.com/bioinformatics/article/31/20/3299/195638/Fast-SL-an-efficient-algorithm-to-identify) to perform single gene and single reaction deletions much faster. I have included a piece of test code in ``test/testFastDeletion/testFastDelCorrenctness.m`` to validate. The following is what shows up on my machine for testing:
````
load iAF1260.mat
load iLC915.mat
load iNJ661.mat

>> testFastDelCorrectness(iAF1260)
Gene deletion:
Fast single gene deletion analysis in progress ...
100%    [........................................]
Single gene deletion analysis in progress ...
100%    [........................................]
Difference in grRatios -- min: 3.0e-13, max: -3.8e-13; vec distance = 2.8e-12
# differences in hasEffect (~saving): 631
Fast deletion completed in 37.8% of time! (18.5 sec vs 48.8 sec)
Reaction deletion:
Fast single reaction deletion analysis in progress ...
100%    [........................................]
Single reaction deletion analysis in progress ...
100%    [........................................]
Difference in grRatios -- min: 3.0e-13, max: -5.1e-13; vec distance = NaN
# differences in hasEffect (~saving): 1951
Fast deletion completed in 17.9% of time! (19.4 sec vs 108.4 sec)

>> testFastDelCorrectness(iNJ661)
Gene deletion:
Fast single gene deletion analysis in progress ...
100%    [........................................]
Single gene deletion analysis in progress ...
100%    [........................................]
Difference in grRatios -- min: 4.1e-12, max: -1.1e-13; vec distance = 8.7e-12
# differences in hasEffect (~saving): 68
Fast deletion completed in 73.4% of time! (3.3 sec vs 4.4 sec)
Reaction deletion:
Fast single reaction deletion analysis in progress ...
100%    [........................................]
Single reaction deletion analysis in progress ...
100%    [........................................]
Difference in grRatios -- min: 8.6e-12, max: -4.5e-12; vec distance = 2.9e-11
# differences in hasEffect (~saving): 586
Fast deletion completed in 39.0% of time! (9.3 sec vs 24.0 sec)

>> testFastDelCorrectness(iLC915)
Gene deletion:
Fast single gene deletion analysis in progress ...
100%    [........................................]
Single gene deletion analysis in progress ...
100%    [........................................]
Difference in grRatios -- min: 1.8e-13, max: -8.1e-14; vec distance = 1.0e-12
# differences in hasEffect (~saving): 368
Fast deletion completed in 38.7% of time! (12.6 sec vs 32.6 sec)
Reaction deletion:
Fast single reaction deletion analysis in progress ...
100%    [........................................]
Single reaction deletion analysis in progress ...
100%    [........................................]
Difference in grRatios -- min: 2.1e-13, max: -1.5e-13; vec distance = 1.8e-12
# differences in hasEffect (~saving): 1064
Fast deletion completed in 25.7% of time! (20.3 sec vs 79.1 sec)
````

Since this is just a minor modification of existing `singleGeneDeletion` and `singleRxnDeletion`, I haven't performed other tests. This is my first contribution in a while --- please let me know if there are any issues, and if you require any clarifications.


laurentheirendt:
Great @karthikraman!

May you move `test/testFastDeletion/testFastDelCorrectness.m` to `test/verifiedTests`? Only files in that folder are currently being run by the continuous integration server. May you also follow the [test guidelines](https://github.com/opencobra/cobratoolbox/blob/master/docs/source/guides/testGuide.md) in order to fit the scheme required by the CI? Thanks!

I can see that there are several `.mat` files in your PR. We do not want to distribute `.mat` files within The COBRA Toolbox. Do you mind adding the model to the list in `retrieveModels` as explained [here](https://github.com/opencobra/cobratoolbox/blob/master/docs/source/guides/testGuide.md#4-load-a-model-andor-reference-data)?

Thanks again for your contribution, @karthikraman ! 🎉

karthikraman:
The models were for demonstration of the correctness. Are there any other tests that might be run? I would pitch for replacing `singleRxnDeletion.m` with `fastSingleRxnDeletion.m` and similarly for genes. I suspect it's a commonly used function in COBRA toolbox, and it would be good to have it run faster (with identical results output, as evidenced by the test file I uploaded). Could you or other COBRA community members review this?

laurentheirendt:
It would be preferable to download the models through `retrieveModels.m` instead of distributing them via the repository. For instance, the model `iAF1260` could be downloaded from the [BiGG database](http://bigg.ucsd.edu/models/iAF1260).

What do you think of combining `singleRxnDeletion.m` and `fastSingleRxnDeletion.m`, i.e. have an option in `singleRxnDeletion` to call your faster algorithm? Just a thought.

karthikraman:
Thanks @laurentheirendt, I think that's a possibility, but given that we are only discarding "useless" LPs, it should be OK to only do the fast deletion. I'll probably wait for @rmtfleming's review as well. Many thanks for looking into this!

laurentheirendt:
@karthikraman, sounds good. In the meantime, may you already add the models to the list in `retrieveModels` and remove the `.mat` files from this PR?

karthikraman:
Thanks, the `.mat` files are really not necessary. Have removed them in the latest commit. Also, the file `testFastDelCorrectness` works with any input model. However, some of the BiGG models don't seem to have the `grRules` field, required by 'singleGeneDeletion.m`.

tpfau:
I would suggest to just merge the fast Versions into the existing ones. 
As you said you are essentially just dropping unnecessary LPs, so this can just be merged into the existing code (and noted in the documentation, that this is using your idea), as otherwise the code is just inefficient.

rmtfleming:
new algorithms for the same functionality should be options to existing interfaces
singleRxnDeletion should be upgraded with the enhanced functionality envisaged in fastSingle...

karthikraman:
Thanks, will do so. What would be the default option, fast or regular?

tpfau:
fast, no need to make things slower than necessary

karthikraman:
As I look closer at the code, I would argue that the current `singleRxnDeletion` computes `hasEffect` incorrectly (as I understand: `hasEffect` is meant to flag if a reaction deletion affects _anything_).

If a reaction that carries **zero flux** in the WT solution is deleted, it wouldn't affect anything (essentially, this is the improvement proposed). Currently, `hasEffect` just returns a column vector of `true`, size of `rxnList`.

tpfau:
Actually: hasEffect in singleGeneDeletion is (according to the documentation), not an indicator of whether the gene has an effect on the objective. Instead it is essentially a question: "Does deleting the single gene lead to the deactivation of at least one reaction".
he hasEffect you are looking for would be: grRateKO<grRateWT
So you would at least have to look whether a gene can deactivate a reaction to get hasEffect (if the output is requested).

rmtfleming:
The documentation should follow the code, rather than the other way around. It would be wise to add a call to findFluxConsistentSubset into that function if the model does not contain the boolean structures
%    fluxConsistentMetBool:      `m` x 1 boolean vector indicating flux consistent `mets`
%    fluxConsistentRxnBool:      `n` x 1 boolean vector indicating flux consistent `rxns`
%    fluxInConsistentMetBool:    `m` x 1 boolean vector indicating flux inconsistent `mets`
%    fluxInConsistentRxnBool:    `n` x 1 boolean vector indicating flux inconsistent `rxns`

tpfau:
True wrt to the documentation, but the documentation for singleRxnDeletion would be "hasEffect: a vector of true the size of rxns", which kind of points to this being useless information (at the moment).

Could you expand, why you want to do an additional fluxConsistencyCheck here?

rmtfleming:
Any flux inconsistent reaction should get NaN in the hasEffect output.

tpfau:
hasEffect should be a boolean, and not NaN.
In addition, the first optimizeCbModel will already lead to an assignment of 0 to hasEffect for any flux inconsistent reaction (since it can't carry flux). So this would just be additional processing time particularily on larger models, as the initial LP has to be solved anyways.

rmtfleming:
Any flux inconsistent reaction should get NaN in the hasEffect output.

karthikraman:
I wouldn't say NaN, as the reaction is _not required to_ carry flux; not that it _can't_. For example, if there are two alternate reactions (e.g. a synthetic lethal pair), only one of them (arbitrarily) need carry flux in any solution. But both of them won't be lethal on deletion, and we need to check only the one that carried flux in the original solution for lethality while running `singleRxnDeletion`.

I would say we drop `hasEffect` altogether from `singleRxnDeletion`, and perhaps also `delRxn`. They both make sense only for gene deletions.

karthikraman:
Agreed. Done. Please review and merge. Thanks!

tpfau:
Looks good.
Just wondering:
Wouldn't the same approach for the fluxSolution in singleRxnDeletion (i.e. repmat of wildtype and only alter if there is an effect), be even more efficient?
I'm not entirely sure though...

rmtfleming:
Is everyone happy for us to merge?

karthikraman:
Is there any further action required from my side? Thanks!

rmtfleming:
@syarra

syarra:
I'll fix the conflicts and merge.

syarra:
Thanks @karthikraman, @laurentheirendt, @tpfau and @rmtfleming for this PR!

PR #570, opened 2017-06-18
Adding optKnock tutorial

snmendoz:
[*I added optKnock tutorial and modified related functions*]


laurentheirendt:
@snmendoz, may you give me more information about the file `tutorials/optKnock/optKnockTutorial.m`? I don't think that the naming and/or the location of this file is right. Is that simply the code of the tutorial? 

Did you generate this file from the `.mlx` file? The idea is to only have  4 files in the `tutorials/optKnock` folder, with the `.htm`, `.pdf`, and `.m` all being generated from the `.mlx` file (same content, including all comments, not only the code).

syarra:
@snmendoz, I ran your tutorial this morning and it gives me the following error:
optknock
Could you please have a look to fix this issue?
Thanks

snmendoz:
@syarra, sorry for that! 

I just fixed the error and checked that the tutorial is running properly

tpfau:
Hi,
please don't actively change the global variables (e.g. CBT_MILP_SOLVER) in a tutorial, you don't even have to mention it. Under normal circumstances a user should never have to manually modify (or even check) these, as it can otherwise lead to odd problems. (e.g. if a user doesn't have gurobi installed, setting the global to gurobi leads t failure which are hard to understand, as it fails in a place that it should never have reached if changeCobraSolver would have been used.)
Also: Please add some sections (ctrl+alt+Return) in the tutorial. E.g. after the initialisation block, after setting the constraints,  before I) SUCCINATE OVERPRODUCTION  etc.
That way a user can run the tutorial step by step. 

when loading the model:
If possible please use readCbModel (even if it is a .mat file). That way the model structure gets converted to the current format, and the tutorial doesn't get stuck if there are future changes. 

For your first TroubleShooting paragraph:
I would suggest something like:
"Optknock by default looks for up to 5 knockouts during the optimisation procedure. If the algorithm runs too long you can reduce the number of knockouts using the numDel option."...
EXAMPLE3: Is this example necessary? It is not introducing anything beyond the numDel parameter, so it essentially only runs longer and could (I think) be skipped.
The same applies to examples 2 and 3 for lactate.

snmendoz:
I will fix that about CBT_MILP_SOLVER. I think I put it that way because not always CBT_MILP_SOLVER was assigned when running changeCobraSolver and I just wanted to ensure that it was assigned. However, I can change it.

I will try to add some section breaks. 

I will add that troubleshooting paragraph and other like that. Do you prefer a single parragraph instead of separeting the paragraph in sections "problem", "possible reason" and "solution"? I followed that order according to the tutorial template

I removed example 3 from succinate production and example 2 from lactate production. I keep example 3 of lactate production because the type of solutions found between examples 1 and 3 are different. I will add some plots to show the difference.

tpfau:
wrt changeCobraSolver:
You are correct, 'all' doesn't necessarily change all solvers but only those which it is applicable to. And until yesterday, it didn't correctly state what it changed and what not. 
A single TROUBLESHOOTING paragraph is sufficient.
Its more to give some explanation as to why parameters are changed.

snmendoz:
@laurentheirendt I don't know why I didn't get a notification of your last comment. This is the second time I don't recieve an alert of your messages. Anyway, I answer here. Before I knew that tutorials must be created in .mlx files (like 9 monthts ago) I created the tutorials in .m. The file optKnockTutorial.m contains the code and some comments of the tutorial but it has not the exact content.

I didn't know that the idea is to have 4 files in the tutorial folder. How should I generate the .htm, .pdf and .m files? just copy and paste the content in the .mlx file and saving it on the corresponding file extension?

rmtfleming:
File>Export As... in matlab will be able to do it. Although in principal it is convenient to just do this once the tutorial is finalised, in practice it is useful to do it at intermediate steps so we can e.g. view the pdf of the tutorial and comment.

snmendoz:
@tpfau all the changes are ready.

Regards,

PR #571, opened 2017-06-18
adding a zero entry to model.C for all rxn indices

dayena:
@rmtfleming A zero entry has been added to model.C for all reaction indices. Previously, model.C was only filled for rxnInd of the coupled reaction indices (rxnInd). 


src/modelAnalysis/coupling/constrainRxnListAboveBound.m
```diff
@@ -11,7 +11,7 @@
 %    rxnList:             cell array of reaction names
 %
 % OPTIONAL INPUTS:
-%    C:                   `k x n` matrix in `C*v>=d`
+%    c:                   `k x n` matrix in `C*v>=d`
```
syarra: Is `C` a matrix or a vector?

rmtfleming:
Was originally a matrix, but now it is just a vector. The commenting needs to be updated.

PR #577, opened 2017-06-19
File io updates

tpfau:
This is a follow up of #462 
The SBML Binaries got updated and should no longer lead to the segfault experienced before.
Made a "default" minimum range of fields obtained from readCbModel. 


syarra:
Waoo! This is a long PR. I checked it quickly and  I didn't see anything odd. Thanks @tpfau 🎉 🏅

tpfau:
Please have a look at getDefinedFieldProperties and the COBRAModelFields.xlsx. 
You said, that the style I haad for the model field definitions is not readable, so its now in an easily readable excel sheet.

rmtfleming:
Hi Thomas,
please migrate COBRAModelFields.xlsx to a .m file with one line per field.
Regards,
Ronan

tpfau:
No.
Such a file would be unmaintainable, i.e. it can't be edited in a sensible way (since there is no good formatting, that allows an easy overview of it).
If you want the corresponding data in matlab format there is the function getModelFieldDefinitions(), which will, depending on some inputs, return you a cell array with the respective data.

PR #581, opened 2017-06-20
Setting of a value for epsil variable

arichelle:
the variable "epsil" was not defined in the script


laurentheirendt:
Thanks @arichelle! 🔝

PR #582, opened 2017-06-20
Add opt gene tutorial

snmendoz:
[*I added a tutorial for optGene and improved optgene related functions*]


syarra:
@snmendoz I ran your tutorial and I get the following errors:
optgene
Could you, please, have a look at them and fix them? If you need any help, please ping me.

snmendoz:
Hi @syarra. I got the same error when I was using glpk but I changed to gurobi and this issue was solved. What solver are you using? I think I'm going to use a simpler function to avoid this kind of problems.

snmendoz:
@syarra, please try again. I fixed the function analyzeOptKnock.

regards,

src/design/GetOptGeneSol.m
```diff
@@ -14,10 +14,14 @@
...
 %    isGeneList:      boolean
+%    saveFile:        boolean. Default = 0;
 %
 % OUTPUT:
 %    optGeneSol:      Solution in the desired format
 
+if nargin < 9 || isempty(saveFile)
+    saveFile = 0;s
+end
 writeDirect = 'C:\';
```
syarra: oh, this is dangerous ... Only compatible with Windows ....

snmendoz: I agree! I will change it

syarra: Thanks @snmendoz!

syarra:
I tried again your tutorial with your fixes  and 🎉 it is working now (with Gurobi)!
I still have no luck with GLPK, though it runs without error message:
screen shot 2017-06-21 at 17 44 25


I guess you should enforce people to use Gurobi at the beginning of the tutorial. What do you think?

Could you explain me the meaning of the timings reported in the `TIMING` section?
The runs are about 6 minutes... What about the `(2-3 minutes for run)` statement?

Thanks @snmendoz for your good work!

snmendoz:
Good idea. I will put a statement encouraging people to use gurobi. 

optGene takes ~2 minutes each time it is executed (2 minutes per run), so since optGene is executed three times, the total time is 6 minutes. I will change it to 6 minutes, so peple will know the total time.

snmendoz:
@syarra all the pending tasks were performed. 

Regards,

syarra:
Thanks a lot @snmendoz for your contributions! Good job!

PR #586, opened 2017-06-20
Model merge without genes

tpfau:
Reintroduced model merging without gene related fields.


rmtfleming:
This is urgent and it only affects one file, so why not merge it right away rather than adding the do not merge label?

syarra:
Waiting for the  CI to complete its job!

rmtfleming:
If Thomas has tested this function himself, there should be no issue.

tpfau:
I agree with Sylvain that the CI should always be finishing its job, as this is the only way we have to be sure, that there are no unintended side effects. If someone urgently wants the changes he can cherry pick them to his/her copy without any issue. Or (in this instance) just do the minimal modification themselves, and stash the changes before updating again.

PR #587, opened 2017-06-20
relaxFBA partial fix

rmtfleming:

### DIFF
--- a/src/base/solvers/configEnvVars.m
+++ b/src/base/solvers/configEnvVars.m
@@ -194,7 +194,8 @@ function subDir = generateSolverSubDirectory(solverName)
 
         % check for 32-bit
         if ~isempty(strfind(computer('arch'), '64'))
-            subDir = [filesep, osPath, filesep, 'matlab'];
+            subDir = [filesep, 'matlab'];
+            %subDir = [filesep, osPath, filesep, 'matlab'];
         end
     end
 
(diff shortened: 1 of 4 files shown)


rmtfleming:
This is one to cherry pick. It is out of date with the devel as it was an old branch from a week ago without card opt recon x.

tpfau:
What is this pr supposed to do? 
The change in the env-vars, will make this fail on all systems which use the default installation path, since it now looks for a non default location.
There is a style output change in printConstraints, which makes sense and we might actually think about in general changing %f and %e to %g, but there are no functional changes to relaxFBA (only the reordering of the docu, and a creation of an unused variable). 
Did you miss some commit here or accidentially created the wrong PR?

rmtfleming:
I restarted on this branch, realising that I needed to be on a new, upto date branch of develop. The env-vars is a fix of something I already fixed already, why it is not in the master by now, I dont know. I am looking at the relaxFBA at the moment.

tpfau:
The env-vars isn't in because this is something that is specific to your machine. On any other machine that used a default gurobi or other solver install, this will fail, so it was not merge with master, as it will fail for a majority of people.

> I restarted on this branch, realising that I needed to be on a new, upto date branch of develop. [...] I am looking at the relaxFBA at the moment.

But why did you put in a pr then, when you are still working on it?

rmtfleming:
@laurentheirendt

PR #592, opened 2017-06-20
installDevTools function

laurentheirendt:
- `installDevTools()` function to install the MATLAB.devTools from within MATLAB.
- updated documentation and `README` instructions

*Note:* Tested on `Windows`, `macOS` and `Linux`. CI test to be added at a later point.

### DIFF
--- a/src/base/installDevTools.m
+++ b/src/base/installDevTools.m
@@ -14,14 +14,15 @@ function installDevTools()
     currentDir = pwd;
 
     % change to the local directory
-    cd([CBTDIR filesep '..']);
+    localDir = [CBTDIR filesep '..'];
+    cd(localDir);
 
     installFlag = false;
-    localDir = '';
 
     if exist([CBTDIR filesep '..' filesep 'MATLAB.devTools'], 'dir') == 7
 
-        reply = input([' > There is already a copy of the MATLAB.devTools installe in the default location.\n   Do you want to enter another location to install the MATLAB.devTools? Y/N [Y]: '], 's');
+        reply = input([' > There is already a copy of the MATLAB.devTools installed in the default location.\n   Do you want to enter another location to install the MATLAB.devTools? Y/N [Y]: '], 's');
+        localDir = '';
 
         if isempty(reply) || strcmpi(reply, 'y') || strcmpi(reply, 'yes')
         % enter another location for the MATLAB.devTools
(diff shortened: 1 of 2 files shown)


src/base/installDevTools.m
```diff
@@ -0,0 +1,105 @@
...
+                indexDir = strfind(CBTDIR, filesep);
+                devToolsDir = [CBTDIR(1:indexDir(end)), 'MATLAB.devTools'];
+            else
+                devToolsDir = dirReply;
+            end
+
+            fprintf(['Done.\n > The MATLAB.devTools have been installed into: ', strrep(devToolsDir, '\', '\\'), '\n\n']);
+
+            % add the path to the devTools
+            addpath(genpath(devToolsDir));
```
tpfau: maybe it would be worth excluding the ".git" folder from the genPath command.
I had it a few times now, that a branch name shadowed the actual function when using ctrl-D

laurentheirendt: Good point. I will add that.

src/base/installDevTools.m
```diff
@@ -0,0 +1,105 @@
...
+
+        reply = input([' > There is already a copy of the MATLAB.devTools installed in the default location.\n   Do you want to enter another location to install the MATLAB.devTools? Y/N [Y]: '], 's');
+        localDir = '';
+
+        if isempty(reply) || strcmpi(reply, 'y') || strcmpi(reply, 'yes')
+        % enter another location for the MATLAB.devTools
+
+            dirReply = '';
+            while isempty(dirReply)
+                dirReply = input(['\n -> Please define the installation path of the MATLAB.devTools\n    Enter the path: '], 's');
```
laurentheirendt: I am not a big fan of that. Maybe later.

src/base/installDevTools.m
```diff
@@ -0,0 +1,105 @@
...
+    installFlag = false;
+
+    % check if the devTools already exist
+    if exist([CBTDIR filesep '..' filesep 'MATLAB.devTools'], 'dir') == 7
+
+        reply = input([' > There is already a copy of the MATLAB.devTools installed in the default location.\n   Do you want to enter another location to install the MATLAB.devTools? Y/N [Y]: '], 's');
+        localDir = '';
+
+        if isempty(reply) || strcmpi(reply, 'y') || strcmpi(reply, 'yes')
+        % enter another location for the MATLAB.devTools
```
laurentheirendt: Well, the default is `[Y]`, so by hitting `ENTER`, a user may proceed.

tpfau: true

src/base/installDevTools.m
tpfau: It's just a matter of preference :) 
As a programmer: I completely agree, from a users point of view typing in a directory is a pain. ^^

syarra: It is a good idea, but for later ;)

syarra:
Thanks @laurentheirendt! Let's spread the use of this new function throughout the community.

PR #593, opened 2017-06-21
Updated 'all' behaviour

tpfau:
At the moment, `changeCobraSolver('glpk','all')` will lead to statements indicating that all solvers have been set to GLPK, which is wrong, as only e.g. LP and MILP will have been set to GLPK. 
However, the user is not made aware of this, and also not of the reason why this is happening.

This pr introduces feedback during the `changeCobraSolver(solver,'all')` call in accordance with the actual changed settings, and return the appropriate error messages (as text).

### DIFF
--- a/src/base/solvers/changeCobraSolver.m
+++ b/src/base/solvers/changeCobraSolver.m
@@ -264,22 +264,15 @@ if ~any(strcmp(supportedSolversNames, solverName))
 end
 
 % Attempt to set the user provided solver for all optimization problem types
-if strcmpi(solverType, 'all')
-    for i = 1:length(OPT_PROB_TYPES)
-        try
-            solverOk = changeCobraSolver(solverName, OPT_PROB_TYPES{i});            
-            if printLevel > 0
-                if (solverOk)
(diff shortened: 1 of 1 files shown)


rmtfleming:
If the tests pass, please merge.

src/base/solvers/changeCobraSolver.m
```diff
@@ -266,8 +266,20 @@
 % Attempt to set the user provided solver for all optimization problem types
 if strcmpi(solverType, 'all')
     for i = 1:length(OPT_PROB_TYPES)
-        changeCobraSolver(solverName, OPT_PROB_TYPES{i}, 0);
-        fprintf([' > CBT_', OPT_PROB_TYPES{i}, '_SOLVER has been set to ', solverName, '.\n']);
+        try
```
syarra: There is actually a structure called `SOLVERS` that holds the type of problem a solver can solved.
To access it you can do:
```
global SOLVERS
SOLVERS.glpk.type
```
so the `for` loop could be replaced by: `for i = 1:length(SOLVERS.(solverName).type)` and the `try ...catch` is not needed anymore.

tpfau: Good point.
Actually this way, a no existing solver would throw an error, while an existing one can display what it was set for.
I updated the pr accordingly.

PR #594, opened 2017-06-21
Fixed old input parser

tpfau:
Small fix to readCbModel when using old style inputs.
The problem was that the concatenation of cell arrays led to compSymbols/compNames being aded as 2 instead of one additional cell. 


rmtfleming:
ok to merge.

PR #597, opened 2017-06-22
Adding flipDirectionality

tpfau:
Simple functionality to flip the directionality for reactions (including all bounds and objective coefficients). Could be useful if there are methods that only work in the positive direction on a reaction that is commonly working in reverse.


rmtfleming:
This is flipping the orientation of the reaction. The direction is still the same in reverse.

tpfau:
Renamed the function to reflect that only the orientation is changed.

PR #598, opened 2017-06-22
Examples for parsers

tpfau:
Adding in examples for methods with parsers for users unfamiliar with this MATLAB concept. 
This is a pure documentation update. 


rmtfleming:
@laurentheirendt  
sampleCbModel.m will need to be reverted to an optional parameter structure at some point.

PR #599, opened 2017-06-22
Script to run tutorials on the CI

syarra:


.ci/runTutorial.m
```diff
@@ -0,0 +1,24 @@
+function runTutorial(tutorialName)
+
```
laurentheirendt: May you add some documentation here?

.ci/runTutorial.m
```diff
@@ -0,0 +1,24 @@
...
+% include the root folder and all subfolders
+addpath(genpath(pwd))
```
laurentheirendt: Is it necessary to add the entire path? What about hidden folders? 

Unless I am mistaken, it should be enough to only add the root itself, i.e. `addpath(pwd)`. The remaining paths are added when running `initCobraToolbox`.

.ci/runTutorial.m
```diff
@@ -0,0 +1,24 @@
...
+
+% run the official initialisation script
+initCobraToolbox
+
+% Mute progress bars and initCobraToolbox
+global WAITBAR_TYPE;
```
laurentheirendt: May you declare the `global` variables further up?

.ci/launchTutorials.sh
```diff
@@ -0,0 +1,58 @@
...
+
+report="Tutorial report\n\n"
+report+="Name                               passed    failed    time(s)\n"
+report+="--------------------------------------------------------------\n"
+failure=0
+
+# Set time format to seconds
+TIMEFORMAT=%R
+
+for tutorial in "${tutorials[@]}"
```
laurentheirendt: 🎉

.ci/runTutorial.m
syarra: Done!

.ci/runTutorial.m
syarra: You're right. I changed it.

.ci/runTutorial.m
syarra: Moved after docstrings

laurentheirendt:
Thanks @syarra 🌵

PR #611, opened 2017-06-26
reset the global variables if ENVARS.STATUS is 0

syarra:


laurentheirendt:
Good catch @syarra 👍

PR #612, opened 2017-06-26
Tutorial path vectors  and minimal cut sets 

syarra:
Thanks Susan for this contribution.


rmtfleming:
@syarra model.rev is depreciated
Please add citation to both new function interfaces
Klamt S, Saez-Rodriguez J and Gilles ED (2007) Structural and functional analysis of cellular networks with CellNetAnalyzer. BMC Systems Biology 1: 2
Also add pointer to download of CNA
http://www2.mpi-magdeburg.mpg.de/projects/cna/download.html

src/modelAnalysis/topology/minimalCutSets.m
```diff
@@ -0,0 +1,93 @@
...
+%
+%    output = minimialCutSets(targets, varargin)
+%
+% INPUT:
+%    targets (mandatory): a binary matrix that row-wise contains the
+%                         target paths/cycles/elementary modes; a '1' in
+%                         the i-th row and j-th column in targets indicates
+%                         participation of element (reaction) j in mode i;
+%
+%    mcsmax:              maximal size of cutsets to be calculated; must
```
tpfau: This should be splitted into optional and required inputs.

src/modelAnalysis/topology/minimalCutSets.m
```diff
@@ -0,0 +1,93 @@
+function [output] = minimalCutSets(targets, varargin)
+% Computes minimal cut sets for paths/cycles/elementary modes with Berge algorithm
```
tpfau: The signature has to be changed to use optional arguments and not parameter/value pairs as required by @rmtfleming in #497

src/modelAnalysis/topology/pathVectors.m
```diff
@@ -0,0 +1,195 @@
+function [output] = pathVectors(model, directory, varargin)
+% Computes elementary mode and extreme pathway
```
tpfau: The signature has to be changed to use optional arguments and not parameter/value pairs as required by @rmtfleming in #497

src/modelAnalysis/topology/pathVectors.m
```diff
@@ -0,0 +1,195 @@
...
+    error('iso_flag should be either 0 or 1');
+end
+
+if positivity ~= 0 && positivity ~= 1
+    error('positivity should be either 0 or 1');
+end
+
+%% Convertining reversible reactions into irreversible reactions
+if positivity
+    ind = model.rev == 1;
```
tpfau: as mentioned b< @rmtfleming : should be model.lb < 0 and .rev should be ignored (its also not checked by CellNetAnalyser)

src/modelAnalysis/topology/pathVectors.m
```diff
@@ -0,0 +1,195 @@
...
+    rev = [model.rev; model.rev(ind)];
+    rxns = strcat(model.rxnNames(ind), '-rev');
+    c = [model.c; model.c(ind)];
```
tpfau: CellNetAnalyser assumes that the model objective in a COBRA model is always maximisation. 
We have to either coordinate with the CNA authors, that they consider the osense field, or we have to modify the csense according to the osense field.

src/modelAnalysis/topology/minimalCutSets.m
syarra: Done in 0f9ec34.

src/modelAnalysis/topology/pathVectors.m
syarra: Done in 0f9ec34.

PR #614, opened 2017-06-26
addition of a function to update model.metNames using an exisiting model

dayena:
This function allows us to update model.metNames using an existing model with corrected metNames


rmtfleming:
@tpfau please review this.

src/reconstruction/refinement/updateMetNames.m
```diff
@@ -0,0 +1,30 @@
...
+%    modelToUpdate:      COBRA model structure that needs to have its
+%                        model.metNames updated 
+%
+% OUTPUT:
+%    updatedModel:       COBRA model structure with corrected metNames
+%
+% .. Authors:
+%       - written by Diana El Assal 27/06/2017
+
+indMets = findMetIDs(originalModel, modelToUpdate.mets);
```
tpfau: This could (potentially) be done indenpendent of the compartment (i.e. creating a metID -> metName matching) which could cover a larger fraction if the new model has additional compartments.

src/reconstruction/refinement/updateMetNames.m
```diff
@@ -0,0 +1,30 @@
...
+updatedModel = modelToUpdate;
+updatedModel.metNames = [];
```
tpfau: Is it intentional, that every existing metName gets removed? 
It could be useful to e.g. only update those names, where the "reference" model has a value, and leave the others as they are, (and only create the metNames field if it does not exist).
Also: metNames should be a cell array {} not a double array [] and empty entries should be initialized with '' and not [].

src/reconstruction/refinement/updateMetNames.m
dayena: Thanks Tomas, I have now changed this function to update metNames, independently of compartment. Yes, it is intentional that every existing metName in the modelToUpdate is removed, because we have had cases when model.mets is not matching model.metNames

tpfau: You mean metNames is wrong? 
I would suggest an option to "wipeExisting", as I would assume that the "normal" situation is that someone wants to take information from another model without removing the current information if there is any.

src/reconstruction/refinement/updateMetNames.m
```diff
@@ -17,14 +17,22 @@
...
 %       - written by Diana El Assal 27/06/2017
 
-indMets = findMetIDs(originalModel, modelToUpdate.mets);
+metsReference = strtok(referenceModel.mets, '[');
+metsReference = metsReference(~cellfun('isempty',metsReference));
+metsReference = [metsReference, referenceModel.metNames];
+
+metsUpdate = strtok(modelToUpdate.mets, '[');
+metsUpdate = metsUpdate(~cellfun('isempty',metsUpdate));
+
```
tpfau: As a suggestion:
```
%remove the compartment info
metsReference = regexprep(referenceModel.mets,'\[[^]\]$','');
metsUpdate = regexprep(modelToUpdate.mets,'\[[^]\]$','');
updatedModel = modelToUpdate;
[metsToUpdate,positionsToUpdateFrom] = ismember(metsUpdate,metsReference);

%reset metNames field in modelToUpdate (if requested), or initialize if not existing
if wipeExisting || ~isfield(updatedModel,'metNames')
  updatedModel.metNames = cell(size(updatedModel.mets));
  updatedModel.metNames(~metsToUpdate) = {''}; % init metabolites not present in the second model.
end

updatedModel.metNames(metsToUpdate) = referenceModel.metNames(positionsToUpdateFrom(metsToUpdate));
```
In the current version, if you have any metabolite without a compartment id, the code will fail, as you remove them with the ~cellfun command. The regexp will just remove it.

dayena:
Hi @tpfau 

Thanks for your suggestion of keeping the metNames as an option (i.e. wipeExisting). I tested with and without this option. It works fine when you wipeExisting = 1, but when 0, model.mets and model.metNames become unmatched. Could you please look into this? 

Thanks, 

Diana

tpfau:
Could you give me an example?

tpfau:
Was your modelToUpdate correctly sized? i.e. was `all(size(modelToUpdate.mets) == size(modelToUpdate.metNames))` true?

dayena:
Hello Tomas,

My original intention for this function was exactly because of this type of situation when model.mets is not sized equally to model.metNames!

Best wishes,

Diana

tpfau:
I'm just wondering how that would happen except by manual manipulation of the fields (which is hihgly discouraged). 
However, if thats apparent, it would anyways be necessary to dismiss the whole existing data (i.e. the metNames field), as noone would know whether the missing entry originates from an insertion at the end or somewhere in the middle.
so in that instance, I would suggest the following:
```
if wipeExisting || ~isfield(updatedModel,'metNames') || ~(length(model.metNames) == length(model.mets))
  updatedModel.metNames = cell(size(updatedModel.mets));
  updatedModel.metNames(~metsToUpdate) = {''}; % init metabolites not present in the second model.
end
```
Also, add a optionality switch for wipeExisting at the beginning of the function:
```
if ~exist('wipeExistening','var')
  wipeExisting = false;
end
```
If the fields are inconsistent, I agree to just replace anything present, as its validity is unknown. But this should only happen, if the model was modified manually, i.e. not using functions from the Toolbox.
If you have a situation of where the latter happens, please let me know, because then we need to correct the offending function.

dayena:
Hi Tomas, 

I think the output of checkStoichiometricConsistency.m doesn't update the metNames. This code is however too intense for me! :) that's why perhaps it's better to just add this function of updatingMetNames.... i will adapt your suggestions now and test them! Thank you!

tpfau:
I'm not sure where checkStiochiometricConsistency would actually modify any basic fields. 
However, if you encounter any methods that do not correctly adapt existing fields, please raise an appropriate issue, as this needs to be fixed because we rely on methods not invalidating existing fields.

Issue #627, opened 2017-06-28
extractSubNetwork and extractCompModel do not reduce the model.genes and model.rxnGeneMat


Gpreciat:
Hi @tpfau 
When I use the functions extractSubNetwork and extractCompModel the fields genes and rxnGeneMat are not reduced. 
If some reactions are extracted, some columns (genes) of rxnGeneMat might have only zero entries which should be deleted same with the same index of the genes field.

```
load ecoli_core_model.mat
compModel = extractCompModel(model, 'c', true);
length(compModel.genes)
```
The length is 137 but it should be 88

**I hereby confirm that I have:**

- [x] Tried to solve the issue on my own
- [x] Retried to run my code with the latest version of The COBRA Toolbox
- [x] Checked that a similar issue has not already been opened

tpfau:
There are two sides to this. 
Its either leaving in the genes (which does not hurt at all in any further computation), or removing the genes and recalculating e.g. the rules field (on larger models this can take quite some time.
I can implement an option to remove the genes, but I would leave them in by default, as it makes the functions a lot faster.

tpfau:
Currently working on this and I got a question:
In case a gene is removed from the model:
What do you think, how should we update the rules field? 
e.g. if we have a rule:
Gene1 and Gene2 or Gene3 and we remove Gene2 from the model, how should the rule look?
While in general, I would claim "you shouldn't do this, ever", I think the most sensible way would be to look whether there is an and left or right of the removed gene and elimiante the gene + that and and only if there is no and remove the gene plus an or. (or in case its the only gene, just remove it).
i.e. the above would become:
Gene1 or Gene3
Whats your take @syarra @Gpreciat ?

Gpreciat:
Hi @tpfau , 

In my own opinion, I agree that the gene rules don't have to be modified, in your example "Gene1 AND Gene2 OR Gene3" if Gene 2 is not involved in any of the reactions (Gene2 column in rxnGeneMat has only zeros) Gene2 could be deleted from model.genes and model.rxnGeneMat. Therefore, if Genes 1 and 3 are active we will have: true AND false OR true 

I think it is only necessary to make a rule like _**if a gene is not present in the model is equal to false**_. However, at least with the models that I have been reducing, genes that are not present in reactions are not found in the rules.

tpfau:
As with so many things, its not so easy to restrict the question to this specific situation, but rather how in general to handle these types of situations.
[...]

PR #637, opened 2017-06-30
Sub net update

tpfau:
Addressing the request in #627 
The main issue is the question of how to modify GPRs if a Gene present in a GPR is removed from the reconstruction. The current approach is:
Eliminate a adjacent and clause >> eliminate an adjacent or clause >> eliminate the gene.
i.e. if there is an at least one adjacent "and" (e.g. G1 and G2 and G3 or G4, all genes except for gene 4 have an adjacent and) an adjacent "and" will be removed. Only if thats not possible, an adjacent "or" will be removed and if thats also not possible we can be sure that this is a solitary literal which can be entirely removed.


rmtfleming:
Too many unrelated changes to the issue at question. A separate branch please.

laurentheirendt:
Merging as discussed. Thanks @tpfau.

PR #638, opened 2017-06-30
Fixed gprFlag behaviour if no grRules field is present, using rules f…

tpfau:
Added a switch to allow printout of protein rules based on the rules field (if e.g. proteins are defined in an sbml import). Also fixed a bug when grRules was not present. Using rules now.


rmtfleming:
go ahead @syarra when you are ready

laurentheirendt:
Thanks @tpfau.

PR #656, opened 2017-07-04
fixing curl for AntCore model

snmendoz:
[*I modified the url for obtaining the AntCore model. Also, I added -L to curl command to avoid redirection when communicating with github*]

### DIFF
--- a/test/models/retrieveModels.m
+++ b/test/models/retrieveModels.m
@@ -25,7 +25,7 @@ end
 
 % define the array of models to be downloaded (name of the file and URL)
 modelArr = {
-    'AntCore.mat', 'https://cdn.rawgit.com/snmendoz/Models/master/AntCore.mat';
+    'AntCore.mat', 'https://raw.github.com/snmendoz/Models/master/AntCore.mat';
     'iIT341.xml', 'http://bigg.ucsd.edu/static/models/iIT341.xml';
     'Abiotrophia_defectiva_ATCC_49176.xml', 'https://webdav-r3lab.uni.lu/public/msp/AGORA/sbml/Abiotrophia_defectiva_ATCC_49176.xml';
     'Sc_iND750_flux1.xml', 'http://gcrg.ucsd.edu/sites/default/files/Attachments/Images/InSilicoOrganisms/yeast/Sc_iND750_flux1.xml';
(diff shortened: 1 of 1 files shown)


laurentheirendt:
@snmendoz, why exactly do you want to change the URL?

snmendoz:
@laurentheirendt The group of Costas D. Maranas noticed that the model AntCore was not been downloaded when retrieveModels was run. I checked that and I confirmed that the model was not been downloaded. So, I changed the url to download the model. I also added a -L because it wasn't working otherwise. I also checked that the 12 other models were corrected downloaded as well using -L.

laurentheirendt:
ok, but it shouldn't make a difference (unless rawgit.com wasn't online).

The advantage of the rawgit.com URL is that raw files are downloaded directly from GitHub with proper Content-Type headers.

snmendoz:
this is the value of the variable "result_curl" when running line 54 of retrieveModels
`
[status_curl, result_curl] = system(['curl --max-time 15 -s -k --head ', modelArr{i, 2}]);

HTTP/2 200 
date: Tue, 04 Jul 2017 22:26:56 GMT
content-type: application/octet-stream
x-content-type-options: nosniff
x-robots-tag: none
access-control-allow-origin: *
etag: "ea814c2158c158c68fd52228abd73b855497d7e7"
cache-control: max-age=86400
vary: Accept-Encoding
rawgit-cache-status: HIT
server: NetDNA-cache/2.2
x-cache: HIT
`

laurentheirendt:
Interesting. On my end:
```bash
$ curl --max-time 15 -s -k --head https://cdn.rawgit.com/snmendoz/Models/master/AntCore.mat
```
yields:
```
HTTP/1.1 200 OK
Date: Wed, 05 Jul 2017 02:13:01 GMT
Content-Type: application/octet-stream
Connection: keep-alive
X-Content-Type-Options: nosniff
X-Robots-Tag: none
Access-Control-Allow-Origin: *
ETag: "ea814c2158c158c68fd52228abd73b855497d7e7"
Cache-Control: max-age=86400
Vary: Accept-Encoding
RawGit-Cache-Status: HIT
Server: NetDNA-cache/2.2
X-Cache: MISS
```
The issue is with `OK` not being returned in the status. 

Anyway, I am fine with your proposed change. Thanks. 🎉

PR #665, opened 2017-07-06
Fixing optForce

snmendoz:
[*According to Costas D Maranas group comments, I changed the default value of printExcel to 0 when the operation system is different from windows; I highlighted in section "Equipment setup" that the latest version of GAMS should be installed and that issues could appear when using versions older than 24.7.1;  I removed the hard-coded lines;  I fixed the headers of the outputs files. Now, the correct headers are shown; Now, I show min and max fluxes for wild-type and mutant strains; I added an additional step to the tutorial; I  added a loop option to search from k= kMin to kMax; I added additional files (pdf and html)*]


laurentheirendt:
Thanks @snmendoz and @syarra! 🎉

PR #676, opened 2017-07-09
Speed up addExchangeRxn and modify createMultiSpeciesModel

shjchan:
When running `createMultiSpeciesModel` for large models, I found it very slow when it tries to add a lot of exchange reactions using `addExchangeRxn` because of the duplicate check in `addReaction` which calls `ismember` to compare sparse rows. Since they are only exchange reactions which can be easily checked for duplication, I suggest adding duplicate check in `addExchangeRxn` prior to calling `addReaction`.
I tested the original and modified `addExchangeRxn` with the following codes:

````
load('iAF1260.mat', 'model');
rxnEx = sum(model.S ~= 0, 1) == 1;
model2 = removeRxns(model, model.rxns(rxnEx), 'metFlag', false);
tic;
model3=addExchangeRxn(model2, model.mets(metEx));  % no duplicated reactions
toc;
model4 = removeRxns(model, model.rxns(find(rxnEx, sum(rxnEx) - 10)), 'metFlag', false);
tic;
model5=addExchangeRxn(model2, model.mets(metEx));  % 10 duplicated reactions
toc;
````

In the both cases (with/without duplicated reactions), **the original code took > 60 sec while the modified code took only around 5 sec**. The resultant models were checked to be identical using either codes. In the community model tested in `tutorial_SteadyCom` just submitted with another pull request, the original code took >20 min while the modified code took around 1 min.

A few lines are also added in `createMultiSpeciesModel` to add new metabolites with metNames and metFormulas prior to calling addReaction. This retains the information for metabolites and suppresses the warning as well (which I hoped when writing the tutorial for SteadyCom). This change apparently does not increase the computational time (addReaction will call addMetabolite anyway for new mets).

### DIFF
--- a/src/analysis/multiSpecies/createMultipleSpeciesModel.m
+++ b/src/analysis/multiSpecies/createMultipleSpeciesModel.m
@@ -200,6 +200,12 @@ for i = 1:length(modelNew.mets)
 
         % remove nameTag from metabolites
         MexG{cnt} = regexprep(MexG{cnt}, nameTag, '');
+        % add the metabolite in advance to avoid warning by addReaction and also give metNames and metFormulas to the metabolite to be added
+        if isfield(modelNew, 'metFormulas')
+            modelNew = addMetabolite(modelNew, MexG{cnt}, 'metName', modelNew.metNames{i}, 'metFormula', modelNew.metFormulas{i});
+        else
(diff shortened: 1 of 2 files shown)


rmtfleming:
As far as I see this is acceleration, rather than a change to existing behaviour, so once the test functions pass, it should be merged, assuming the tests are comprehensive enough.

laurentheirendt:
Thanks @shjchan. Nice PR. 🎉

PR #692, opened 2017-07-10
Fix addReaction with 'reactionName' as the 1st name-value argument

shjchan:
As I tried:
```
load('iAF1260.mat')
addReaction(iAF1260, 'test', 'reactionName', 'test', 'reactionFormula', 'glc__D_c ->');
```
It returns the error:
```
Error using addReaction (line 170)
The value of 'metaboliteList' is invalid. It must satisfy the function: iscell. 
```
It is because 'reactionName' was not in `optionalParameters` in line 66. So it treats it as the old style. Adding it back resolves the error. 
Where is the test for `addReaction` located? Does this need to be added into a test?

### DIFF
--- a/src/reconstruction/refinement/addReaction.m
+++ b/src/reconstruction/refinement/addReaction.m
@@ -25,7 +25,7 @@ function [model, rxnIDexists] = addReaction(model, rxnID, varargin)
 %                         * reversible - Reversibility flag (Default = true)
 %                         * lowerBound - Lower bound (Default = 0 or -vMax`)
 %                         * upperBound - Upper bound (Default = `vMax`)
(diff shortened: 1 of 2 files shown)


shjchan:
@laurentheirendt . Test added. And a few more changes for `addReaction`

laurentheirendt:
Thanks @shjchan

shjchan:
Hi @laurentheirendt , as suggested by @rmtfleming , I just added some test for backward compatibility and I found one more small change needed. Is it true that you can reopen this pull request and allow the additional commit? Or need a new pull request?

laurentheirendt:
Oh nice - tests are always welcome 👍 

I can't reopen an already merged PR. May you put in a new one? Thanks!

PR #702, opened 2017-07-11
tutorials from PR 664

laurentheirendt:
Cherry-pick of tutorials as per instructions in PR 664.

Tutorials:
- tutorial_relaxedFBA
- tutorial_constrainingModels
- tutorial_relaxedFBABounds

cc: @rmtfleming

PR #718, opened 2017-07-11
Tut4

rmtfleming:
This was a change from the past couple of days and I dont know if it is new, probably it is.


laurentheirendt:
I already cherry-picked `b3a28c9` in PR #702 (merged).

`f9b0cae` only adds the auto-generated `.m` file. I will regenerate these files later. Thanks.

PR #757, opened 2017-07-13
review tutorial constrainingModel

arichelle:
The tutorial have been completely restructured. Different points still need to be checked:
- the references need to be double checked as some of the initial references where not mentioned in the original text
- a reference for table 1 need to be added
- line 24, the reaction 'HMR_1735' is not present in recon2
- I suggest to change the anticipated results section as it far from straightforward to understand how the addition of constraints shrink the solution space using sparseFBA function even more as in this case the minimal number of reaction is bigger in the initial model...

Therefore an additional review of this tutorial is recommended


laurentheirendt:
cc: @dayena.

laurentheirendt:
Thanks @arichelle 👍

PR #758, opened 2017-07-13
Modifications of optionsMethods for extraction_transcriptomic tutorial

arichelle:
New set of optionMethod structures for running the extraction_transcriptomic tutorial.

Thanks @snmendoz for catching that !


laurentheirendt:
Thanks @arichelle 👍

PR #765, opened 2017-07-13
Rev tutorial atomic

fplanes:
We added a note to make sure Java and Chemaxon is in the PATH of the system. We used GLPK instead of Gurobi  and found the same results. We also corrected some typos.

- [X]  Tested my code on my own machine (Ubuntu 64-bits)


laurentheirendt:
Thanks @fplanes. May you add as well the `.m`, `.pdf`, and `.html` files? You may generate them directly from MATLAB by clicking on `save` in the `Live Editor`.

tutorials/atomicallyResolveReconstruction/tutorial_atomicallyResolveReconstruction.m
```diff
@@ -110,7 +119,7 @@
 % The variable |isChemaxonInstalled| contains a logical value defined by the user 
 % if the license is installed or not.
 
-isChemaxonInstalled = false; % Change varibale to "true" if ChemAxon is installed
+isChemaxonInstalled = true; % Change varibale to "true" if ChemAxon is installed
```
laurentheirendt: @fplanes, may you change this to `false`? Otherwise, the tutorial will fail on the CI server. Thanks 👍

syarra:
Thanks @fplanes!

PR #770, opened 2017-07-13
FVA tutorial review

arichelle:
Minor text modifications, removing of the recon3D example part to avoid confusion.

* the name of the authors of this tutorial are missing


laurentheirendt:
Thanks @arichelle

PR #776, opened 2017-07-14
Fixing bugs with solvers and arguments in functions.

fplanes:

I have deleted a call for Cplex objects before the function changeCobraSolver('ibm_cplex','LP'). This function also adds the paths to MATLAB of the solver APIs, so any call to Cplex objects should be performed after this line. I have tested the change using Linux and cplex solver.

In addition, there was a mistake in the input argument order for fastcore() function.  
I have changed from "A = fastcore(coreInd, model, epsilon, printLevel)" to A = fastcore(model, coreInd, epsilon, printLevel), as the fastcore function is defined as:
tissueModel = fastcore(model, core, epsilon, printlevel)


laurentheirendt:
Thanks @fplanes 🎉

PR #785, opened 2017-07-14
WIP: using the sign of the stoichiometry, rather than the stoichiometry itself

dayena:
Previously, the flux vector V was being multiplied by the stoichiometry. However, we only want to multiply V by the sign of the stoichiometry, not by the stoichiometry itself. 

### DIFF
--- a/src/dataIntegration/metabotools/computeFluxSplits.m
+++ b/src/dataIntegration/metabotools/computeFluxSplits.m
@@ -22,7 +22,7 @@ function [P,C,vP,vC] = computeFluxSplits(model,mets,V)
 %
 % .. Author: - Hulda S. Haraldsdottir, December 2, 2016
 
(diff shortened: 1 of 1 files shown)


laurentheirendt:
@dayena, thanks. May you elaborate more why this change is necessary? Do you have test code?

dayena:
Hi @laurentheirendt 
The stoichiometry can sometimes be -6 or -2 or -1 etc. This means that the stoichiometrically weighted flux would be multiplied by -6 or -2 or -1. @MasoudAhoo double checked with me that this change is required in this particular function. 
I do not have test code....
Best wishes, 

Diana

laurentheirendt:
Would it not be better to implement an additional option?

laurentheirendt:
Closing temporarily as discussed with @dayena.

PR #822, opened 2017-07-23
fixing bugs in updateGenes and extractModel

akaviaLab:


laurentheirendt:
@akaviaLab, thanks for this. May you have a look at why these tests are failing? One of the errors is:
```
Error occurred in testBuildPairwiseModels/testBuildPairwiseModels and it did not run to completion.

    ---------
    Error ID:
    ---------
    'MATLAB:badsubscript'

    --------------
    Error Details:
    --------------
    Index exceeds matrix dimensions.
    
    Error in mergeTwoModels (line 79)
                    model2.rules{i} = updateRule(model2.rules{i},oldindices2(model2.rxnGeneMat(i,:)),newindices2(model2.rxnGeneMat(i,:)));
    
    Error in createMultipleSpeciesModel (line 174)
                [modelJoint] = mergeTwoModels(modelJoint, model, 1);
    
    Error in buildPairwiseModels (line 53)
            [pairedModel] = createMultipleSpeciesModel(models, nameTagsModels);
    
    Error in testBuildPairwiseModels (line 19)
        buildPairwiseModels
```

akaviaLab:
Okay, this ended up being a bit long.
The problem starts because test/models/Acidaminococcus_fermentans_DSM_20731.mat has capital letter AND instead of and. That causes the new updateGenes() function to create a gene called AND, because lines 25-26 are caps specific.
In general, I don't think the model is supposed to have capital letters AND (or OR). So perhaps this specific model should be fixed.
It also has "gap_filled" as a gene, which I'm not sure about. It also seems to have two prefixes for the genes (591001.3.peg. and g.2555.peg.), which is kind of weird.

I added another commit that makes updateGenes case insensitive by repeating the replacing code twice, which seemed simpler than doing a case-insensitive regular expression. I also did the equivalent change in GPR2Genes.

At this point, all tests except testFASTCORE & testSolveCobraLP succeed. I'm using MATLABdevTools, but I think it will open a new PR.

akaviaLab:
Sorry. My mistake - I reopened this PR instead of submitting another one, since the tests are fine except for testFATCORE and testSolveCobraLP, and they don't work on the develop branch either.

PR #834, opened 2017-07-25
Fix minor bugs in updateGenes and extractModel

akaviaLab:
Made updateGenes work better, fix bug in extractMetModel, made updateGenes and GPR2genes case insensitive. This is an attempt to resubmit PR #822


laurentheirendt:
@akaviaLab, do you have any test for these functions? Or an example that shows why your changes are required would help. Thanks.

akaviaLab:
Tests - not right now.
I can describe when and how each function failed, so it needed to be modified. If you want, I can see about making them into tests.

GeneUpdates - commit 1012071
========================
This is an example of how it fails and could be a test for it
`
foo = readCbModel('/Users/uda2001/Documents/MATLAB/fork-cobratoolbox/test/models/Recon2.v04.mat');
foo2 = updateGenes(foo);`

After these lines, foo2 has 1140 genes, foo has 2140 genes.
After my commits, foo2 has 2140 genes.

Case-sensitive - 	commits 8bb56aa & eb535e4
====================================
test/models/Acidaminococcus_fermentans_DSM_20731.mat has capital letter AND instead of and. That causes the updateGenes() function (after commit 1012071) to create a gene called AND, because lines 25-26 are caps specific. See PR #822 for a longer comment about this

ExtractModel - commit 1012071
=======================
`drugModel = extractMetModel(foo, 'pppg9', 0, 1)`
drugModel has 3 rxns, 4 metabolites, but over 5000 metCharges (and other fields)
After my commit, the new model has only 4 metCharges

test/verifiedTests/reconstruction/testModelManipulation/testUpdateGenes.m
```diff
@@ -0,0 +1,41 @@
...
+
+% save the current path
+currentDir = pwd;
+
+fileDir = fileparts(which('testUpdateGenes'));
+cd(fileDir);
+
+modelDirectory = strcat('..', filesep, '..', filesep, '..', filesep, 'models', filesep);
+
+model = readCbModel(strcat(modelDirectory, 'Recon2.v04.mat'));
```
laurentheirendt: May you use here:
```
model = readCbModel([CBTDIR filesep 'test' filesep 'models' filesep 'Recon2.v04.mat']);
```
There is no need for a relative `modelDirectory`, which is not very robust.

test/verifiedTests/reconstruction/testModelManipulation/testUpdateGenes.m
```diff
@@ -0,0 +1,41 @@
+% The COBRAToolbox: testUpdateGenes.m
+%
+% Purpose:
+%     - testUpdateGenes tests updateGenes
+%
+% Authors:
+%     - Uri David Akavia August 2017
+
+% save the current path
```
laurentheirendt: (goes along with the comment below):
```
global CBTDIR
````

**Comment by laurentheirendt** [comment]
cc: @jacekwachowiak

**Comment by laurentheirendt** [comment]
@akaviaLab, may you have a look at why your test fails on R2014b? This is the error:
```

Error occurred in testExtractMetModel/testExtractMetModel and it did not run to completion.

    --------------
    Error Details:
    --------------
    Matrix dimensions must agree.
    
```
Thanks.

akaviaLab:
Okay. The problem seems to be something fundamental in how Matlab 2014b treats logical matrices of size 0. Matlab 2015b and 2016b are fine with those matrices, Matlab 2014b isn't (when trying to compare them to other matrices).

In order to avoid this error, I've added some code to extractMetModel that creates the right size matrix when no metabolites were found (either metabolite not in model given, or metabolite like ATP which is filtered out).

laurentheirendt:
Thanks @akaviaLab 🎉

PR #843, opened 2017-07-30
fixed FVA bug wtr orig model properties

marouenbg:
-FVA function modifies original model properties such as csense, osense
-added test to check the existence of fields in the original model


laurentheirendt:
Thanks @marouenbg 👍

PR #845, opened 2017-08-01
readSBML: extractGPRRule: Modify gene rules conversion to use descending order of genes

stefaniamagg:
Modified the extractGPRRule function within readSBML to use a descending order of genes when replacing genes by their position. Done because using ascending version resulted in partially replacing genes with high numbers with lower numbered genes:
```
>>fbc_grRules{1052}
ans =
(g459 and g460 and g461 and g4 and g4 and g4 and g462 and g463 and g464 and g4 and g465 and g466 and g467)
```
Using ascending order of genes:
```
>> model.rules{1052}
ans =
(x(4)59 & x(4)60 & x(4)61 & x(4) & x(4) & x(4) & x(4)62 & x(4)63 & x(4)64 & x(4) & x(4)65 & x(4)66 & x(4)67)
```
Using descending order of genes:
```
>> model.rules{1052}
ans =
(x(459) & x(460) & x(461) & x(4) & x(4) & x(4) & x(462) & x(463) & x(464) & x(4) & x(465) & x(466) & x(467))
```


laurentheirendt:
@stefaniamagg, may you add these great examples as a test? Thanks.

test/verifiedTests/base/testIO/testReadCbModel.m
```diff
@@ -0,0 +1,69 @@
...
+    'Abiotrophia_defectiva_ATCC_49176.xml']);
+
+% re-insert parentheses and square brackets to reactions
+modelXML.rxns = strrep(modelXML.rxns, '_LPAREN_', '(');
+modelXML.rxns = strrep(modelXML.rxns, '_RPAREN_', ')');
+modelXML.rxns = strrep(modelXML.rxns, '_LSQBKT_', '[');
+modelXML.rxns = strrep(modelXML.rxns, '_RSQBKT_', ']');
+
+% load mat model version
+load(['..' filesep '..' filesep '..' filesep ...
```
laurentheirendt: May you use here:
```
global CBTDIR
readCbModel([CBTDIR filesep 'models' filesep 'Abiotrophia_defectiva_ATCC_49176.mat']);
```

stefaniamagg: Sure, I actually thought I changed this line to use CBTDIR already.

stefaniamagg:
Yes, the two tests can for sure be combined, readCbModel will use readSBML anyway when reading in an xml file. It's probably a good idea to have the IO tests together in one folder, I didn't realize that there was already a test for readSBML since it's in a different folder.

laurentheirendt:
Yes, feel free to move the tests in `testSBML` to `testIO`.

laurentheirendt:
Closing and re-opening to relaunch CI.

stefaniamagg:
Am also working on combining the tests for readCbModel and writeCbModel for the different file types.

laurentheirendt:
@stefaniamagg, may you have a look at the error thrown or R2016b:
```
14:19:49 ================================================================================
14:19:49 Error occurred in testReadWriteCbModel/testReadWriteCbModel and it did not run to completion.
14:19:49 
14:19:49     ---------
14:19:49     Error ID:
14:19:49     ---------
14:19:49     'MATLAB:xlswrite:dlmwrite'
14:19:49 
14:19:49     --------------
14:19:49     Error Details:
14:19:49     --------------
14:19:49     Error using xlswrite (line 187)
14:19:49     An error occurred on data export in CSV format.
14:19:49     
14:19:49     Error in model2xls (line 111)
14:19:49     xlswrite(fileName,tmpData,'Reaction List');
14:19:49     
14:19:49     Error in writeCbModel (line 177)
14:19:49             model2xls(model, strcat(fileName, '.', format), input.compSymbols, input.compNames);
14:19:49     
14:19:49     Error in testReadWriteCbModel (line 36)
14:19:49             writeCbModel(model, writeTypes{i}, 'testData');
14:19:49     
14:19:49     Caused by:
14:19:49         Error using dlmwrite (line 112)
14:19:49         The input cell array cannot be converted to a matrix.
14:19:49 ================================================================================
```
That was working before merging the tests, no?

laurentheirendt:
Let me restart the ci ...

stefaniamagg:
The 2015b check claims to have been successful, but it wasn't according to the console text output:
```
Running testReadWriteCbModel
Document written
[Warning: Could not start Excel server for export.
XLSWRITE will attempt to write file in CSV format.] 

Error occurred in testReadWriteCbModel/testReadWriteCbModel and it did not run to completion.

    --------------
    Error Details:
    --------------
    An error occurred on data export in CSV format.
    
    Caused by:
        Error using dlmwrite (line 112)
        The input cell array cannot be converted to a matrix.
    
.
Done testReadWriteCbModel
```

stefaniamagg:
I am making these tests on a Windows 2016b. Is it possible that it's failing because of some Windows/Linux differences?

laurentheirendt:
Most likely, especially when dealing with Excel.

stefaniamagg:
Ok, let me change to Linux and try to find the issue..

laurentheirendt:
The test worked before on Linux as well, so I am not sure yet what is going on.

stefaniamagg:
I made some changes in the files that read xlsx and convert to cobra models because these did not work on my computer (matlab would run out of memory trying to convert the xlsx to cobra). I'll see if any of these changes pop up as errors when I run the test on linux.

laurentheirendt:
Maybe the out-of-memory has to do with reading the full file (see [this line](https://github.com/opencobra/cobratoolbox/blob/master/src/base/io/utilities/xls2model.m#L98)).

stefaniamagg:
It did, I removed this part from the function and then it worked fine. Is this line important for importing xls in linux?

laurentheirendt:
Yes, that line is important on Linux, as Linux doesn't read the maximum number of lines. We can maybe reduce that number.

stefaniamagg:
Ah! We've found our suspect then. Can I replace this with:
```
if isunix
[~, Strings, rxnInfo] = xlsread(fileName,'Reaction List', '1:65536');
[~, MetStrings, metInfo] = xlsread(fileName,'Metabolite List', '1:65536');
else
[~, Strings, rxnInfo] = xlsread(fileName,'Reaction List');
[~, MetStrings, metInfo] = xlsread(fileName,'Metabolite List');
end
```
?

laurentheirendt:
Yes, of course you can replace it. Although that would mean that eventually on Linux, an out-of-memory maybe occur. What if we reduce the limit to `20000`? And print a warning? I doubt `xls` is the right format for models with more than 20000 reactions.

stefaniamagg:
Ok, I'll replace it with 20000 and add a warning.

laurentheirendt:
Maybe the warning should only be printed if the number of reactions read is close to `20000`(i.e. when the limit is actually reached. We could also add a line to the doc about that.

stefaniamagg:
Actually, we may have to go even lower than 20,000. I'm out of the my 32GB memory again trying to read in the xlsx model of E. coli core.

laurentheirendt:
Wow. Maybe then split however the OS as you suggested, as the functions that read in `xls` are not the same on `pc` and `unix`. On `unix`, set the number of reactions to `20000` then with a warning when hitting the limit. Do you agree?

stefaniamagg:
Sounds good, I'll do that!

stefaniamagg:
I also added a warning when reaching the max. xlsread import limit for windows (65536).

stefaniamagg:
I was going to have warnings for the windows limit but I'm getting conflicting information for what that number is so I removed the warnings. The linux is still truncated at 20,000 and produces warnings if the imported arrays have 20,000 rows.

stefaniamagg:
According to the 2015b check output it works now, although it seems to give out the warning for max. rxns/mets reached even though the actual list is nowhere near that size (95 rxns, 72 mets). The console output print out of the rxns/mets prints out the correct number of rxns/mets, so I'm not sure how to fix this.
```
Running testReadWriteCbModel
Document written
[Warning: Added specified worksheet.] 
[Warning: Added specified worksheet.] 
[Warning: XLS is not recommended for large models.] 
[Warning: Maximum number of reactions reached. Model reaction list truncated at
19,999 reactions.] 
[Warning: XLS is not recommended for large models.] 
[Warning: Maximum number of metabolites reached. Model metabolite list truncated
at 19,999 metabolites.] 
```

laurentheirendt:
@stefaniamagg, may you solve this conflict? Let me know if you need my help. Thanks.

syarra:
@stefaniamagg, great job! 🎉

PR #847, opened 2017-08-01
Fix findRxnsFromGenes to work with new style models without rxnGeneMat

stefaniamagg:
findRxnsFromGenes now works with old and new COBRA style models; new style models do not have the field rxnGeneMat. Prefix "gene_" now added to all genes regardless of whether they start with a string or a number. Shortened code. 


laurentheirendt:
@stefaniamagg, thanks for this. Would you mind adding a brief test for this function?

test/verifiedTests/analysis/testExploration/testFindRxnsFromGenes.m
```diff
@@ -0,0 +1,53 @@
...
+
+% save the current path
+currentDir = pwd;
+
+% initialize the test
+fileDir = fileparts(which('testFindRxnsFromGenes'));
+cd(fileDir);
+
+% load model
+load('Test_EcoliCoreModel.mat')
```
laurentheirendt: What has changed between `Test_EcoliCoreModel.mat`and `ecoli_core_model`. If we keep this model, may we name it `testData_ecoli_core_model.mat`?

stefaniamagg: Ah, this is supposed to use the ecoli_core_model.mat. I had to save the model locally for the test since my Matlab kept crashing when I used initCobraToolbox and using CBTDIR without first running initCobraToolbox returned CBTDIR = []. I've changed the test to use CBTDIR and get the E. coli core model from the test/models folder.

laurentheirendt: I see. Thanks for changing that. 

May you use `readCbModel` instead of `load` and then `convertOldStyleModel(model);`? Just saw that now.

PR #898, opened 2017-09-04
Added the definition of Basic FBA fields 

tpfau:
Added the definitions of basic FBA fields to the Field Definition file and updated convertCobraModel to add those fields to models that lack them. This allows reading most basic models from mat files using readCbModel.


syarra:
Thanks @tpfau!

PR #899, opened 2017-09-04
Adding modelID to models loaded from a .mat file 

tpfau:
Similar to loadIdentifiedModel, readCbModel now adds the struct name from the loaded mat file as ID of the model if no ID is present.


laurentheirendt:
@tpfau, may you check why the tests fail? Thanks.

tpfau:
Because adding that ID made the loaded model not the same as the models saved for the test :)

PR #900, opened 2017-09-05
Fix for FVA when csense is longer than nMets

jennmodamio:


tpfau:
While this allows the code to continue csense being longer than mets would indicate a problem in the model according to the [definitions](https://github.com/opencobra/cobratoolbox/blob/master/docs/source/notes/COBRAModelFields.md) (i.e. this should not happen). IF it does, I would rather have the system throw an error, since its not clear, which value from csense belongs to which entry in mets.

laurentheirendt:
Good point. I will add a warning pointing to these definitions for now.

syarra:
Merging this PR knowing that we will have to carefully think about this issue in the following weeks.

PR #910, opened 2017-09-12
added support for tomlab solver

marouenbg:

### DIFF
--- a/src/base/solvers/solveCobraLP.m
+++ b/src/base/solvers/solveCobraLP.m
@@ -43,8 +43,6 @@ function solution = solveCobraLP(LPproblem, varargin)
 %                   are ILOG cplex and Tomlab parameter syntax. see example
 %                   for details.
 %
-%    parameters:    solver-specific parameters structure
-%
 % Optional parameters can also be set through the
 % solver can be set through `changeCobraSolver('LP', value)`;
 % `changeCobraSolverParams('LP', 'parameter', value)` function. This
--- a/src/base/solvers/solveCobraLP.m
+++ b/src/base/solvers/solveCobraLP.m
@@ -1105,6 +1105,9 @@ switch solver
         % Result = tomRun('cplex', tomlabProblem, 0);
         % This is faster than using tomRun
 
+        % set parameters (user parameters override defaults)
+        tomlabProblem.MIP.cpxControl = solverParams;
+        
         % set parameters
         tomlabProblem.optParam = optParamDef('cplex',tomlabProblem.probType);
         tomlabProblem.QP.F = [];
@@ -1122,9 +1125,6 @@ switch solver
         if exist('optTol','var')
             tomlabProblem.MIP.cpxControl.EPOPT = optTol;
         end
-
-        % set parameters
-        tomlabProblem.MIP.cpxControl = solverParams;
         
         % solve
         Result = cplexTL(tomlabProblem);
--- a/src/base/solvers/solveCobraLP.m
+++ b/src/base/solvers/solveCobraLP.m
@@ -38,6 +38,10 @@ function solution = solveCobraLP(LPproblem, varargin)
 %                   without regularisation.
 %
 %    primalOnly:    {(0), 1}; 1 = only return the primal vector (lindo solvers)
+%   
+%    solverParams:  solver-specific parameter structure. Formats supported
+%                   are ILOG cplex and Tomlab parameter syntax. see example
+%                   for details.
 %
 %    parameters:    solver-specific parameters structure
 %
@@ -1119,6 +1123,9 @@ switch solver
             tomlabProblem.MIP.cpxControl.EPOPT = optTol;
         end
 
+        % set parameters
+        tomlabProblem.MIP.cpxControl = solverParams;
+        
         % solve
         Result = cplexTL(tomlabProblem);
 
--- a/src/base/solvers/solveCobraLP.m
+++ b/src/base/solvers/solveCobraLP.m
@@ -1111,16 +1111,16 @@ switch solver
         tomlabProblem.QP.F = [];
         tomlabProblem.PriLevOpt = printLevel;
 
-        % if basis is availible use it
-        if isfield(LPproblem,'basis') && ~isempty(LPproblem.basis)
+        if isfield(LPproblem,'basis') && ~isempty(LPproblem.basis) && ...
+                ~contains('basis',fieldnames(solverParams))
             tomlabProblem.MIP.basis = LPproblem.basis;
         end
 
         % set tolerance
-        if exist('feasTol','var')
+        if exist('feasTol','var') && ~contains('EPRHS',fieldnames(solverParams))
             tomlabProblem.MIP.cpxControl.EPRHS = feasTol;
         end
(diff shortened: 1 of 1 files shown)


src/base/solvers/solveCobraLP.m
```diff
@@ -1119,6 +1123,9 @@
             tomlabProblem.MIP.cpxControl.EPOPT = optTol;
         end
 
+        % set parameters
+        tomlabProblem.MIP.cpxControl = solverParams;
+
```
tpfau: This overwrites any existing settings set above (and it always does so independent on whether the struct was provided or not). 
You would need a test to see if the struct was provided (which is probably a pain due to the messy input structure), or this will make other option settings impossible and have tomlab always run with default options.

src/base/solvers/solveCobraLP.m
```diff
@@ -38,6 +38,10 @@
 %                   without regularisation.
 %
 %    primalOnly:    {(0), 1}; 1 = only return the primal vector (lindo solvers)
+%   
+%    solverParams:  solver-specific parameter structure. Formats supported
+%                   are ILOG cplex and Tomlab parameter syntax. see example
+%                   for details.
```
tpfau: This comment says it works for ILOG and Tomlab (shouldn't this be restricted to tomlab, as it is only used there?
it's also already mentioned as "parameters" below (which should serve the same purpose)

src/base/solvers/solveCobraLP.m
marouenbg: :+1:

tpfau:
Having had a look at this again:
My personal opinion is that the input to this function is a complete mess, and easily leads to parameters being set two times, and, depending on the solver the one, or the other is actually choosen.
E.g. I can provide tolerances to solvers either as a parameter value pair, or in the solverParams struct.
solverParams is also not exactly linked to solver parameters but is also used for COBRA specific commands (e.g. line 231, where lifting is performed based on solverParams). 
Further, they are sometimes overwritten (gurobi_mex), partially ignored (gurobi, where e.g. a possible feasibility tolerance/optTol is always overwritten, Matlab, were its not applicable. Overwriting other parameters (e.g. now with tomlab) or any other combination...

Personally, I would suggest the following:
Parameters that CAN (and are) set by get/setCobraSolveParams should not be set by a "solver Specific Structure", as they are (obviously) not solver specific. 
They should either be set by get/setCobraSolverParameters, or as parameter/value pairs when calling `solveCobraLP()`. Thus, I would ignore any those parameters given in the solver specific struct.
Unfortunately it seems as if the "parameter" structure is a double purpose structure, which is really bad, and I think we should clearly state (as this PR suggests) that the additional structure is ONLY for solver specific parameters.
For backward compatability, we can keep the lifting exception in (with a warning), but I really think this should be streamlined.

rmtfleming:
The parameters structure, including the parts specific and non-specific to a solver, should override everything else. That was intentional when the option to add solver specific and nonspecific parameters was added to other structures. The reason is that when an expert knows they are overriding the default approach, then they should know what they are doing, otherwise they should use the default way of changing solver parameters for all parameters common to all solvers of a set class. An expert should not have to change one set of parameters one way and another another way.

marouenbg:
The parameter structure in this PR overrides everything else.

tpfau:
Could you make sure that this is also true for the other solvers?

tpfau:
My main issue with this struct is that it is great for setting solver specific parameters that cannot be set otherwise, but it can easily lead to misconceptions as to what parameters are actually used. 
e.g. if I create an 
`options = cplexoptimset`
, define some parameters in that set,
```
options.barrier.algorithm.Cur = 2;
options.mip.pool.intensity.Cur = 3;
```
 and then call:
`solveCobraLP(LPproblem,'feasTol',1e-7,options)`
When later reading that code, I would expect, that the feasibility Tolerance used was 1e-7. 
What would be your expectation? 
The same applies to:
```
changeCobraSolverParams('LP','feasTol',1e-7);
solveCobraLP(LPproblem,options)
```
I understand the argument that an expert should be able to use specific settings, I just think, that its easier for an expert, to adapt, than for a non expert, who can end up with code that looks like its doing the one thing but does something else.
The problem is essentially, that we have multiple ways to set the same values, which can easily lead to confusion.

marouenbg:
Thomas, 
I don't get your point. 
Is it wtr to other solvers or about precedence of parameters?

If it is about other solvers, I will open a new PR after this one as the developers require one feature per PR.

In the case you presented, feasOpt will be set to 1e-7, there is no ambiguity. Would you expect another value?
https://github.com/marouenbg/cobratoolbox/blob/efd0e5c889005c81d15002d41371442834ba41bb/src/base/solvers/solveCobraLP.m#1120

In the second case, the global parameter from the CBT is provided thus feasopt= 1e-7.
https://github.com/marouenbg/cobratoolbox/blob/efd0e5c889005c81d15002d41371442834ba41bb/src/base/solvers/solveCobraLP.m#L121

Generally, if the same parameter is provided in both the structure and through the name/key value, the parameter structure overrides anything else.

I will rebase this PR, then feel free to contribute to it because I didn't quite get the issue here.

tpfau:
@marouenbg You just showed the problem I mentioned:
The actual used feasibility if the paramegter struct overrides any other supplied parameter would be 1e-6. Because cplexoptimset creates a structure that contains the feasibility Tolerance on its cplex default of 1e-6, but you would not notice and thus assume it is 1e-7 as that is explicitly mentioned in the code. (And I'm not necessarily speaking about what the code will do at the moment, but what would happen if the struct is taking priority).

Otherwise: Yes, I was speaking about the other solvers and setting it thus that the parameters from the parameter struct always override any other settings (even though as mentioned, I'm, not sure, whether thats a good idea).

marouenbg:
would

```
options.barrier.algorithm.Cur = 2;
options.mip.pool.intensity.Cur = 3;
```
set your feasOpt to 1e-6?

tpfau:
No.
But `options = cplexoptimset('cplex')`  does create a large set of options. And the feasibility Tolerance is set in there. And thats what cplex suggests to use when setting up options.

marouenbg:
ok I understood the feasOpt 1e-6 would comes from
`options = cplexoptimset`

Then again I echo the recommendations of Ronan to have the parameter structure overriding everything else. 
The fact that you can change parameters in many different ways imposes naturally a hierarchy.

tpfau:
Yes, but as you noticed yourself, it is far from obvious which parameter is used. What could be an idea, is to throw a warning if the parameter structure provides values for parameters that can be set from a different source. 
Even though people will commonly not read it, it might help when problems arise.

marouenbg:
ok, deal. I will add that.

laurentheirendt:
@tpfau, as discussed, may you add a warning for ambiguous parameter settings in a separate PR so that we ensure that a solver specific parameter structure overrides any other set of parameters?

@rmtfleming, this PR can be merged once rebased, passing the CI.

cc: @syarra

rmtfleming:
No need for a warning. Rather add a note adjacent to the parameters annotation to say it overrides everything else. Please can people not rename variables.

PR #918, opened 2017-09-13
Find exc reaction update

tpfau:
Altered findExcReaction to also find Exchangers with non 1/-1 values and return all exchange reactions, that can transport compounds into the system as uptakes.

### DIFF
--- a/src/analysis/exploration/findExcRxns.m
+++ b/src/analysis/exploration/findExcRxns.m
@@ -11,8 +11,7 @@ function [selExc, selUpt] = findExcRxns(model, inclObjFlag, irrevFlag)
 % OPTIONAL INPUTS:
 %    inclObjFlag:    Include objective `rxns` in the exchange rxn set (1) or not (0)
 %                    (Default = false)
-%    irrevFlag:      Model is in irreversible format (1) or not
-%                    (Default = false)
+%    irrevFlag:      Deprecated.
 %
 % OUTPUTS:
 %    selExc:         Boolean vector indicating whether each reaction in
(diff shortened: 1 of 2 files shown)


rmtfleming:
findExcReaction is a duplicate of the function model = findSExRxnInd(model, nRealMet, printLevel)
 and the former should be deleted, or all code deleted and a wrapper maintained.

tpfau:
I would contradict that.
For one, findSExcRxnInd always includes the objective reaction in the set of Exchange Reactions indicated (while it's not necessarily an exchanger).
Also, findSExRxns uses a lot of heuristics that might (or might not) be correct for identification of Exchange reactions (sink_, DM_, EX_), while findExcReaction relies purely on the definition of an Exchanger as a single non zero entry in the S matrix, 
These are options findExRxnsInd does not offer, so it is not doing the same.

laurentheirendt:
@tpfau, may you rebase your branch? Some tests are failing, but these issues have been fixed in the meantime. Thanks 👍

laurentheirendt:
@tpfau, may you make the changes here as discussed? Thanks.

rmtfleming:
Not strictly a duplicate, but my point is the same, they both approach pretty much the same task. findExcRxns.m should be kept as a shell wrapper calling findSExRxnInd.m which should be extended with an option to elicit whatever aditional functionality findExcRxns.m.

tpfau:
This was done in 1d67ab6 and 6c047b5
The only thing that I did not move to `findSExRxnInd` is the check for the objective function, which (imo) doesn't make any sense in `findSExRxnInd` as it would introduce functionality, that has nothing to do with finding exchange reactions (and which I would only keep in the function for backward copatability)

PR #921, opened 2017-09-13
Adding Direct IO from the BiGG Database

tpfau:
Fixing a bug in readCbModel (not sure how this ever worked in so many tests 'sigh') and adding IO to read/write directly from the BiGG database. 


laurentheirendt:
@tpfau, do you have any test for this?

tpfau:
I can write one.

laurentheirendt:
@tpfau, do you mind looking at why the CI is failing here? Thanks!

tpfau:
Because the fix to the rules field generation from generateRules from #920 is not in. 
Thus .mat files are not correctly read by readCbModel if they do not contain a rules field, as it is generated with wrong dimensions.

tpfau:
Closing till #920 is in.

PR #922, opened 2017-09-14
Test for fastLeakTest

tpfau:
Adding Test for fastLeakTest and cleaning up the function. 

### DIFF
--- a/src/reconstruction/modelGeneration/fastLeakTest.m
+++ b/src/reconstruction/modelGeneration/fastLeakTest.m
@@ -27,18 +27,15 @@ end
 tol = 1e-06;
 modelClosed = model;
 % find all reactions that have only one entry in S
-clear count ExR
-for i = 1 :size(modelClosed.S,2)
-    if length(find(modelClosed.S(:,i))<0)==1
-        count(i,1)=1;
-        modelClosed.lb(i)=0;
-               % modelClosed.ub(i)=0;
-    elseif length(find(modelClosed.S(:,i))>0)==1
-        count(i,1)=1;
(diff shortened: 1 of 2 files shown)


src/reconstruction/modelGeneration/fastLeakTest.m
```diff
@@ -53,18 +50,24 @@
...
 end
 modelexchangesAbbr = unique([modelexchangesAbbr;rxnNames']);
-TestRxnNum = length(modelexchangesAbbr)
+TestRxnNum = length(modelexchangesAbbr);
 FluxExV =[];
 while cnt == 1
     modelClosed = changeObjective(modelClosed,modelexchangesAbbr);
     FF2=optimizeCbModel(modelClosed,'max');
-   ObjValue = FF2.f
+    if FF2.stat == 0
```
laurentheirendt: Shouldn't this read: `if FF2.stat ~= 1` ?

tpfau: No. The Error thrown here is only for an infeasible model. 
Admittedly, for other non optimal solutions there should probably also be an error, but they would mean very different things.

laurentheirendt: The value of `FF2.f` when `FF2.stat == -1` or `FF2.stat == 2`  is `0`. Is that intended behaviour?

tpfau: I'd say no. I added a error message for those instances.

laurentheirendt: Thanks!

Issue #923, opened 2017-09-14
Incorrect GPR parsing


pauljensen:
GPR parsing for Cobra models can be incorrect when a multi-gene GPR with _k_ genes references the first _k_ genes in the model.

#### Step 1: parseBoolean.m
`parseBoolean` converts a human-readable GPR string into a Matlab expression. For example `geneA or geneB or geneC` becomes `x(1) | x(2) | x(3)`. The indices on `x` correspond to the position in the individual rule. `parseBoolean` also returns a list of gene names: `{'geneA', 'geneB', 'geneC'}`.

#### Step 2: changeGeneAssociation.m
After calling `parseBoolean`, `changeGeneAssociation` re-indexes the rule to match the gene's position in the model's gene list. For example, it adds `geneA` to the gene list (if it isn't already in the list), and then replaces `x(1)` with the position in the gene list:

```matlab
model.genes = [model.genes; genes(i)];
nGenes = length(model.genes);
rule = strrep(rule,['x(' num2str(i) ')'],['x(' num2str(nGenes) ')']);
```

#### The Issue
Let's say we add the GPR 'geneA'. Then `genes = {'geneA'}` and `grRules{1} = 'x(1)'`, which is fine. Now we add a second GPR `'geneB and geneC'`. This parses to `x(1) | x(2)` via `parseBoolean`, and now we update the rule. `geneB` is added as the second gene, so `changeGeneAssociation` replaces `x(1)` with `x(2)`, giving the rule `x(2) | x(2)`. Now we add `geneC` as the third gene in the model, and change `x(2)` to `x(3)`. But, this also changes the `x(2)` corresponding to `geneA`, giving a final rule of `x(3) | x(3)`, which is incorrect. As stated above, this problem appears whenever there are multi-gene GPRs near the top of the model.

#### The Fix

We need to distinguish between indices relative to the rule (`parseBoolean`) and indices relative to the model's full list of genes (`changeGeneAssociation`). The simplest way is to insert a dummy character in the `parseBoolean` output, e.g. `x(#1)` to avoid overwriting previous find/replaces in `changeGeneAssociation`.

pauljensen:
PR #924 submitted.

laurentheirendt:
Thanks @pauljensen and @tpfau 👍

PR #924, opened 2017-09-15
Fix for issue #923

pauljensen:


tpfau:
I don't think the fix is that easy.
For one: parseBoolean should also work in other situations (which your current pr breaks). 
So, instead of changing parseBoolean to a different output, I would add 
`rule = regexprep(rule,'x\(([0-9]+)\','x(#$1)')` in changeGeneAssociation, which has the same effect and does not alter parseBooleans output.

laurentheirendt:
Closing in favor of PR #926.

Thanks @pauljensen 👍

PR #926, opened 2017-09-15
Fix for issue #923

laurentheirendt:
Continuation of PR #924.

### DIFF
--- a/src/base/parseBoolean.m
+++ b/src/base/parseBoolean.m
@@ -64,7 +64,7 @@ if ischar(str)  % if it's a string, use MH's code
                     elements{end + 1} = tok;
                 end
                 % The replacement token
-                newTok = ['x(' num2str(cnt) ')'];
+                newTok = ['x(#' num2str(cnt) ')'];
                 % Remove troublesome characters ([])
                 tok = regexprep(tok, '\[', '');
                 tok = regexprep(tok, '\]', '');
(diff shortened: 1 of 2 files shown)


tpfau:
Just as a reminder from the original PR: 
The fix needs to be restricted to the changeGeneAssociation function, as other functions depend on parseBoolean and altered output would make them fail.
As mentioned, the easiest is probably to run:
`rule = regexprep(rule,'x\(([0-9]+)\)','x(#$1)')` in line 66 of changeGeneAssociation (i.e. directly after parseBoolean was called).

laurentheirendt:
@tpfau, may you work with @pauljensen to implement your suggested fix? Thanks 👍

tpfau:
I would be happy to. but essentially its a very small change (reverting the changes to parseBoolean and adding the line I mentioned in changeGeneAssociation). 
If @pauljensen wants to implement it, feel free, otherwise I will create a fix that you can cherry-pick.

laurentheirendt:
Maybe it is faster when you commit and I cherry-pick? Thanks.

tpfau:
Commit `5d7af9b` from my pr924 branch should correct it.

tpfau:
@laurentheirendt Could you check the fails here? Seems like some file is missing from the test models folder for some reason (i.e. the failing has nothing to do with the changes).

PR #928, opened 2017-09-16
CBT defaults are more stringent than solvers defaults

marouenbg:
-feaopt and tolopt are set to 1e-9 which considerably slows solving of medium to large models.
-cplex defaults are set to 1e-6 (EPOPT and EPRHS)
https://tomopt.com/docs/cplexug/tomlab_cplex014.php

- []x Selected `develop` as a target branch (top left drop-down menu)

### DIFF
--- a/src/base/solvers/getCobraSolverParams.m
+++ b/src/base/solvers/getCobraSolverParams.m
@@ -49,8 +49,8 @@ end
 %% Default Values
 valDef.minNorm = 0;
 %valDef.objTol = 1e-6; %cannot find where this is being used -RF
-valDef.optTol = 1e-9;
-valDef.feasTol = 1e-9;
+valDef.optTol = 1e-6;
+valDef.feasTol = 1e-6;
 valDef.printLevel = 0;
 valDef.primalOnly = 0;
 valDef.timeLimit = 1e36; %this should indicate No time limit per default


tpfau:
Changing the solver parameters will make earlier external code that relies on them being at these values invalid or change its results. I agree that we should more clearly communicate our choices for the default parameter values, but I would keep them how they are.
If someone needs less precise solutions for large models he/she can simply use changeCobraSolverParams and set them accordingly for his analysis scripts.

rmtfleming:
Indeed, changeCobraSolverParams is the way to go instead.

marouenbg:
this is up to you, I would expect the CBT defaults to match the solver default. These values are consistent across gurobi, ibm cplex and tomlab. Here it is 1000x smaller, users would have infeasibilites like I did, without knowing where this is coming from.

http://www.gurobi.com/documentation/6.0/refman/optimalitytol.html#parameter:OptimalityTol

marouenbg:
from the CPLEX manual
https://www.ibm.com/support/knowledgecenter/en/SS9UKU_12.4.0/com.ibm.cplex.zos.help/Parameters/topics/EpRHS.html

Any number from 1e-9 to 1e-1; default: 1e-06.

If the feasibility tolerance is set too low, CPLEX® may falsely conclude that a problem is infeasible.

PR #931, opened 2017-09-18
There were grRules that were empty

dayena:
There were some grRules that were empty, causing the function to fail. Fixed now by only using non-empty cells. 


laurentheirendt:
Thanks @dayena 👍

PR #932, opened 2017-09-19
Field Merging and change to updateGenes

tpfau:
updateGenes currently breaks the model structure by completely reordering the gene field without correcting related fields. This leads to a resulting model with incorrectly sized annotation fields wrong rules. 
According to correspondence with @dayena, the intention of updateGenes is to remove surplus genes from models generated by extraction algorithms (and use of e.g. removeRxns). 
This functionality was already present in removeUnusedGenes. 
However, the current version of updateGenes also reorders the genes in alphabetic order, and eliminates duplicate genes, which is an additional function compared to removeUnusedGenes. 

To this end I introduce mergeModelFieldPositions which offers the ability to merge fields in the model.
Merging will add up numeric fields and create unions for fields which are cell arrays of cell arrays or create a concatenated list (with ; as separator), for cell arrays of strings that are merged. Char fields are left with the value of the first index provided.

### DIFF
--- a/src/reconstruction/refinement/mergeModelFieldPositions.m
+++ b/src/reconstruction/refinement/mergeModelFieldPositions.m
@@ -87,24 +87,24 @@ for i = 1:numel(fields)
     %present
     if strcmp(type,'genes')
          if isfield(modelNew,'rules')
-            for i = 1:numel(posToMerge)                
+            for j = 1:numel(posToMerge)                
                 %Replace by new position.
-                modelNew.rules = strrep(modelNew.rules,['x(' num2str(posToMerge(i)) ')'],['x(' num2str(posToKeep) ')']);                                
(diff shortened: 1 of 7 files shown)


src/reconstruction/refinement/mergeModelFieldPositions.m
```diff
@@ -0,0 +1,171 @@
+function modelNew = mergeModelFieldPositions(model,type,positions)
```
laurentheirendt: Do you have any test for `mergeModelFieldPositions`?

tpfau: Thanks for the request @laurentheirendt, otherwise I would have missed some bugs.

laurentheirendt: Great! 👍

PR #914, opened 2017-09-12
changed name of mergeModels to mergeBorgModels to avoid conflict

jtsauls:
I received the following email: 

> Hello John,
> 
> I’m Hao, a postdoc in Chalmers University of Technology. Our lab developed and curates the RAVEN toolbox (PLoS Comput Biol. 2013, 9:e1002980) that enables metabolic model reconstruction and simulation. RAVEN has originally been keeping a function ‘mergeModels’ that specifically merge two or more metabolic models into one.
> 
> Recently, I noticed that the modelBorgifier package was added to COBRA repo, and there is also one ‘mergeModels’ function that does similar work. Since RAVEN have been commonly used by the community, and often together with the COBRA in one environment. Identical names cause conflicting issues for the compatibly between RAVEN and COBRA. Here I would kindly suggest that probably you could modify the name a bit (e.g. modelMerge) so that both functions can be smoothly used by all users.
> 
> Best regards,
> Hao
> -----------------------------------------------------
> Hao Wang Ph.D.
> Postdoctoral researcher
> Division of Systems and Synthetic Biology
> Department of Biology and Biological Engineering
> Chalmers University of Technology
> Room 3005, Kemivägen 10, SE-412 96 Gothenburg
> Office: +46 (0)31 772 3136
> Web: www.sysbio.se

The above commit simply changes the name of the function mergeModels to mergeBorgModels to avoid the above described naming conflict. Testing scripts all continue to run. I updated the tutorials to reflect the new name, however, I was not able to create the correct format for the .png version of the tutorial. 

**I hereby confirm that I have:**

- [X] Tested my code on my own machine
- [X] Followed the guidelines in the [Contributing Guide](https://opencobra.github.io/cobratoolbox/docs/contributing.html)
- [X] Selected `develop` as a target branch (top left drop-down menu)

*(Note: You may replace [] with [X] to check the box)*

PR #949, opened 2017-10-03
OrderModelFields from modelBorgifier overwrites existing function

tpfau:
Changed the modelBorgifier orderModelFields function to orderModelFieldsMB, as it was overwriting the existing orderModelFields function which was used to order the fields in a model according to the field definitions.
Also updated all dependent functions.

### DIFF
--- a/src/reconstruction/comparison/modelBorgifier/orderModelFieldsMB.m
+++ b/src/reconstruction/comparison/modelBorgifier/orderModelFieldsMB.m
@@ -1,10 +1,10 @@
-function Model = orderModelFields(Model)
+function Model = orderModelFieldsMB(Model)
 % Puts the fields in the correct order in the structure.
 % Called by `verifyModel`, `readCbTModel`, calls `TmodelFields`.
 %
 % USAGE:
 %
-%    Model = orderModelFields(Model)
+%    Model = orderModelFieldsMB(Model)
 %
 % INPUTS:
 %    Model:    model structure with unsorted fiels
--- a/src/reconstruction/comparison/modelBorgifier/TmodelFields.m
+++ b/src/reconstruction/comparison/modelBorgifier/TmodelFields.m
@@ -1,7 +1,7 @@
 function fields = TmodelFields()
 % Returns the field names of the `Tmodel` structure. This is
 % to help alleviate control between different scripts.
-% Called by `cleanTmodel`, `readCbTmodel`, `verifyModel`, `organizeModelCool`, `TmodelStats`, `orderModelFields`.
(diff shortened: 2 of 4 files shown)


src/reconstruction/comparison/modelBorgifier/orderModelFieldsMB.m
```diff
@@ -1,10 +1,10 @@
-function Model = orderModelFields(Model)
+function Model = orderModelFieldsMB(Model)
```
laurentheirendt: In PR https://github.com/opencobra/cobratoolbox/pull/914, the function `functionName` has been renamed as `functionBorgName`. May you do the same here?

rmtfleming:
@jtsauls pls review this

jtsauls:
I can review. I will either change it as above, or see if the function can simply be replaced and use the native one.

tpfau:
@laurentheirendt I was going for the same naming scheme as used for verifyModelMB in the folder, but I can change it to Borg if you want.

jtsauls:
@tpfau @laurentheirendt That's my fault, I was inconsistent before.

tpfau:
@jtsauls @laurentheirendt So, which naming scheme should we use? 
Looking at it: It would actually be great if the borgifier could use the field definitions instead of defining them itself (e.g. in the compareCbModels method). If you need additional programmatic data (in this instance likely the scores), we could add those to the extracted data from the getDefinedFieldProperties method. 
The advantage would be, that this way, the code could be directly used if additional fields are introduced and consider them properly.

jtsauls:
This is clearly the ideal solution. Having arbitrary additional fields be used for comparison (based off of the field definitions) would be a significant improvement, though I think this is bigger than the scope of this PR.

At first pass, I think conforming the names and ditching at minimum using the native orderModelFields function is very doable. Thoughts?

tpfau:
@jtsauls Sure, I wasn't talking about making that change in this PR. 
I think however, that orderModelFields and the borgifier variant should not be mixed, as long as the borgifier relies on the field definitions defined in TmodelFields, as this can lead to unexpected behaviour. It would essentially make it necessary to always alter TmodelFields if the field definitions are modified. 
So I would keep the borgifier variant of orderModelFields for now ( while renaming it), and only change it once the remaining code is converted to use the field definitions.

jtsauls:
@tpfau Sounds like a plan. Let me take crack at it shortly.

tpfau:
@laurentheirendt  @jtsauls The question remains: which naming scheme should we use? `functionMB` or `functionBorg` ? 
I will update this pr accordingly (adapting everything to the choosen scheme.

laurentheirendt:
I vote for `functionNameBorg`. @jtsauls?

jtsauls:
@tpfau @laurentheirendt That sounds good to me. It is less ambiguous than MB.

PR #954, opened 2017-10-09
compatability for 2013a

tpfau:
CAUTION:
I advise against merging this PR into the master branch of the toolbox. This code is for compatability with Matlab 2013a and likely to become deprecated at some point (along with being slower, than the existing code).
My suggestion is to put this commit into a separate 2013A branch to support this outdated Matlab version, but not to move this part into the stable code,as it will very likely be deprecated by Mathworks at some point (mathworks actively discourages use of addParamValue for inputParsers and instead recommends addParameter).
This Code was purely built to allow use of Matlab 2013a with parts of the Toolbox which use inputParsers and is submitted on @laurentheirendt request.


rmtfleming:
Please can you add a if statement to check the version using
[v d] = version
then use the 2013 compatible code only if <=2013?

tpfau:
I had this discussion with @laurentheirendt and @syarra :
Adding this would create large blocks of code, that completely bypasses the CI for a version, we are nto supporting. 
I have put in this pull request, as there was a particular demand for it. As mentioned in the PR, I already don't like to have this in in general, but creating more than a hundred lines of untested code, just for an unsupported version is (imo) even worse.

laurentheirendt:
Merging as discussed with @tpfau.

Proposed solution works with all supported MATLAB versions and is backward compatible.

PR #955, opened 2017-10-10
Update to WriteCbModel to properly distinguish between old and new signature

tpfau:
writeCbModel assumed a legacy signature if no additional inputs were supplied beside the model.
This is incorrect, as a pure model input was acceptable, allowing fileName and type to be selected during later processing, which was not possible with the old signature.

Updated the default behaviour to represent this fact. Also, If the second argument is not a char array (which format was required to be in the old signature), it will now return an appropriate error.


rmtfleming:
pls merge

laurentheirendt:
Thanks!

PR #956, opened 2017-10-10
Test for optimizeTwoCbModels and enumerateOptimalSolutions

tpfau:
Test functions for optimizeTwoCbModels and enumerateOptimalSolutions.
Also updated the function to use getCobraSolverParams and readded the removed value objTol that is used in optimizeTwoCbModels to getCobraSolverParams.
Also fixed a bug in enumerateOptimalSolutions.
The test for enumerateOptimalSolutions is a little tricky, as the outputs are not guaranteed to be the same over different solvers, so This test just checks whether the outputs make sense and whether all expected properties are matched without looking into individual solutions.


tpfau:
Reopening to retrigger CI

PR #958, opened 2017-10-11
Test preprocessing

tpfau:
Test for the gene expression preprocessing pipeline


src/dataIntegration/transcriptomics/preprocessing/findUsedGenesLevels.m
```diff
@@ -35,8 +35,7 @@
     	gene_expr(i)=-1;
     elseif length(dataID)==1
     	gene_expr(i)=exprData.value(dataID);
-    elseif length(dataID)>1
-    	disp(['Double for ',num2str(cur_ID)])
+    elseif length(dataID)>1
```
laurentheirendt: May you add a `printLevel` here instead?

tpfau: Personally, I assume, that this was pure debug output, as there is no use to it.

laurentheirendt: Possible ... it would be nice to have an optional `printLevel` though in this function as well.

laurentheirendt:
@tpfau, may you rebase here? I fixed Windows 7 ...

PR #959, opened 2017-10-11
Update to testSbmlTestModelToMat

tpfau:
The previous test looked for the existence of specific xml files (which are present in the model folder and thus automatically succeed).
Also, fixed the original function to represent that the converted file should exist in the target folder (and not just anywhere) and removed the (default) inputs for readCbModel (not necessary any more).
The new test checks, whether the converted models are created and works purely in a temporary folder (which should avoid conflicts, with other test runs).


laurentheirendt:
Closing and re-opening to re-trigger CI.

PR #965, opened 2017-10-17
IO Update

tpfau:
Adjusting multiple IO Methods to properly handle multicharacter compartments.

Fixed a bug in SBML IO that allowed invalid Identifiers to be used for genes.

Corrected createModel to not add a default compartment, if there is already a multichar compartment id.

### DIFF
--- a/src/base/io/utilities/readSBML.m
+++ b/src/base/io/utilities/readSBML.m
@@ -349,6 +349,7 @@ end
 if all(~cellfun(@isempty , regexp(model.comps,'^C_')))
     %And check, that the metabolites are not using these ids.
     model.comps = regexprep(model.comps,'^C_','');
+    model.mets = regexprep(model.mets,'\[C_(.*)\]$','[$1]'); %Replace the C_ in the metabolite ids as well.
 end
 %now, check whether all metabolites start with an M_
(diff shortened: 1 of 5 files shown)


rmtfleming:
we need to make sure this support is propagated throughout the rest of the cobra toolbox where it is envisaged that [x] is the compartment rather than [xyz]?

tpfau:
True. 
However, this is already used to some extend within the toolbox (e.g. getMultiSpeciesModelId uses an (almost) identical regexp to define the compartment in a species).
I think relaxing this constraint would in general allow for easier integration of multi-tissue and multi-species models, as the compartment-ID could contain the Tissue or species assignment. 
And it does not hurt any old things, as they are still perfectly valid.

rmtfleming:
model.mets should be a primary key

tpfau:
Could you elaborate what you mean?

rmtfleming:
https://en.wikipedia.org/wiki/Unique_key

tpfau:
I know what a primary key is.
But what do you want to say with this? 
mets will always be a primary key. and Duplicates wont be allowed. But I don't understand, why you mention it.

tpfau:
done

PR #967, opened 2017-10-18
Fix Gurobi-specific parameter structure in solveCobraMILP.m 

shjchan:

Initialization section:
- assign parametersStructureFlag (used only for gurobi) to all cases to allow for Gurobi-specific parameter structure. Currently the supplied solver-specific parameter structure is not used in gurobi (Test added in testSolveCobraMILP.m)
- set default MIP start (x0) to be []. Currently MILPproblem.x0 must be supplied. But I think most solvers do not require that. (Test added in testSolveCobraMILP.m)
- fix parameter names from matching with optParamNames (currently mismatched) to matching with their names directly (line 160, 164)

gurobi section:
- Add lines to use supplied MIP start (x0). Current not used. (Test added in testSolveCobraMILP.m)

ibm_cplex section:
- Allow display to command window by supplying option logFile = 1. Currently no way to have output to command window. (Test added in testSolveCobraMILP.m)

### DIFF
--- a/src/analysis/topology/conservedMoieties/decomposeMoietyVectors.m
+++ b/src/analysis/topology/conservedMoieties/decomposeMoietyVectors.m
@@ -48,7 +48,7 @@ while rp > 0 % Iterate since decomposed moiety vectors might themselves be decom
         P.osense = 1;
         P.csense = [repmat('E', 2*size(Np,2) + length(c), 1); 'L'; 'G'; 'G'];
         P.vartype = repmat('I', size(P.A,2), 1);
-        P.x0 = suma*[c; c];
+        P.x0 = suma(:);
 
         % Run MILP
         solution = solveCobraMILP(P);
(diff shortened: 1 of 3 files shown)


shjchan:
Hi @laurentheirendt,
Know anyone who can check my edit in decomposeMoietyVectors.m?
P.x0 in line 51 was assigned a scalar equal to 4, instead of a vector with correct dimension. That was the reason for failing.

rmtfleming:
@Gpreciat

tpfau:
I'm not entirely sure, but no, what you currently have is not correct neither.
I guess, what was missing in the original code was a transposition of suma and a `.` before the `*` i.e. I assume the intention was that the initial solution is the solution, where metabolites that contain the moiety are set to 1 (which would be `suma' .* [c;c]`).

shjchan:
@tpfau , I have looked into the MILP formulation a little bit more. I think you are right and changed as you suggested. And the x0 provided is the original moiety vector, which is to be decomposed and is meant to be infeasible.

Is there some instability with the building process recently? The builts that failed seem unrelated to the changes I made. Or they are related?

laurentheirendt:
@shjchan, we are currently working on the CI. It will be back to normal soon. Thanks.

laurentheirendt:
@shjchan, may you do a rebase here? We fixed this issue over the weekend.

PR #969, opened 2017-10-19
Fva speed up

tpfau:
These changes improve speed (depending on the model size and the constraints on the model), by solving an initial optimization problem maximising/minimising all reactions that are checked.
All reactions which are at their upper or lower bound during maximisation need not be checked individually.
This speed up is however only useable when not requesting minimal flux distribution (i.e. 3rd and fourth output argument), as individual solutions are then required.

### DIFF
--- a/src/analysis/FVA/fluxVariability.m
+++ b/src/analysis/FVA/fluxVariability.m
@@ -220,8 +220,8 @@ if  any(strcmp(PCT,{v.Name})) && license('test','Distrib_Computing_Toolbox')
 else
     PCT_status=0;  % Parallel Computing Toolbox not found.
 end
-minFlux = model.lb;
-maxFlux = model.ub;
+minFlux = model.lb(ismember(model.rxns,rxnNameList));
+maxFlux = model.ub(ismember(model.rxns,rxnNameList));
 if ~minNorm
     %We don't have to provide the solutions, so we can speed this up quiet
     %a bit.
--- a/src/analysis/FVA/fluxVariability.m
+++ b/src/analysis/FVA/fluxVariability.m
@@ -245,9 +245,16 @@ QuickProblem = LPproblem;
 [Presence,Order] = ismember(rxnNameList,model.rxns);
 QuickProblem.c(:) = 0;
 QuickProblem.c(Order(Presence)) = 1;
+if ~allowLoops
+    QuickProblem = addLoopLawConstraints(QuickProblem,model,1:nRxns);
+end
 %Maximise all reactions
 QuickProblem.osense = -1;
-sol = solveCobraLP(QuickProblem);
+if ~allowLoops
+    sol = solveCobraMILP(QuickProblem);
+else
+    sol = solveCobraLP(QuickProblem);
+end
 relSol = sol.full(Order(Presence));
 %Obtain fluxes at their boundaries
 maxSolved = model.ub(Order(Presence)) == relSol;
@@ -259,7 +266,11 @@ end
 
 %Minimise reactions
 QuickProblem.osense = 1;
-sol = solveCobraLP(QuickProblem);
+if ~allowLoops
+    sol = solveCobraMILP(QuickProblem);
+else
+    sol = solveCobraLP(QuickProblem);
+end
 relSol = sol.full(Order(Presence));
 %Again obtain fluxes at their boundaries
 maxSolved = maxSolved | (model.ub(Order(Presence)) == relSol);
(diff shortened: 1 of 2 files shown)


src/analysis/FVA/fluxVariability.m
```diff
@@ -220,47 +220,121 @@
...
+    QuickProblem.c(Order(Presence)) = 1;
+    %Maximise all reactions
+    QuickProblem.osence = -1;    
+    sol = solveCobraLP(QuickProblem);
+    relSol = sol.full(Order(Presence));
+    %Obtain fluxes at their boundaries
+    maxSolved = model.ub(Order(Presence)) == relSol;    
+    minSolved = model.lb(Order(Presence)) == relSol;        
+    %Minimise reactions
+    QuickProblem.osence = 1;
```
marouenbg: This should be osense.

src/analysis/FVA/fluxVariability.m
```diff
@@ -220,47 +220,121 @@
...
+maxFlux = model.ub(ismember(model.rxns,rxnNameList));
+if ~minNorm
+    %We don't have to provide the solutions, so we can speed this up quiet
+    %a bit.
+    QuickProblem = LPproblem;
+    [Presence,Order] = ismember(rxnNameList,model.rxns);
+    QuickProblem.c(:) = 0;
+    QuickProblem.c(Order(Presence)) = 1;
+    %Maximise all reactions
+    QuickProblem.osence = -1;
```
marouenbg: This should be osense.

src/analysis/FVA/fluxVariability.m
```diff
@@ -220,47 +220,121 @@
...
 end
+minFlux = model.lb(ismember(model.rxns,rxnNameList));
+maxFlux = model.ub(ismember(model.rxns,rxnNameList));
+if ~minNorm
+    %We don't have to provide the solutions, so we can speed this up quiet
+    %a bit.
+    QuickProblem = LPproblem;
+    [Presence,Order] = ismember(rxnNameList,model.rxns);
+    QuickProblem.c(:) = 0;
+    QuickProblem.c(Order(Presence)) = 1;
```
marouenbg: you are solving for a different problem here: max/min sum(v), which is different from FVA. Think about Ecoli growth and secretion of metabolites. 
btw min sum(v) is pFBA.

src/analysis/FVA/fluxVariability.m
tpfau: thanks

src/analysis/FVA/fluxVariability.m
tpfau: Yes, and that is intentional.
The idea (which works quite well on larger models), is that a lot of fluxes are unrestricted, i.e. can reach their respective upper/lower bounds, and those are all determined by one single LP and don't need to be calculated again. 
I.e. (if you don't want individual solutions), you can simply solve this once, and get a quite large number of min and max values (most of those which are unconstraint), with 2 LPs instead of 2 per reaction LPs.

src/analysis/FVA/fluxVariability.m
tpfau: thanks

src/analysis/FVA/fluxVariability.m
marouenbg: Actually this is a cool idea.

tpfau: Thanks

tpfau:
I noticed, that this speedup also can be used for minNorm runs. 
It avoids recalculating of the min/max value of the flux and thus there are also LPs which can be avoided. The speedup is essentially the same (but since there are minimisation calculations its less on a relative scale.)

tpfau:
Need to check what goes wrong

tpfau:
reopening after fix

tpfau:
close/reopen to retrigger the CI

laurentheirendt:
Closing and re-opening to re-trigger CI.

tpfau:
Please remove the ready to merge tag again. I missed a bug, when allowLoops was set to false.

tpfau:
close/reopenn to trigger CI.

tpfau:
This should be working now. In progress can be removed.

rmtfleming:
Can the fluxVariability.m test coverage be increased from 59.48% closer to 100%? This is such a widely used function that it should be fully tested, esp with the changes proposed, which must be confirmed to work as intended.

tpfau:
It actually is tested. 
The main issue is, that the code coverage does not see code called in parfor loops, which is the main reason for the lack of tested code here. 
The only way I see to have this as part of the coverage is to  make an explicit "do not use parfor" switch, either by an additional argument, or as a hidden global variable.

PR #970, opened 2017-10-19
Corrected parsing of GPR Rules

tpfau:
This PR partially addresses #852 
GPR rules in old style cobra format could sometimes be parsed in the wrong way, as gene names that were prefixes of other gene names could lead to wrong replacements when generating rules. 
This has been fixed here, by ensuring, that only "full" IDs are replaced.

To keep the parsing consistent over different functions, I introduced a parseGPR function, that is now used over multiple different functions.  

### DIFF
--- a/src/base/parseGPR.m
+++ b/src/base/parseGPR.m
@@ -1,0 +1,46 @@
+function [ruleString, totalGeneList,newGeneList] = parseGPR( grRuleString, currentGenes)
+% Convert a GPR rule in string format to a rule in logic format. 
+% We assume the following properties of GPR Rules:
+% 1. There are no genes called "and" or "or" (in any capitalization).
+% 2. A gene name does not contain any of the following characters:
+% (),{},[],|,& and no whitespace.
+% 3. The general format of a GPR is: Gene1 or Gene2 and (Gene3 or Gene4)
+% 4. 'and' and 'or' operators as well as gene names have to be followed and preceded by either a
+% whitespace character or a opening or closing bracket, respectively. Gene
+% Names can also be at the beginning or the end of the string.
+% 
+% 
+% USAGE:
+%
+%    [newGeneList,totalGeneList,ruleString] = generateRules(grRuleString,currentGenes)
+%
+% INPUT:
+%    grRuleString:     The rule string in textual format.
+%    currentGenes:     Names of all currently known genes. Encountered
+%                      genes (column cell Array of Strings)
+% OUTPUT:
+%    ruleString:       The logical formula representing the grRuleString.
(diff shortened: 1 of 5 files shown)


tpfau:
closing to make this parsing consistent between all functions using it in some way.

src/base/parseGPR.m
```diff
@@ -0,0 +1,46 @@
...
+%                      currentGenes
+%
+% .. Author: -  Thomas Pfau Okt 2017 
+
+
+tmp = regexprep(grRuleString,'([\(\{\[])\s*','$1'); %replace all spaces after opening parenthesis
+tmp = regexprep(tmp,'\s*([\]\}\)])','$1'); %replace all spaces before closing paranthesis.
+tmp = regexprep(tmp, '([\]\}\)]\s?|\s)\s*(?i)(and)\s*?(\s?[\[\{\(]|\s)\s*', '$1&$3'); %Replace all ands
+tmp = regexprep(tmp, '([\]\}\)]\s?|\s)\s*(?i)(or)\s*?(\s?[\[\{\(]|\s)\s*', '$1|$3'); %replace all ors
+tmp = regexprep(tmp, '[\]\}]',')'); %replace other brackets by parenthesis.
```
akaviaLab: Wouldn't it make more sense to move lines 35 and 36 up before staring the replacement (lines 31-34), so the replacement lines can have much simpler regular expressions?

tpfau: yes.
Those replacements where only added as an afterthought.. I will reorder it.

akaviaLab: Also, one thing I'm not sure about - Line 34 (for example) has "\s?|\s" - isn't that redundant, since ? indicates zero or one of the preceding character?

tpfau: No its not.
For all I know, | always works on the full extend of any grouped element. 
So its either bracket followed by an optional whitespace, or just the whitespace.
In addition it will try to match the longest possible element.
This makes sure, that I'm not removing whitespaces before or after `and/or`, i.e. I prefer to keep one.
having the \s in the selection [\)\s]\s* would lead to whitespaces following a closing bracket being removed. I'm not sure if its strictly necessary but I would prefer `) | (` to `)|(` in the rules, as textual GPR rules are recreated by these.
Of course we could simply add spaces there, but keeping them makes this easier.

akaviaLab: Okay. Thanks for clarifying!

laurentheirendt:
@tpfau, may you resolve the conflict here? Thanks 👍

tpfau:
done

tpfau:
Anything else to do here or can this be merged?

PR #975, opened 2017-10-24
Fix for metabolites not present in the model

tpfau:
Metabolites not yet in the model caused an error, while previous versions allowed this.
Fixed to only consider present metabolites for duplicate checking restoring the previous functionality.


laurentheirendt:
Merging as hot fix (forum). Thanks @tpfau 👍

PR #980, opened 2017-10-26
Test for Loop Law Constraints

tpfau:
A Small test that checks whether looplaw constraints actually lead to elimination of loops.

### DIFF
--- a/src/analysis/thermo/thermoFBA/addLoopLawConstraints.m
+++ b/src/analysis/thermo/thermoFBA/addLoopLawConstraints.m
@@ -161,7 +161,7 @@ elseif method == 2 % One variables (a)
         -10000*ones(nint, 1); % v > -10000 + 10000*af
         -ones(nint,1); % e<
         -100*ones(nint, 1); % e>
-        zeros(linternal,1);];
+        zeros(linternal,1)];
     
     MILPproblem.c = [LPproblem.c;
         zeros(2*nint,1)];
--- a/src/analysis/thermo/thermoFBA/addLoopLawConstraints.m
+++ b/src/analysis/thermo/thermoFBA/addLoopLawConstraints.m
@@ -1,4 +1,4 @@
-function [MILPproblem] = addLoopLawConstraints(LPproblem, model, rxnIndex)
+function [MILPproblem] = addLoopLawConstraints(LPproblem, model, rxnIndex, method, reduce_vars)
 % Adds loop law constraints to LP problem or MILP problem.
 %
 % USAGE:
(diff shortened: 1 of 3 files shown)


src/base/solvers/buildLPproblemFromModel.m
```diff
@@ -0,0 +1,58 @@
+function [ LPproblem ] = buildLPproblemFromModel( model)
+% Builds an COBRA Toolbox LP problem structure from a COBRA Toolbox model structure. 
+%
+% USAGE:
+%
+%    LPproblem = solveCobraLP(model)
```
laurentheirendt: May you change this to:
```
LPproblem  = buildLPproblemFromModel(model)
```

src/base/solvers/buildLPproblemFromModel.m
```diff
@@ -0,0 +1,58 @@
+function [ LPproblem ] = buildLPproblemFromModel( model)
```
laurentheirendt: Formatting

src/base/solvers/buildLPproblemFromModel.m
```diff
@@ -0,0 +1,58 @@
...
+%                     * .c - Objective coeff vector
+%                     * .lb - Lower bound vector
+%                     * .ub - Upper bound vector
+%                     * .osense - Objective sense (-1 means maximise (default), 1 means minimise)
+%                     * .csense - Constraint senses, a string containting the constraint sense for
+%                       each row in A ('E', equality, 'G' greater than, 'L' less than).
+
+
+
+LPproblem.A = model.S;
```
laurentheirendt: What if `model` has a field `A`?

src/analysis/thermo/thermoFBA/addLoopLawConstraints.m
```diff
@@ -207,19 +216,19 @@
         -ones(nint,1); % e<
         %-100*ones(nint, 1); % e>
         zeros(linternal,1);];
-
+
```
laurentheirendt: May you fix these formatting issues? Thanks.

src/base/solvers/buildLPproblemFromModel.m
tpfau: A (according to the definitions) should not be a model field. A is an LP specific field, and thus a field for an LPproblem. If some model contains A, it is already an LPproblem struct..

src/analysis/thermo/thermoFBA/addLoopLawConstraints.m
tpfau: Not sure, but this is normal matlab indentation of a block. i.e. if its in a command/end block, and automatically indented, all lines in that lbock are set to a specific level.

laurentheirendt: These additional indentations just appear weirdly here on Github ... When using [atom](https://atom.io/), these indentations are automatically corrected.

tpfau: Well, they are what matlab does for automatic indentation (and I actually think it makes sense, if you have empty lines to keep them on the same indentation level. 
I can remove all the empty lines, but removing the indentation seems odd to me.

PR #984, opened 2017-10-30
Additonal Model Manipulation tests

tpfau:
Added additional model manipulation tests.

### DIFF
--- a/test/verifiedTests/reconstruction/testModelManipulation/testModelManipulation.m
+++ b/test/verifiedTests/reconstruction/testModelManipulation/testModelManipulation.m
@@ -50,6 +50,70 @@ model = addReaction(model, 'EX_glc', model.mets, sc, 0, 0, 20);
 % adding a reaction to the model (test only)
 model = addReaction(model, 'ABC_def', model.mets, 2 * sc, 0, -5, 10);
 
+%Trying to add a reaction without stoichiometry will fail.
+errored = 0;
+try   
+    addReaction(model,'NoStoich');
+catch
+    errored = 1;
+end
+assert(errored == 1);
+
+%Try adding a new reaction with two different stoichiometries
+errored = 0;
+try   
+    addReaction(model, 'reactionFormula', 'A + B -> C','stoichCoeffList',[ -1 2], 'metaboliteList',{'A','C'});
+catch
+    errored = 1;
+end
+assert(errored == 1);
+
+%Try having a metabolite twice in the metabolite list or reaction formula
+modelWAddedMet = addReaction(model, 'reactionFormula', 'Alpha + Beta -> Gamma + 2 Beta');
+assert(modelWAddedMet.S(ismember(modelWAddedMet.mets,'Beta'),end) == 1);
+
+%Try to change metabolites of a specific reaction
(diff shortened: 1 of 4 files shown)


test/verifiedTests/reconstruction/testModelManipulation/testModelManipulation.m
```diff
@@ -50,6 +50,70 @@
...
+try   
+    addReaction(model,'NoStoich');
+catch
+    errored = 1;
+end
+assert(errored == 1);
+
+%Try adding a new reaction with two different stoichiometries
+errored = 0;
+try
```
laurentheirendt: You can also test on the length of the error message:
```Matlab
try
    output1 = function1(input1, input2');
catch ME
    assert(length(ME.message) > 0)
end
```

tpfau: Yes, but this does not enure that it did in fact throw an error, as a successful pass would never look at the assert. 
And doing it after the catch, would lead not to a failing but an erroring test, as ME would be undefined

laurentheirendt: But don't you want to test that the function fails/throws an error?

tpfau: Yes.
But If I have the assert in the catch block, I don't see any difference between the function passing (while it should not), and the function erroring (which it should).
And if I refer to the exception after the catch block, instead of failing an assertion if it does not error, I will create an error (which is a different status to failing a test).

laurentheirendt: You want to test that `errored == 1`. Do I see that correctly? So instead of testing the value of an integer, I suggest to actually test on the length of the message thrown.

tpfau: I want to test, that an error is thrown, because these specific inputs should not be processed (missing or double stoichiometric information). Since there are plenty of tests that test the actual functionality of the function, I assume, that it is in general ok, if this test is passed. 
True, I could also test for the specific error message being thrown, but I don't want people having to change the test, just because they make nicer error messages.

laurentheirendt: Ok, so what you are trying to do is to enforce that an error must be thrown. You can test this with:
```Matlab
try
    addReaction(model,'NoStoich');
catch ME
end
assert(length(ME.message) > 0)
```

tpfau: Yes, but:
If no error is thrown, what happens is that this is an errored test (`Undefined variable "ME" or class "ME.message".`) , not a failed test like assert(0), which is a difference.
Also, if at any previous place in the test, there was already an ME, that would have to be cleared before coming to this point.

laurentheirendt:
```Matlab
errored = false;
try   
    addReaction(model, 'NoStoich');
catch
    errored = true;
end
assert(errored);
```

PR #988, opened 2017-10-31
Ensure tests are reading supplied models.

tpfau:
These changes introduce 2 new functions to obtain the models distributed with the toolbox (i.e. changing these functions should be sufficient when changing the folder structure for the distributed models).
It also fixes a small bug in initCobraToolbox where the model file pointed to is no longer at the referred place (and modifies it to use the aforementioned functions).

Tests were altered to use either readCbModel (if the model is supplied with the test) or getDistributedModel (if the model is part of the distributed models), to ensure, that functions are working on the currently valid model structure, and not rely on old model fields.
In the course of this change a few bugs were detected, where code was not fit for the current model structure, and these have been addressed. 
In addition, readCbModel was made more robust when reading mat files and convertOldStyleModel has a few more changes performed, if a model is missing individual fields.
Simultaneously verifyModel was made more robust when checking field dependencies (if a field depends on another field which is not present, the method no longer errors, but marks the missing field in the result).

### DIFF
--- a/docs/source/guides/testTemplate.m
+++ b/docs/source/guides/testTemplate.m
@@ -22,17 +22,31 @@ tol = 1e-8;
 solverPkgs = {'tomlab_cplex', 'glpk', 'gurobi6'};
 
 % load the model
-load([CBTDIR filesep 'test' filesep 'models' filesep 'testModel.mat'], 'model');
+%Either:
+model = getDistributedModel('ecoli_core_model.mat'); %For all models in the test/models folder and subfolders
+%or
+model = readCbModel('testModel.mat','modelName','NameOfTheModelStruct'); %For all models which are part of this particular test.
+
+%Load reference data
 load('testData_functionToBeTested.mat');
 
 %{
 % This is only necessary for tests that test a function that runs in parallel.
 % create a parallel pool
-
-poolobj = gcp('nocreate'); % if no pool, do not create new one.
-if isempty(poolobj)
-    parpool(2); % launch 2 workers
+% This is in try/catch as otherwise the test will error if no parallel
+% toolbox is installed.
+try
+    parTest = true;
+    poolobj = gcp('nocreate'); % if no pool, do not create new one.
+    if isempty(poolobj)
+        parpool(2); % launch 2 workers
+    end
+catch
+    parTest = false;
+    disp('Some info whether the test is not run if no parallel toolbox is present')
 end
+if parTest 
+% if paralell toolbox has to be present (if not, this can be left out).
 %}
 
 for k = 1:length(solverPkgs)
@@ -43,7 +57,7 @@ for k = 1:length(solverPkgs)
     if solverLPOK
         % <your test goes here>
     end
-
+    verifyCobraFunctionError(@() testFile(wrongInput));
     % output a success message
     fprintf('Done.\n');
 end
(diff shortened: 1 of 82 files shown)


tpfau:
Anything else to do on this PR? Or is it ready?

rmtfleming:
@laurentheirendt this can be merged if you are fine with it.

src/base/io/utilities/getDistributedModelFolder.m
```diff
@@ -0,0 +1,33 @@
...
+%    modelName:         The name of the model including the file extension
+%    
+%
+% OUTPUTS:
+%    modelDir:          The folder the model should eb located in.
+% 
+
+global CBTDIR
+if isempty(CBTDIR)
+    error('The Toolbox is not initialized. Cannot identfy the Model folder.\nPlease run initCobraToolbox() and try again');
```
laurentheirendt: As you did in `getDistributedModel`, may you run here:
```Matlab
global ENV_VARS
ENV_VARS.printLevel = false;
initCobraToolbox;
ENV_VARS.printLevel = true;
```
instead of throwing an error?

src/base/io/utilities/getDistributedModelFolder.m
```diff
@@ -0,0 +1,33 @@
...
+% USAGE:
+%
+%    modelDir = getDistributedModelFolder(modelName) 
+%
+% INPUT:
+%    modelName:         The name of the model including the file extension
+%    
+%
+% OUTPUTS:
+%    modelDir:          The folder the model should eb located in.
```
laurentheirendt: Typo

docs/source/guides/testTemplate.m
```diff
@@ -22,17 +22,31 @@
...
+% toolbox is installed.
+try
+    parTest = true;
+    poolobj = gcp('nocreate'); % if no pool, do not create new one.
+    if isempty(poolobj)
+        parpool(2); % launch 2 workers
+    end
+catch
+    parTest = false;
+    disp('Some info whether the test is not run if no parallel toolbox is present')
```
laurentheirendt: May you provide here an error message that developers can reuse?

docs/source/guides/testTemplate.m
```diff
@@ -22,17 +22,31 @@
...
 end
+if parTest 
+% if paralell toolbox has to be present (if not, this can be left out).
```
laurentheirendt: `parallel`

src/base/io/utilities/getDistributedModelFolder.m
tpfau: Nope. because this is called in initCobraToolbox... and this could otherwise cause an infinite loop if something goes wrong.

laurentheirendt: ok 👍

tpfau: Hmmm. Thinking about it. If setting of CBTDIR fails we have bigger issues than a infinite loop :)

tpfau:
Rebased and added requested changes

src/base/io/utilities/getDistributedModelFolder.m
laurentheirendt: Yes, true. In `changeCobraSolver`, these lines are present, which is also called in `initCobraToolbox`... Let's leave it for now.

laurentheirendt:
Thanks @tpfau 🎉

PR #990, opened 2017-11-01
Write sbml improvments

akaviaLab:
It was annoying me that writeSBML can't deal with ~ as home directory, but all other functions can. So I added a quick one-line fix to replace ~ with the environment home directory.  It should not do anything on Windows.

### DIFF
--- a/src/base/io/writeCbModel.m
+++ b/src/base/io/writeCbModel.m
@@ -212,7 +212,9 @@ switch format
         end
         %% SBML
     case 'sbml'
-        fileName = strrep(fileName, '~', getenv('HOME'));
+        if (isunix)
+            fileName = strrep(fileName, '~', getenv('HOME'));
+        end
         outmodel = writeSBML(model, fileName, input.compSymbols, input.compNames);
         %% Mat
     case 'mat'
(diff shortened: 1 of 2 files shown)


src/base/io/writeCbModel.m
```diff
@@ -212,6 +212,9 @@
         end
         %% SBML
     case 'sbml'
+        if (isunix)
+            fileName = strrep(fileName, '~', getenv('HOME'));
```
syarra: @akaviaLab, it would be safer to do something like this instead
```
if strcmp(fileName, '~') || strncmp(fileName, '~/', 2)
    if ~isempty(getenv('HOME'))
        fileName(1) = [];
        fileName = [getenv('HOME'), fileName];
    end
end
```
as done [here](http://www.mathworks.com/matlabcentral/fileexchange/28249-getfullpath). `strrep` will replace all occurrences of `~` so a `fileName` like `/home/user/tmp/allo~1/test.m` will also get updated.

Could you modify this?

Thanks

akaviaLab:
Changes implemented as requested. Thanks for the review!

PR #991, opened 2017-11-06
fix for FVA non parallel and Update to FVA test

tpfau:
Fix for a bug in non parallel FVA (parallelMode is not set) and not necessary.
Also adjusted the Test to also run without the parallel toolbox installed.


laurentheirendt:
Merging as hot fix (forum). Thanks @tpfau :+1:

PR #999, opened 2017-11-09
Update remove fields

tpfau:
Added a sanity check to catch models with invalid rules fields, as they would otherwise get stuck in the while loop.

### DIFF
--- a/src/base/io/readCbModel.m
+++ b/src/base/io/readCbModel.m
@@ -334,10 +334,10 @@ for i = 1:numel(structFields)
             else
                 % We have errors. lets see if osense/csense are missing and
                 % if, add them
-                if isfield(res.Errors, 'missingFields')
+                if isfield(res.Errors, 'missingFields') || isfield(res.Errors,'propertiesNotMatched')
                     % first, see if it contains an S matrix, only then will
                     % we add the fields.
-                    if ~any(ismember(res.Errors.missingFields, 'S'))
+                    if any(ismember(fieldnames(cfield), 'S'))
                         cfield = convertOldStyleModel(cfield, 0);
                     end
                     % if we reach this place, the conversion worked,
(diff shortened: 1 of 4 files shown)


src/base/io/readCbModel.m
```diff
@@ -353,10 +353,11 @@
                     end
                 end
             end
-        catch
+        catch ME
             % IF we are here, there was a problem in verifyModel or convertOldStyleModel, so this is
             % not a model.
-    end
+            0;
```
laurentheirendt: Why is there a `0`?

src/reconstruction/refinement/removeFieldEntriesForType.m
```diff
@@ -76,7 +76,11 @@
     end
     %update the rules fields.
     if isfield(model,'rules') %Rely on rules first  
-        %However, we first normalize the rules.
+        rulesFieldOk = verifyModel(model,'simpleCheck',true,'restrictToFields',{'rules'}, 'silentCheck', true);
+        if ~rulesFieldOk
+            error('Rules field does not satisfy the field definitions. Please check that it satisfies the definitions given <a href="https://github.com/opencobra/cobratoolbox/blob/master/docs/source/notes/COBRAModelFields.md">here</a>');
```
laurentheirendt: May you implement the error message using [hyperlink.m](https://github.com/opencobra/cobratoolbox/blob/master/src/base/hyperlink.m) such that the error message is correctly displayed when MATLAB is executed in console mode?

tpfau: Done.

src/base/io/readCbModel.m
tpfau: Debugging... 
Needed a breakpoint, so I needed some code.
Removed it.

laurentheirendt:
cc: @jennmodamio

PR #1003, opened 2017-11-10
Update change cobra solvercheck

tpfau:
Added an actual check for each solver, whether it can run (i.e. licenses are available, etc) before returning true.
Also adjusted matlab to respect the feasTol/optTol parameters in solveCobraLP and printLevel in solveCobraLP/NLP.


tpfau:
Something goes really wrong... need to test again

tpfau:
Opening again. Found the problem (the almost empty model caused a crash on glpk.)

laurentheirendt:
Thanks @tpfau 🎉

PR #1008, opened 2017-11-13
Fix change to addReaction and Test for it

tpfau:
Unfortunately df51b559fe230aafe0a427ee69a93708520a73c3 was invalid. The change modified addReaction such that the properties of subSystems were no longer correct (no longer cell arrays of cell arrays of strings).
This PR reverts the change and adds a test that this does not happen in the future. 


laurentheirendt:
Merging as hot fix. Thanks @tpfau 👍

PR #1013, opened 2017-11-14
Infer metabolite formulae under minimum inconsistency

shjchan:

Add the following functions:

1. `computeMetFormulae`: to infer the chemical formulas for metabolites without formulas using a set of metabolites with known formulae by minimizing the overall inconsistency in elemental balance. This is particularly useful for checking the molecular weight of the biomass produced in the model, which should always have MW = 1000 g/mol for prediction fidelity. Please see https://academic.oup.com/bioinformatics/article-abstract/33/22/3603/3965323

2. `findElementaryMoietyVectors`: to find all elemenetary conserved moiety vectors in a model by computing the extreme rays or a rational basis for the left null-space of the stoichiometric matrix.

3. `eleMatrixToFormulae`: to convert an elemental composition matrix into a set of chemical formulae

4. `testComputeMetFormulae`: test the new functions as well as the new functionality added in the functions below.

Extend the following functions (without changing the original behavior):

5. `computeElementalMatrix`: add the `genericFormula` flag to enable the parsing of generic formula and return the corresponding elemental matrix and elements. It parses any elements starting with 'A'-'Z', followed by 'a'-'z' or '_' of indefinite length, followed by a real number (can be -ve), and also support '()', '[]', '{}'. E.g. '([H2O]2(CuSO4))2Generic_element-0.5' can be read as H8O12Cu2S2Generic_element-0.5. This allows for computing the _in silico_ formula for biomass, naming conserved moieties with undefined formulae in the network and easily handling of charges.

6. `computeMW`: similarly add the `genericFormula` flag to calculate the molecular weight of a generic formula, with hypothetical elements not contributing to any weight.

Modify the following function:

7. `decomposeMoietyVectors`: simplify the MILP structure and the iteration procedure. The output is not changed. Pass `testMoieties` and `tutorial_atomicallyResolveReconstruction` calling this function gives the same results. (This function is called by a test code for `findElementaryMoietyVectors`)


shjchan:
Thanks @laurentheirendt . Maybe something in the tomlab_cplex in solverCobraMILP needs some fixing? I don't know. I am not familiar with that.

src/reconstruction/modelGeneration/massBalance/computeMW.m
```diff
@@ -31,75 +38,44 @@
...
-                if warnings
-                    display('Warning');
-                    display(formula)
-                    display(comp);
-                end
+% molecular weight of elements in a structure
+elementMwStruct = struct('H', 1, 'C', 12, 'N', 14, 'O', 16, 'Na', 23, 'Mg', 24, 'P', 31, ...
+    'S', 32, 'Cl', 35, 'K', 39, 'Ca', 40, 'Mn', 55, 'Fe', 56, 'Ni', 58, 'Co', 59, ...
+    'Cu', 63, 'Zn', 65, 'As', 75, 'Se', 80, 'Ag', 107, 'Cd', 114, 'W', 184, 'Hg', 202);
+elementNames = fieldnames(elementMwStruct);  % elements' names
```
tpfau: Would it make sense to:
A: create a script that extracts this information from a database and writes it to a csv and 
B: load it from that csv? 
Along those lines (apart from probably otherwise failing a test), is there a reason that the weights are given as integer values instead of the isotope averages comonly available in databases?

src/analysis/subspaces/computeElementalMatrix.m
```diff
@@ -31,73 +45,278 @@
...
+        [compounds, tok] = regexp(formula, '([A-Z][a-z]*)(\d*)', 'match', 'tokens');
+        tok = tok{1,1};
+        for j = 1:length(tok) % go through each token.
+            t = tok{1,j};
+            comp = t{1,1};
+            q = str2num(t{1,2});
+            if (isempty(q))
+                q = 1;
+            end
+            switch comp
```
tpfau: Would it make sense to update this switch into something shorter (along with the later comment of having a general Elemental Weights Map loaded from a database?
e.g. something along the lines:
```Matlab
persistent ElementalWeightMatrix
if isempty(ElementalWeightMatrix)
elementalWeightMatrix = getElementalWeightMatrix(); %Read in the data from the weight matrix
end
elements = {'C', 'N', 'O', 'H', 'P', 'Other'};
 for n = 1:length(metIDs)
      i = metIDs(n);
      formula = model.metFormulas(i);
      [compounds, tok] = regexp(formula, '([A-Z][a-z]*)(\d*)', 'match', 'tokens');
      tok = tok{1,1};
        for j = 1:length(tok) % go through each token.
            comp = t{1,1};
            q = 1;
            q = str2num(t{1,2});
            elementpos = ismember(elements,comp);
            isspecialElement = any(elementpos(1:5));
            isElement = any(ismember(ElementalWeightMatrix(:,1),comp));
            if isspecialElement
                  EMatrix(elementpos) = Ematrix(elementpos) + q;
            elseif isElement
                  EMatrix(6) = Ematrix(6) + q;
            else
                    if warnings
                        display('Warning');
                        display(formula)
                        display(comp);
                    end
          end
     end
end
```

src/reconstruction/modelGeneration/massBalance/computeMW.m
```diff
@@ -31,75 +38,44 @@
...
+            q = str2num(t{1, 2});
+            if (isempty(q))
+                q = 1;
+            end
+            mwt = 0;
+            if any(strcmp(elementNames, comp))
+                % if the element is in the structure, give it a weight
+                mwt = elementMwStruct.(comp);
+            elseif warnings
+                display('Warning');
```
tpfau: I know this is not whats done in the original version neither, but shouldn't a formula that misses an element or has unknown elements in the formula be assigned a NaN weight? 
You could return the missing parts and the weigh without them, but I think those should be additional outputs and not the default. This could possibly break a current test but I would argue that this test is invalid anyways.

src/reconstruction/modelGeneration/massBalance/eleMatrixToFormulae.m
```diff
@@ -0,0 +1,66 @@
+function metForm = convertMatrixFormulas(element, metEle, dMax)
+%Convert the matrix form of the chemical formulae into string form
```
tpfau: Could you update the documentation to fit the documentation template (in order for the autodoc being able to process it)

src/analysis/subspaces/computeElementalMatrix.m
shjchan: Hi @tpfau, thanks for the thoughtful comments. When I did this, I just tried to minimize the changes to the original code and keep the original output behavior. It would be nice to have what you suggested. I think `parse_Atomic_Weights_and_Isotopic_Compositions_for_All_Elements` contains such information but needs some processing as in `getMolecularMass`.

tpfau: I actually did not see this function before. 
as for getMolecularMass:
This could also be very much reduced given that the isotopic distribution is (for all I know) also the distribution in organic molecules. So you can really just add up weighted sums of all isotopes with non NaN Isotopic compositions.
The only manual modification would be to modify the result from `parse...` to contain H for Hydrogen, Deuterium and Tritium... 
Do you want to go for this?

src/analysis/subspaces/computeElementalMatrix.m
```diff
@@ -31,73 +45,278 @@
...
+    [yn, id] = ismember({'C', 'N', 'O', 'H', 'P'}', elements(:));
+    idPrimeElement = id(yn);
+    idOtherElement = setdiff(1:numel(elements), idPrimeElement);
+    elements = elements([idPrimeElement; idOtherElement(:)]);
+    Ematrix = Ematrix(:, [idPrimeElement; idOtherElement(:)]); 
+end
+
+end
+
+function [Ematrix, element] = parseGenericFormula(formulae, element, selfCall)
```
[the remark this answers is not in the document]
shjchan: Well, I just did not want to add new functions that give similar output to existing functions. It should be easier for users. Or do you think it is better to be standalone?

src/analysis/subspaces/computeElementalMatrix.m
```diff
@@ -31,73 +45,278 @@
...
+% OPTIONAL INPUT:
+%    element:         element from previous call to preserve the order
+%
+% OUTPUTS:
+%    Ematrix:         elemental composition matrix (#formulae x #elements)
+%    element:         cell array of elements corresponding to the columns of Ematrix
+%
+% E.g., [Ematrix, element] = parseGenericFormula({'H2O'; '[H2O]2(CuSO4)Generic_element-0.5'}) would return:
+%  element = {'H'; 'O'; 'Cu'; 'S'; Generic_element}
+%  Ematrix = [2, 1, 0, 0; 4, 6, 1, 1, -0.5]
```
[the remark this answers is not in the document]
shjchan: I will change the output for the element.
Allowing negative values was basically for simpler coding in `computeMetFormulae` (to lazily avoid additional lines for charges). In the model output I discard the charges in the elemental matrix and put them back into model.charges. But I can change for the sake of a good convention.

src/analysis/subspaces/computeElementalMatrix.m
```diff
@@ -31,73 +45,278 @@
...
+    selfCall = false;
+end
+%for recalling the original formula at the top level if there are parentheses
+%in the formula leading to iterative calling
+persistent formTopLv
+if ~selfCall
+    formTopLv = '';
+end
+
+nE_max = 150;  % maximum number of different elements
```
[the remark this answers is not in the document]
shjchan: Well, just that I hate the warning in the Matlab Editor saying the 'preallocating size to speed up' thing.

src/reconstruction/modelGeneration/massBalance/computeMW.m
shjchan: I just tried to stick to the original output. Sure it is better to do what you said, as discussed above.

src/analysis/subspaces/computeElementalMatrix.m
shjchan: I am not sure. I remember that the weights by `getMolecularMass` are a bit different from the standard molecular weights for some elements? But this should be the way to go.
And more generally, I think many of the functions here can be combined and re-organized to have a structure like:
- one function to find elemental compositions (with flags allowing only specific elements or generic elements or supporting brackets or not)
- one function to get the molecular weight of chemical elements (real number for true elements, NaN for hypothetical).
- one function to get molecular weight of metabolites (simply multiplying the output from the above two), with the part for hypothetical elements returned as residual formula not accounting for MWs as you suggested
- one function to check mass charge balance (simply multiplying the output from the first function by the S-matrix)

rmtfleming:
Nice contribution.

Modularisation of the code would be good, so long as existing functionality is unchanged.

It would be interesting to compare the output of the moiety computation with this tutorial.
https://github.com/opencobra/cobratoolbox/tree/master/tutorials/analysis/atomicallyResolveReconstruction

http://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1004999

rmtfleming:
Perhaps modularisation is the way to avoid duplication of functionality.

src/analysis/subspaces/computeElementalMatrix.m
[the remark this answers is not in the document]
tpfau: In that case you could also put in zeros(0,0)

src/analysis/subspaces/computeElementalMatrix.m
tpfau: I completely agree.
And yes, we would need to make sure, that the standard atomic weights are considered (that was my thought about obtaining the data from a database). 

A suggestion from my side:
This is all concerned with massBalance, so it should go into `reconstruction/modelGeneration/massBalance`
To keep backward compatability, existing functions should be modified to use computation from the "basic" functions you suggest.
We currently have the following functions:
- `checkMassChargeBalance` - this should (imo) be the overall Mass/Charge balance check function. This can be adapted for speedup and other improvements, but I would keep it as a basic function. 
- `getMolecularMass`: This function by default assumes, that only the predominant isotopic species is present (which is in contrast to what standard weights for molecules assume). In the code it is only ever called with this option, and a quick test shows that the alternative option doesn't work (at least it gives me an error). So I would suggest, to replace code in this function to actually use the standard weights (i.e. isotope mixture), yes this might change some things, but I would claim that this is to be preferred than obtaining potentially wrong results.
Also if there are undefined elements in the formula, without a special flag I would return NaN or [NaN, KnownWeight, UnknownElements], to make clear, that the MW is unclear and indicate the known weight parts.

- About the function for atomic element weights:
  I'd suggest to create a function that uses the output from `parse_...` to define the element weights. This function should have an option to take the average isotopoic weight or the weight of the predominant isotope (the former being the default).
One issue here: What do we do about H/D/T ? My suggestion would be: Use the average isotopic for H, and specific isotope weight for D and T (if averages are requested), assuming that D and T in a formula are quite clear to be D or T and not a general H.
- to get the elemental composition:
  there is currently no function for this, so your `parseGenericFormula` would be a good option  here. I'd actually suggest to rename it to `getElementalComposition`, and by default parse everything (including brackets), without allowing negative numbers. Only if a `formulaContainsCharge` flag (or something similar) is set, this function should assume a specific element name for `charge` and add it as an element. But otherwise I would really prohibit negative numbers. 

Does this sound reasonable?

rmtfleming:
Not sure that getMolecularMass should use the average atomic weight by default. If we later want to match up with metabolomic data, then would the predominant atomic weight not be more useful and therefore the default?

tpfau:
Phew... Difficult:
Yes, when going for metabolomics it might make sense to use a specific atom weight. But if e.g. comparing growth rates, you would need the average (as otherwise your atomic mass and thus your biomass weight is off).
But even for metabolomics, shouldn't you add in potential deviations for isotopic effects? At which point any individual mass can be right, or wrong. 
Not sure though, how the atomic mass from metabolomics would be incorporated into model analysis.

rmtfleming:
I am more thinking of people using a models properties to help analyse raw metabolomic data. We should stick with the default behaviour for now cause probably not all of the downstream code is being tested by the CI. I can recall it having an impact on the thermodynamic code.

tpfau:
Ok. Even though I would mark this as something to address, since I find it odd to ave molecular Weights returned which are different to those commonly reported in databases.

shjchan:
Hi, at this point, I added three new functions based on the discussion:
1. `getElementalWeightMatrix`
As @tpfau suggested, call `parse_Atomic_Weights_and_Isotopic_Compositions_for_All_Elements` and get the full list of atomic weights of elements. It has the flexibility to get the standard weights or the weights defined in `getMolecularMass` or from user-supplied isotope abundance profile.
2. `getElementalComposition`
the previous subfunction `parseGenericformulae`. Parse general chemical formulae into an elemental composition matrix.
3. `getFormulaWeight`
Return MW (NaN if containing unknown elements), MW for known parts and unknown elements as suggested by @tpfau . Call  `getElementalWeightMatrix` and `getElementalComposition`

Other modifications:
1. `computeElementalMatrix`
Similar to what @tpfau suggested and keep the original function output behavior. In addition, keep the extension to accept generic formulae by calling `getElementalComposition`.
2. `computeMW`
Keep the original function output behavior which just uses the mass number as molecular weight. But also allow for the extension to calculate weights for general chemical formulae using standard atomic weights or specifically defined isotope abundance profiles by calling `getFormulaWeight`.

I think keeping the backward compatibility for at least a period of time is important. May consider adding warning to functions that should be reorganized and support them for some time until reorganization.

rmtfleming:
This all sounds good overall, but the impact on existing code is hard to judge. Is there test coverage for what is being modified already? Are there test functions for the new contributions? The latter are merger criteria.

tpfau:
To address Ronans comments:
There is a test that uses computeMW, which still passes. 
copmuteElementalMatrix is only ever used in computeMW.
All other functions are new, and not used in existing code. 
What we would have to do is write a specific test for `getMolecularMass` to obtain the result it currently returns for multiple formulas (including those with artifical elements etc, even if they are errors).
Then, modify the code to call directly call `getFormulaWeight`, and ensure, that this still passes the test.
Finally we can either remove `getMolecularMass` by the appropriate calls to `getFormulaWeight` OR we can leave it in as a wrapper using a specific set of parameters. 

One other minor detail:
your test folder is currently named:
`testCompetMetFormulae` I think this should be `testComputeMetFormulae`

rmtfleming:
That sounds good wrt the tests, but why not update the functionality of getMolecularMass rather than replace it with a new function?

tpfau:
Hmmm.
What about the following:
create a Test for `getMolecularMass` (supplying 1 for the optional argument for all tests to determine the current default).
replace the code in `getMolecularMass` by the one currently in `getFormulaWeight`.
Adjust the code if the test does not directly succeed.
Replace all occurences of `getMolecularMass(formula)` in the current code (not many) by `getMolecularMass(argument,1)` to keep the original behaviour. 

I agree, that having two functions just complicates things, and `getMolecularMass` is a better fit namewise (I think).

rmtfleming:
We need to be aware that lots of people have COBRA code that is not yet in the cobra toolbox. Changing getMolecularMass(formula) to getMolecularMass(argument,1), will break their code. Why not just use getMolecularMass(formula), for the original behaviour and getMolecularMass(argument,'text') for a new behaviour where 'text' is some understandable string?

shjchan:
I agree. I will try to update `getMolecularMass` and keep its original behavior then.

shjchan:
Hi @rmtfleming and @tpfau ,
I added code to support user-supplied isotope abundance profile into `getElementalWeightMatrix` in the same fashion as in `getMolecularMass`, rewrite `getMolecularMass` to call `getElementalWeightMatrix` and support new functionalities while being completely backward compatible (the old and updated `getMolecularMass` both pass the first half of the test in `testGetMolecularMass` while only the updated passes the second half.) And updated other functions I created or modified which now depend on `getMolecularMass` for computing molecular weights.

PR #1017, opened 2017-11-15
:bug: Fixing bug for MILP solver using IBM CPLEX

lvalcarcelg:
Whenever I tried to compute an infeasible problem with ibm_cplex using solveCobraMILP, I obtained an error. This error was not produced by GLPK, which returns "nan" as the solution when trying to solve an infeasible problem.

I have included some lines of code to return a solution similar to GLPK when trying to solve an infeasible problem using ibm_cplex.

### DIFF
--- a/src/base/solvers/solveCobraMILP.m
+++ b/src/base/solvers/solveCobraMILP.m
@@ -436,11 +436,12 @@ switch solver
         end
         
         % Get results
-        try x = Result.x; catch x=zeros(size(cplexlp.Model.ub)); end %return zeros if infeasible
-        try f = osense*Result.objval; catch f=nan; end %return Nan if infeasible
         stat = Result.status;
         if (stat == 101 || stat == 102 || stat == 1)
             solStat = 1; % Opt integer within tolerance
+            % Return solution if problem is feasible, bounded and optimal
+            x = Result.x;
+            f = osense*Result.objval;
         elseif (stat == 103 || stat == 3)
             solStat = 0; % Integer infeas
         elseif (stat == 118 || stat == 119 || stat == 2)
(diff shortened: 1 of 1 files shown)


laurentheirendt:
@lvalcarcelg, great PR. Thanks! 🎉

src/base/solvers/solveCobraMILP.m
```diff
@@ -432,8 +436,8 @@
         end
         
         % Get results
-        x = Result.x;
-        f = osense*Result.objval;
+        try x = Result.x; catch x=zeros(size(cplexlp.Model.ub)); end %return zeros if infeasible
+        try f = osense*Result.objval; catch f=nan; end %return Nan if infeasible
```
tpfau: x should be empty if its infeasible. 
There are enough instances, where people do not check for the status of the solution and would use the zero vector.

lvalcarcelg: The problem with ibm_cplex is that for infeasible problems, there is no field in Results called x.
Whenever the code arrive to this line, this produces an error. The test to confirm if the solution is optimal or not is done in the following lines of code.
The change in code was done to return a solution similar to gplk with infeasible problems.

tpfau: I would simply move the lines into the first if block (i.e. after `if (stat == 101 || stat == 102 || stat == 1)`).
If the solution is not optimal, don't return something that looks like a solution. and by default (see the lines 191-194), `x` and `f` are set to `[]`.

lvalcarcelg:
@tpfau , I have changed my previous commit to return a solution only if there is an optimal solution.
thanks for your previous coments

tpfau:
looks good to me

syarra:
Thanks @lvalcarcelg for your PR! 👍

PR #1019, opened 2017-11-17
Non Metabolic Constraints

tpfau:
Update to model modification functions to cope with No Metabolic Constraints of the type:
`C * v  dsense d`.

This pr officially introduces the `C` field as constraint matrix of dimension (nCtrs * nRxns), the `ctrs` field as place for constraint ids, the `d` vector of right hand side variables and the `dsense` vector for non metabolic constraint senses.

This PR does not include any code that makes the different COBRA methods use the C Matrix, and will (currently) break the use of C in optimizeCbModel (which is not yet tested). 
@laurentheirendt will update that part of the code, and we will then start working on implementing the constraints into the different COBRA based functions (or add warnings, if they are not implemented).

These fields are also not yet considered by any IO method (except transfer in matlab structs).


tpfau:
Their name depends on the actual use case.
For ME Models: yes they are called coupling constraints.
GECKO doesn't give them a name at all. 
The original code of Ines calls them RatioReactions. 
I'm happy to also add a wrapper addCouplingConstraint, but in general they are some additional constraints, which are not directly derived from stoichiometry.

laurentheirendt:
@tpfau, may you put in another revised PR with the new name as suggested by @rmtfleming? Eventually, you can also combine the new PR with PR #1020.

Thanks! 👍

PR #1020, opened 2017-11-17
Toy model for Non Metabolic Constraints

tpfau:
This PR provides a toy model for non metabolic constraints. The constraints do influence the Objective (which cannot be above 30), while without considering the constraints it would be 40.
The basic model can easily be adapted to be of use for multiple functions.
 

rmtfleming:
They are coupling constraints.

tpfau:
I'm calling them non metabolic constraints as they are general constraints that yes, most commonly couple reactions. But they could have a different meaning/intention.

PR #1023, opened 2017-11-21
Structural changes for C * v = d constraints (v2)

tpfau:
Resubmitting with changes upon request (#1019)

Update to model modification functions to properly handle Coupling or other additional constraints of the type:
`C * v  dsense d`.

This pr officially introduces the `C` field as constraint matrix of dimension (nCtrs * nRxns), the `ctrs` field as place for constraint ids, the `d` vector of right hand side variables and the `dsense` vector for  constraint senses.

It adds the addCouplingConstraint function along with basic, reusable functions (add/removeCOBRAConstraints, which can in the future be modified to also provide options for e.g. Quadratic constraints).

This PR does not include any code that makes the different COBRA methods use the C Matrix, and will (currently) break the use of C in optimizeCbModel (which is not yet tested). 
@laurentheirendt will update that part of the code, and we will then start working on implementing the constraints into the different COBRA based functions (or add warnings, if they are not implemented).

These fields are also not yet considered by any IO method (except transfer in matlab structs).

This PR  also updates several helper functions, which needed to be adapted to the C matrix field and corrects small issues in these functions, introducing tests for them.


tpfau:
Closing for changes

tpfau:
Reopening.
modifications to coupleRxnList2Rxns reverted in the last commit. This function currently produces an LP-Problem style struct that is no longer in sync with a model struct. Since this is hard tested in some tests, I would suggest to either adapt the tests (and adapt the function to use `C * v dsense d` constraints), or rename the outputs to clearly indicate that this is no longer a  COBRA model but a COBRA LP in order to avoid confusion.

PR #1024, opened 2017-11-21
Submission of gMCS implementation

lvalcarcelg:
Implementation of the Minimal Cut Sets and genetic Minimal Cut Sets algorithms, based in:

Apaolaza, Iñigo, et al. "An in-silico approach to predict and exploit synthetic lethality in cancer metabolism." Nature Communications 8.1 (2017): 459.

Tobalina, Luis, Jon Pey, and Francisco J. Planes. "Direct calculation of minimal cut sets involving a specific reaction knock-out." Bioinformatics 32.13 (2016): 2001-2007.

von Kamp, Axel, and Steffen Klamt. "Enumeration of smallest intervention strategies in genome-scale metabolic networks." PLoS computational biology 10.1 (2014): e1003378.


src/analysis/gMCS/GPR2models.m
```diff
@@ -0,0 +1,259 @@
...
+            'printLevel',printLevel-1);
+    else
+        model = modelParser(model,genes,child,node{i},node{i},printLevel);
+    end
+end
+
+end
+
+function modelOut = reduceModel(model)
+% As the algorithm generates redundant nodes, this function is used to
```
tpfau: Wouldn't it be sufficient to activate checkDuplicate during the addReaction step? 
That way no duplicate reactions should be present.

src/analysis/gMCS/GPR2models.m
```diff
@@ -0,0 +1,259 @@
...
+%     model = removeRxns(model, model.rxns(rxn_gen_out), 'metFlag',false); 
+    % remove stoichimetri of intermediate nodes
+    model.S(node_idx,:)=0;
+    % remove output of genes to intermediate nodes
+    model.S(:,rxn_gen_out)=0;
+end
+
+% remove unused reactions and unused metabolites
+% model = removeTrivialStoichiometry(model);    % problems, removes mets 2 times
+anyRowBool = any(model.S,2);
```
tpfau: Could you use removeFieldEntriesforType here?
This would ensure, that no model fields (also in the future) are missed during the resizing.

src/analysis/gMCS/GPR2models.m
```diff
@@ -0,0 +1,259 @@
...
+% if ~isempty(selMets)
+%     model = removeMetabolites(model, selMets, false);
+% end
+% modelOut = model;
+
+end
+
+
+function model = convergeIsoforms(model, separate_isoform, printLevel)
+% Some COBRA models have isoforms instead of genes. Most of the isoforms
```
tpfau: Whats used are transcripts (at least the example you give is referring to different transcripts. Thats not necessarily an isoform (Some transcripts do have different functionality).

src/analysis/gMCS/GPR2models.m
lvalcarcelg: This function is created to remove unnecesary nodes that are generated in the process. The way the code is generated, each literal node is created as a "metabolite", and then they converge into the real demand node, the gene. This way I reduce the model to eliminate redundant nodes:
gene -> node1 -> rule1  == gen -> rule1
         -> node2 -> rule2                -> rule2

src/analysis/gMCS/calculateGeneMCS.m
```diff
@@ -0,0 +1,478 @@
...
+    for i = 1:n_tmp_related
+        pos = find(pos_set(:, 1) == tmp_related(i));
+        tmp_related(i) = pos_set(pos, 2);
+    end
+    related = tmp_related(1:n_tmp_related/2);
+    related(:, 2) = tmp_related(n_tmp_related/2+1:end);
+end
+
+% Splitting
+S = [model_struct.S -model_struct.S(:, logical(model_struct.rev))];
```
tpfau: rev is deprecated. please use `model.lb < 0`

src/analysis/gMCS/calculateMCS.m
```diff
@@ -0,0 +1,404 @@
...
+b = 1e-3;   % used to activate KnockOut constraint
+phi = 1000; % b/c;
+
+% Build the K Matrix
+[~, n_ini_rxns] = size(model_struct.S);
+K = eye(n_ini_rxns);
+
+% Splitting
+S = [model_struct.S -model_struct.S(:, logical(model_struct.rev))];
+K = [K K(:, logical(model_struct.rev))];
```
tpfau: as above

src/analysis/gMCS/buildGmatrix.m
```diff
@@ -0,0 +1,289 @@
...
+        else
+            act_model = models_or_and{i};
+            nbio = find(act_model.c);
+            rxns = act_model.rxns;
+            tmp = repmat({'DM_'}, length(rxns), 1);
+            DM = cellfun(@strfind, rxns, tmp, 'UniformOutput', false);
+            DM = ~cellfun(@isempty, DM);
+            n_DM = sum(DM);
+            DM = rxns(find(DM));
+            act_model.rev = zeros(size(act_model.ub));
```
tpfau: see above

src/analysis/gMCS/GPR2models.m
tpfau: I think I get it. Maybe convertToDNF() + reduce() on the head node would make sense here.
Like that all formulas would be disjunct normal forms. Admittedly I'm not sure, which version is faster.

src/analysis/gMCS/GPR2models.m
lvalcarcelg: We have use the concept of isoform due to the mention in this paper 
"Although Recon1 (Duarte et al., 2007) was the only human metabolic model that introduced isoform-level annotations for some genes through bibliographic research, it adopted a custom annotation for isoforms, and therefore did not allow any mapping to known identifiers from public databases. For these reasons, such annotations have been subsequently lost and then simply ignored, mapping transcriptomic data with gene-level resolution only (Ryu et al., 2015). "
https://academic.oup.com/bioinformatics/advance-article/doi/10.1093/bioinformatics/btx562/4107935

In addition, we can change the name to convergeTranscripts. The main concept is that several knockout techniques make not possible to transcribe any of the transcripts of a given gene. We have offered this oportunity for the gMCS.

tpfau: Well... Sometimes I hate simplifications (like the one done here splice-isoform = isoform and different transcripts = different isoforms). While its most often true that different transcripts have similar (or identical) function, the opposite is definitely not true (i.e. not all isoforms are encoed by the same gene).

In general, I think this is fine, but yes, renaming it would make sense to avoid confusion (also in the parameters).

laurentheirendt:
@lvalcarcelg, great submission! 👍 

May you also add a few tests for this new submission? A template can be found [here](https://github.com/opencobra/cobratoolbox/blob/master/docs/source/guides/testGuide.md).

lvalcarcelg:
I have elimiated all the references to 'model.rev' along the code, as well as I have changed all 'isoform' for 'transcripts', which is more correct as @tpfau  has highlighted.

Test file has been generated calculating the genetic Minimal Cut Sets of order 2 for Ecoli core. It will work in the future with 'ibm_cplex'. It requieres the merge of my last pull request, which is already in process.

Furthermore, we are currently working in a tutorial, which we hope will be easy to understand.

laurentheirendt:
Closing/re-opening to trigger CI.

laurentheirendt:
@lvalcarcelg, thanks for your test. 👍 

May you move it to `test/verifiedTests/analysis` instead of `test/additionalTests/testGeneMCS/testGeneMCS.m`? Only tests in `verifiedTests` are actually run by the CI.

lvalcarcelg:
I have moved the test file, which seems to work fine with "ibm_cplex" and "gurobi" 
16:48:34     'testGeneMCS/testGeneMCS'                                                      passed = true

Now we are working in the tutorial, which we hope to upload soon.

test/verifiedTests/analysis/testGeneMCS/testGeneMCS.m
```diff
@@ -0,0 +1,89 @@
...
+currentDir = pwd;
+
+% initialize the test
+cd(fileparts(which('testGeneMCS')));
+
+% define the solver packages to be used to run this test
+solverPkgs = {'ibm_cplex','gurobi'}; 
+
+% Load Ecoli core
+load([CBTDIR filesep 'test' filesep 'models' filesep 'mat' filesep 'ecoli_core_model.mat'], 'model');
```
tpfau: Could you please change this to:
`model = getDistributedModel('ecoli_core_model.mat')`
The `getDistributedModel` function allows us to better maintain potential code structure changes (e.g. if the model folder for some reason gets relocated).

rmtfleming:
Great job @lvalcarcelg

PR #1033, opened 2017-11-29
Activate the Sampler Test on Windows

tpfau:
sampleCbModel was not tested on Windows systems. 
While it is true, that the CHRR sampler uses linux binaries, and thus can't be used on a Windows OS, ACHR sampling can readily be tested on a Windows OS, and thus should be tested.

### DIFF
--- a/src/analysis/subspaces/nullspace/getNullSpace.m
+++ b/src/analysis/subspaces/nullspace/getNullSpace.m
@@ -28,13 +28,26 @@ if m>n & 0%TODO need to check this
 end
 
 gmscale=1;%by default, use geometric mean scaling of S
+archstr = computer('arch');
+archstr = lower(archstr);
+switch archstr
+    case {'win32', 'win64'}        
+        if issparse(S) && any(size(S) > 3000)
+            [Z,rankS] = sparseNull(S);
+            
+        else % for small matrices or those which are full, use the normal matlab functions
+            Z = null(full(S));
+            rankS = rank(full(S));
+        end
+    case {'glnxa64','maci64','glnxa86'}
+        nullS = nullSpaceOperator(S,gmscale,printLevel);        % forms a structure nullS.
+        rankS = nullS.rank;
+        V     = speye(n-rankS,n-rankS);       % is a sparse I of order n-rankS.
+        Z     = nullSpaceOperatorApply(nullS,V); % satisfies S*Z = 0.
 
-nullS = nullSpaceOperator(S,gmscale,printLevel);        % forms a structure nullS.
-rankS = nullS.rank;
-V     = speye(n-rankS,n-rankS);       % is a sparse I of order n-rankS.
-Z     = nullSpaceOperatorApply(nullS,V); % satisfies S*Z = 0.
+        % Check if S*Z = 0.
+end
 
-% Check if S*Z = 0.
 SZ    = S*Z;
 normSZ= norm(SZ,inf);
 
--- a/src/analysis/subspaces/nullspace/getNullSpace.m
+++ b/src/analysis/subspaces/nullspace/getNullSpace.m
@@ -44,10 +44,9 @@ switch archstr
         rankS = nullS.rank;
         V     = speye(n-rankS,n-rankS);       % is a sparse I of order n-rankS.
         Z     = nullSpaceOperatorApply(nullS,V); % satisfies S*Z = 0.
-
-        % Check if S*Z = 0.
 end
 
+% Check if S*Z = 0.
 SZ    = S*Z;
 normSZ= norm(SZ,inf);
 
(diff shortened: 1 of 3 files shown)


rmtfleming:
In nullSpaceOperator, this statement is incorrect:
         error('%s\n','Chrr sampler not compatible with windows OS.')

In getNullSpace 
There should be a section like:
archstr = computer('arch');
archstr = lower(archstr);
switch archstr
    case {'win32', 'win64'}

Which should switch to matlab's nullspace algorithm.

tpfau:
I'm wondering, whether the architecture switch could be done in `getNullSpace`. 
Because otherwise it would be necessary to recalculate the matrices required for `nullSpaceOperatorApply` from the computed matlab nullspace. 
From my understanding (maybe I'm wrong here), both `nullSpaceOperator` and `nullSpaceOperatorApply` are mainly tools to convert the LUSOL factorization output into a nullspace. 
Doing this for the matlab internal functions would be (I think) counterproductive. 

So, would a architecture switch in `getNullSpace` be acceptable here? Or would we need to figure out how to get from the computed nullspace to the values requested by `nullSpaceOperatorApply`.

rmtfleming:
yes in getNullSpace

src/analysis/subspaces/nullspace/getNullSpace.m
```diff
@@ -28,13 +28,26 @@
...
+        nullS = nullSpaceOperator(S,gmscale,printLevel);        % forms a structure nullS.
+        rankS = nullS.rank;
+        V     = speye(n-rankS,n-rankS);       % is a sparse I of order n-rankS.
+        Z     = nullSpaceOperatorApply(nullS,V); % satisfies S*Z = 0.
 
-nullS = nullSpaceOperator(S,gmscale,printLevel);        % forms a structure nullS.
-rankS = nullS.rank;
-V     = speye(n-rankS,n-rankS);       % is a sparse I of order n-rankS.
-Z     = nullSpaceOperatorApply(nullS,V); % satisfies S*Z = 0.
+        % Check if S*Z = 0.
```
laurentheirendt: Shouldn't this comment be outside the loop?

src/analysis/subspaces/nullspace/getNullSpace.m
```diff
@@ -28,13 +28,26 @@
 end
 
 gmscale=1;%by default, use geometric mean scaling of S
+archstr = computer('arch');
+archstr = lower(archstr);
+switch archstr
+    case {'win32', 'win64'}        
+        if issparse(S) && any(size(S) > 3000)
```
laurentheirendt: Shouldn't there be an error/warning for models with `~issparse(S && any(size(S) > 3000)`?

test/verifiedTests/analysis/testSampling/testSampleCbModel.m
```diff
@@ -13,84 +13,76 @@
...
-
-                        options.nStepsPerPoint = 1;
-                        options.nPointsReturned = 10;
-
-                        [modelSampling, samples, volume] = sampleCbModel(model, 'EcoliModelSamples', 'CHRR', options);
+catch
+    %No pool
+end
+% define the solver packages to be used to run this test
+solverPkgs = {'gurobi6', 'tomlab_cplex'};
```
laurentheirendt:
```
solverPkgs = {'gurobi', 'tomlab_cplex'};
```

src/analysis/subspaces/nullspace/getNullSpace.m
tpfau: No. 
They will still be computed with matlabs internal methods.
Its just that for smaller models, I would expect the matlab internal functions to still perform reasonably well (while being more precise), and only on larger models approximations should be done (and LU is essentially an approximation).

src/analysis/subspaces/nullspace/getNullSpace.m
tpfau: yes

src/analysis/subspaces/nullspace/getNullSpace.m
laurentheirendt: Just for my understanding: where are the smaller models computed with MATLAB internal functions?

tpfau: In the else clause.
```matlab
        if issparse(S) && any(size(S) > 3000)
            [Z,rankS] = sparseNull(S);
            
        else % for small matrices or those which are full, use the normal matlab functions
            Z = null(full(S));
            rankS = rank(full(S));
end
```

laurentheirendt: I still think that there should be a warning for large models ...

tpfau: Well...
We don't have it for large models under linux (which use essentially the same methodology only with an external implementation that is faster because its C based), so I don't think its necessary for windows.

laurentheirendt: Yes, under UNIX, there is no need for a warning as we have the compiled language implementation.  But on Windows, I consider it being a valuable warning ...

PR #1035, opened 2017-11-30
Further update for Coupling Constraints

tpfau:
Updated `optimizeCbModel` to work with the new structure (i.e. accept and check CvD constraints from `C`, `dsense`, `d` fields).
Fixed a bug in addCOBRAConstraint where the input dimension of a vector was not adjusted leading to an error when adding the constraint. 
Added test cases for error cases of `optmizeCbModel` and (a single) non linear constraint. This should be tested more vigorously in the future.


laurentheirendt:
As this is not done yet, I reverted PR #1023 for now. May you:
- rebase against `develop` (& solve the conflict)
- merge `tpfau/UpdateModelFields`

Thanks 👍 

cc: @syarra

tpfau:
The problem is, that a revert keeps the original commits, thus a merge doesn't do anything. 
I'll revert the revert in my commit then...

laurentheirendt:
Thanks @tpfau 👍 

Closing this temporarily. This might need a rebase after the maintenance session.

PR #1036, opened 2017-11-30
Update to rBioNet

tpfau:
Update to rBioNet.
Fixes to code that used Windows specific functions.
Automatically updates the metabolite and Reaction tables when starting and adding reactions/metabolites. (No need to press refresh)
Added options to load models from sbml
Added options to load metabolites from an SBML file.


laurentheirendt:
@tpfau, great! 🎉 

May you also add the `.m` of the tutorial? Thanks.

PR #1038, opened 2017-11-30
Tutorial: gMCS

lvalcarcelg:
Tutorial for the calculation of genetic Minimal Cut Sets.


laurentheirendt:
@lvalcarcelg, thanks for this 👍 

We are generating the PDF and the HTML automatically. May you remove them? Thanks!

laurentheirendt:
@tpfau, may you check the part here with I/O? Thanks 👍

tutorials/analysis/gMCS/tutorial_gMCS.m
```diff
@@ -0,0 +1,227 @@
...
+% gMCSs will be calculated for Recon2.v04 metabolic model$$^4$.
+%% 1. Toy example (10 sec ~ 1 min)
+% The toy example under study is presented below:
+% 
+% 
+% 
+% First, we are going to load the MAT-file which contains the metabolic network 
+% in the toy example.
+%%
+load('gMCStoyExample.mat');
```
tpfau: please use readCbModel instead of load.
This is to ensure that the models used are definitely compatible with the toolbox, and thus that the functions called are. 
Please also make sure, that this is also updated in the mlx file.

lvalcarcelg:
I have changed all files to contain readCbModel. In addition, I have deleted the MAT file which contains the toy model from the test folder. 
I prefer to have the model in the tutorial folder, for easier understanding.

tutorials/analysis/gMCS/tutorial_gMCS.m
```diff
@@ -0,0 +1,227 @@
...
+% regarding this issue can be found in Apaolaza et al., 2017(a)$$ ^2$.
+% 
+% We are now going to calculate 6 gMCSs for the human metabolic model Recon2.v04$$^4$ 
+% involving gene RRM1 (Entrez ID: 6240). To do so, the folllowing piece of code 
+% has to be executed.
+% 
+% First, we are going to load the metabolic model.
+
+global CBTDIR
+load([CBTDIR filesep 'test' filesep 'models' filesep 'mat' filesep 'Recon2.v04.mat']);
```
tpfau: Sorry, I didn't notice this one before.
I'd suggest to either use `getDistributedModel('Recon2.v04.mat')` here, which is more stable, as it will automatically update the path if that ever changes, or use `readCbModel`, if you want to be clearer in indicating the source of the model. (Again also update the mlx file)

lvalcarcelg: Thanks for the comment. I am also working in a little mistake in the test file. Now is working in my PC, so I think it will work in Jenkins.

tpfau:
Just wondering, is this "NaN" entry necessary? Or could the output just be truncated? 
Essentially, the size of the resulting vector should indicate the number of possibilities if the maximal number is larger than the number of actual existing strategies. 
I find this NaN rather unintuitive and don't really see a reason for it.

tpfau:
One other thing I noticed:
it might be good to initialize the optional_inputs struct before assigning values.
Mainly to avoid other settings being used that stem from a different run and where not removed.

iaemparanza:
Thanks @tpfau for the interesting observations! Initially, we thought it would be easier for the user to see that no more feasible solutions exist with a NaN. However, I think you are right. We are now correcting the functions to truncate the solutions when no more feasible ones exist.
In addition, we are also working on the tutorial to initialize the optional_inputs variables in order to avoid any problems.
Thanks very much for you comments!

PR #1040, opened 2017-12-02
Find MIIS

marouenbg:
findMIIS and testfindMIIS finds the Minimal Infeasible Irreducible Submodel in a given infeasible model (solution status =0). Fast and effective tool for model debugging. Only with IBM Cplex for now.
It gives back the reactions and metabolites causing the infeasiblities. The user can then check the original data and correct the reactions accordingly.


test/verifiedTests/analysis/testfindMIIS/testfindMIIS.m
```diff
@@ -0,0 +1,53 @@
...
+%     - Marouen BEN GUEBILA 02/12/2017
+
+global CBTDIR
+global ILOG_CPLEX_PATH
+
+addpath(genpath(ILOG_CPLEX_PATH));
+
+%Test presence of required toolboxes.
+v = ver;
+bioPres = any(strcmp('Bioinformatics Toolbox', {v.Name})) && license('test','bioinformatics_toolbox');
```
tpfau: Are these toolboxes really required for the test? 
From the code it looks like it should only depend on the CPLEX solver.
You could actually write an assert(changeCobraSolver('ibm_cplex') == 1), but I'm not sure, whether the solver is installed on all CI systems.
If not, the current test will pass without it actually having tested anything.
@laurentheirendt ?

laurentheirendt: Why not do:
```Matlab
solverOK = changeCobraSolver('ibm_cplex');
if solverOK
% ...
end
```

tpfau: Ok, so ibm_cplex is installed on all CI systems? 
I just was not sure about it.

src/analysis/findMIIS/findMIIS.m
```diff
@@ -0,0 +1,118 @@
...
+% Output:
+%        MIIS.rxns:     Reactions of MIIS
+%        MIIS.mets:     Mets of MIIS
+%        MIIS.rxnsStat: Status of infeasiblity of reactions
+%        MIIS.metsStat: Status of infeasiblity of metabolites
+%        please refer to this link for status meaning
+% https://www.ibm.com/support/knowledgecenter/de/SSSA5P_12.7.0/ilog.odms.cplex.help/refcallablelibrary/macros/Solution_status_codes.html
+% .. Author: - Marouen Ben Guebila 24/07/2017
+
+if ~changeCobraSolver('ibm_cplex')
```
laurentheirendt: This should read (output suppressed, silent check):
```Matlab
if ~changeCobraSolver('ibm_cplex', 'LP', 0)
    error('This function requires IBM ILOG CPLEX');
end
```

test/verifiedTests/analysis/testfindMIIS/testfindMIIS.m
```diff
@@ -0,0 +1,53 @@
...
+%     an infeasible model and returns to  Minimal Irreducible Infeasible
+%     Submodel
+%
+% Author:
+%     - Marouen BEN GUEBILA 02/12/2017
+
+global CBTDIR
+global ILOG_CPLEX_PATH
+
+addpath(genpath(ILOG_CPLEX_PATH));
```
laurentheirendt: This is not required, and shouldn't be done. The path is added and set in `changeCobraSolver`.

test/verifiedTests/analysis/testfindMIIS/testfindMIIS.m
laurentheirendt: `changeCobraSolver` solver takes care of that.

laurentheirendt: If these toolboxes are required, then this check should be in `findMIIS`, not in the test only.

src/analysis/findMIIS/findMIIS.m
```diff
@@ -0,0 +1,118 @@
+function MIIS = findMIIS(LPProblem,printLevel)
+% Finds the Minimal Irreducible Infeasible Subset (MIIS) in an infeasible
```
laurentheirendt: Documentation needs formatting.

test/verifiedTests/analysis/testfindMIIS/testfindMIIS.m
marouenbg: @tpfau cplex is not installed on windows and mac machines, thus the `if solverOK`

tpfau: ok. 
I just wasn'T sure, and if it ends up not being tested at all that would be inconvenient :)

PR #1042, opened 2017-12-04
Updated Result table

tpfau:
Updated the TestAll Result table to properly display the reason for a failed test. 


tpfau:
I currently don't have one, and the problem is, it will only show the interesting bits if tests fail, which is not so easy to do.

PR #1043, opened 2017-12-04
Utility functions

tpfau:
Added utility functions to print/display GPR Rules for a given set of Reactions. 
`findGPRFromRxns` is just a wrapper, but I think people might actually look for either printXYZ or findXYZ so, I added both options. 

### DIFF
--- a/src/analysis/exploration/findGPRFromRxns.m
+++ b/src/analysis/exploration/findGPRFromRxns.m
@@ -3,7 +3,7 @@ function gprs = findGPRFromRxns(model,rxnIDs)
 % reactions. 
 % USAGE:
 %
-%    gprs = findGPRFromRxns(mode,rxnID)
+%    gprs = findGPRFromRxns(model,rxnIDs)
 %
 % INPUTS:
 %    model:             The model to add the Metabolite batch to.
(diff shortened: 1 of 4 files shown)


src/analysis/exploration/findGPRFromRxns.m
```diff
@@ -0,0 +1,23 @@
+function gprs = findGPRFromRxns(model,rxnIDs)
+% Print the GPRs (in textual format) for the given RxnIDs
+% USAGE:
+%
+%    gprs = findGPRFromRxns(mode,rxnID)
```
laurentheirendt:
```
gprs = findGPRFromRxns(model, rxnID)
```

src/analysis/exploration/findGPRFromRxns.m
```diff
@@ -0,0 +1,23 @@
...
+%
+% INPUTS:
+%    model:             The model to add the Metabolite batch to.
```
laurentheirendt: Is this description correct?

src/analysis/exploration/findGPRFromRxns.m
tpfau: updated

src/analysis/exploration/findGPRFromRxns.m
tpfau: no... corrected.

PR #1044, opened 2017-12-04
TuneParam

marouenbg:
tuneParam and testtuneParam optimize the parameters of IBM CPLEX solver to improve runtime. MILP prgrams i.e., pFBA do not scale well, and tuning the solver parameters can improve the scaling of large programs particularly when multiple runs of the same model are required e.g., FVA.


src/analysis/tuneParam/tuneParam.m
```diff
@@ -0,0 +1,158 @@
...
+%         printLevel:    0/1/2/3
+%
+% OUTPUT:
+%         optimParam: structure of optimal parameter values directly usable as
+%                     contFunctName argument in solveCobraLP function
+% .. Author: Marouen Ben Guebila 24/07/2017
+
+if ~changeCobraSolver('ibm_cplex')
+    fprintf('This function requires IBM ILOG CPLEX');
+end
```
tpfau: Shouldn't the function return in this instance (or throw an error)?

src/analysis/tuneParam/tuneParam.m
```diff
@@ -0,0 +1,158 @@
...
+
+% Initialize the CPLEX object
+try
+    ILOGcplex = Cplex('fba');
+catch ME
+    error('CPLEX not installed or licence server not up')
+end
+
+ILOGcplex.Model.sense = 'minimize';
```
tpfau: shouldn't this depend on the osense of the LPProblem object

src/analysis/tuneParam/tuneParam.m
```diff
@@ -0,0 +1,158 @@
...
+            %CPLEXParamSet
+            cpxControl=eval(contFunctName);
+        else
+            cpxControl=[];
+        end
+    end
+end
+
+if ~isfield(LPProblem,'A')
+    if ~isfield(LPProblem,'S')
```
tpfau: An LPproblem should normally have an A matrix. i.e. a model should not be used as an LPproblem, as they are distinct things. Not sure if this flexibility is good (maybe a wrapper `tuneParamsForModel`, would be better to keep it separate?

src/analysis/tuneParam/tuneParam.m
```diff
@@ -0,0 +1,158 @@
...
+
+%cplex expects it dense
+b=full(b);
+
+
+%call cplex
+tic;
+
+%complex ibm ilog cplex interface
+if ~isempty(csense)
```
tpfau: I don't think we should accept an empty `csense` vector (i.e. an input which has the csense  field, but it is empty). So the else clause is unnecessary.

src/analysis/tuneParam/tuneParam.m
```diff
@@ -0,0 +1,158 @@
...
+    contFunctName.tune.timelimit = timelimit;
+end
+if exist('nrepeat','var')
+    contFunctName.tune.repeat = nrepeat;
+end
+if exist('printLevel','var')
+    contFunctName.tune.display = printLevel;
+end
+interface='ILOGcomplex';
+if ~exist('contFunctName','var')
```
tpfau: Shouldn't this be checked first? 
Or is the intention, that a custom function + specific parameters is not allowed?
Currently, if contFuncName is a callable function and additional arguments are supplied, this throws an error with an odd error message. If you flip the initialisation, the individual parameters would override whatever comes from the function. 
I would actually check, that contFuncName is either empty or a char, as otherwise error messages would be odd.

test/verifiedTests/analysis/testtuneParam/testtuneParam.m
```diff
@@ -0,0 +1,45 @@
...
+    cpxControl.lpmethod=4;
+    %Optimize parameter values
+    timeLimit=5;%seconds
+    nRuns=5;%runs
+    printLevel=0;
+    optimParam = tuneParam(model,cpxControl,timeLimit,nRuns,printLevel);
+
+    %test results
+    %check if automatic is the optimal solver algorithm for Ecoli
+    assert(isequal(optimParam.lpmethod,0))
```
tpfau: is this the only thing that changes in this case? 
Or would it be possible to have some additional testing for other changes?

test/verifiedTests/analysis/testtuneParam/testtuneParam.m
```diff
@@ -0,0 +1,45 @@
...
+% Author:
+%     - Marouen BEN GUEBILA 04/12/2017
+
+global CBTDIR
+global ILOG_CPLEX_PATH
+
+addpath(genpath(ILOG_CPLEX_PATH));
+
+%Load the model
+model = getDistributedModel('ecoli_core_model.mat');
```
tpfau: As mentioned above, I think there should either be a wrapper for models (which converts a model to an LPproblem, or the model should be converted in this test. Personally I think the wrapper would be better. 
I know, that a lot of people used to throw models into `solveCobraLP`, but I think thats bad practice, and for models `optimizeCbModel` should be used.

marouenbg: Hey Thomas, thanks for helping with this, I addressed most of your comments. Although, I am a bit confused about the model vs lpproblem point. See if the last commit convenes.

tpfau: What I mean with LPproblem vs Model:
The COBRA Model struct has a well defined structure, which has certain properties, while a `LPproblem` structure has other properties (e.g. a `model` - normally, has an `S` matrix, while the `LPproblem` has a general `A` matrix), similarily, a model might have other fields, that influence the behavior (e.g. the C matrix which is currently being introduced), Thus, there might be essential differences between handling the structs.
Therefore I would suggest to write a second function that should be used with a model struct, and indicate that this function should only be used with an LPproblem (and actually not check if it has an `S` matrix). 
This also (imo) helps in separating LP-specific code from model specific code.

tpfau: Oh, apart from that it looks good. :)

marouenbg: Good, are there examples of doing so in other functions?

tpfau:
The easiest is probably:
```Matlab
function optimParam = tuneParamForModel(model,varargin) 
%(Almost copy of docu here including the optional arguments, as they are just a copy...)
%... set optional Parameters)
optimParam = tuneParam(buildLPproblemFromModel(model),varargin{:});
```
This way, you build the LPproblem object, can be sure, that it is a proper LPproblem - well at least if called through this function (or an error was thrown, before you entered the function).

tpfau:
@marouenbg 
Coudl you add the function I mentioned above and change the documentation in the tuneParam function to describe `LPProblem` as `A COBRA LPproblem struct`

tpfau:
Thats not what I meant :(
I thought of having this asd an additional function.
i.e. You have one function tuneParam, which works on LPProblems, and one function tuneModelParam which works on models by just calling the tuneParam function with the converted Model.

marouenbg:
This is it for now, I will add your suggestion to my to-do list. Thanks for taking to time to review the code Thomas.

PR #1047, opened 2017-12-05
Hotfix optimize cb model

tpfau:
Small bugfix for optimizeCbModel.
model.osense was completely ignored in the function and only the supplied osenseStr argument was taken into consideration. 
Corrected the behaviour and updated the test accordingly.

### DIFF
--- a/src/analysis/FBA/optimizeCbModel.m
+++ b/src/analysis/FBA/optimizeCbModel.m
@@ -151,23 +151,23 @@ function solution = optimizeCbModel(model, osenseStr, minNorm, allowLoops, zeroN
 %    This means that it is up to the person calling `optimizeCbModel` to adapt their code to the
 %    case when no solution is returned, by checking the value of `solution.stat` first.
 
-if exist('osenseStr', 'var') % Process arguments and set up problem
-    if isempty(osenseStr)
-        osenseStr = 'max';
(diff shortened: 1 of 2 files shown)


rmtfleming:
I think that is on purpose because 1, and -1, do not necessarily mean minimise and maximise. It depends on convention. That is why we force people to say 'max' or 'min.

tpfau:
The thing is, we don't force them, as osenseStr is optional. And we define osense in the field definitions as -1 indicating maximisation and 1 indicating minimisation. Thus we should either remove the field from the model structure completely (if it is anyways ignored), or we should have the functions respect the settings if they are not overwritten. 
At least thats my take on this

rmtfleming:
Only optimizeCbModel should only look for osenseStr
The lower level of abstraction, e.g., solveCobraLP, looks for osense, which is set by optimizeCbModel.

tpfau:
Ok, so to get this right:
`osense` is part of the COBRAToolbox `LPproblem` structure, while `osenseStr` is part of the COBRA `model` structure. And those should not be confused. 
I.e. when the input to a function is `model` it expects (and only interprets) `model.osenseStr` but if the input is `LPproblem` it only accepts and interprets `osense`. Correct?

PR #1050, opened 2017-12-07
Making test13C fail if no standard solver is present

tpfau:
Making `testFitC13Data` fail if no solver which is known to produce reliable FVA solutions is present. 
This addresses the issue raised in #1039, where, due to a wrong FVA, the test ran ages.


rmtfleming:
Thanks Thomas.

PR #1052, opened 2017-12-08
Fix ChangeGeneAssociation

tpfau:
`changeGeneAssociation` missed replacing `|` and `&` by `or` and `and`, respectively.
Reusing createGrRules adding an option to only update specified fields.

Allso adding tests for the change function and introduced an equality check for logical formulas.


test/verifiedTests/reconstruction/testModelManipulation/testGenerateRules.m
```diff
@@ -15,28 +15,25 @@
 fileDir = fileparts(which('testGenerateRules'));
 cd(fileDir);
 
-modelsToTry = {'Acidaminococcus_intestini_RyC_MR95.mat', 'Acidaminococcus_sp_D21.mat', 'Recon1.0model.mat', 'Recon2.v04.mat', 'ecoli_core_model.mat', 'modelReg.mat'};
+modelsToTry = {'Recon1.0model.mat', 'ecoli_core_model.mat'};
```
laurentheirendt: Why did you remove these models for testing?

src/reconstruction/refinement/GPRLogic/Node.m
```diff
@@ -131,41 +131,96 @@ function addChild(self,childNode)
...
-        %
-        % OUTPUTS:
-        %    geneSets:    A cell array of gene Combinations that would make this node active
-        %
-        %    genePos:     A set of positions (according to the parsed
-        %                 Rules) of genes that would make this tree
-        %                 evaluate to true.
-        % 
-            dnfNode = self.convertToDNF();                                        
+            % Get all functional gene sets useable for this node
```
laurentheirendt: Any reason for these indents here?

test/verifiedTests/reconstruction/testModelManipulation/testGenerateRules.m
tpfau: Mainly because proper testing (and not just string matching, which doesn't necessarily work, as the formulas can differ in ordering), would take too long if used on all models and using Recon1 + Ecoli should be sufficient. 
In addition: Testing this with a wrong model (i.e. Recon2), doesn't make a lot of sense to me.

src/reconstruction/refinement/GPRLogic/Node.m
tpfau: err no. I'll correct this

test/verifiedTests/reconstruction/testModelManipulation/testGenerateRules.m
laurentheirendt: These models are in there for a reason ... May you put them back? Thanks.

src/reconstruction/refinement/GPRLogic/Node.m
tpfau: Actually: This is the default indentation matlab does.
So maybe keep it, as any `ctrl+A / ctrl-I` call will reintroduce it?

tpfau:
CI seems to be stuck on 2015b.
I'll close and reopen to retrigger.

PR #1054, opened 2017-12-11
Update to SBML 5.16

tpfau:
Update to SBML 5.16
This update includes:
Corrected IO for subSystems (they are now exported via the groups plugin - in accordance with BiGG models). They can now also be read from those.
Fixed issues with macOS high Sierra/Sierra and current Matlab versions. 
SBML IO should work now with all current Matlab/macOS combinations indicated by mathworks.


src/base/io/writeCbModel.m
```diff
@@ -179,6 +179,26 @@
     else
         return;
     end
+else
+    if strcmp(format,'toselect') % no format was given. try to detect from fileName.
+        [~, ~, extension] = fileparts(fileName);
+        switch extension
```
laurentheirendt: May you put this `switch` in an internal function?

initCobraToolbox.m
```diff
@@ -522,6 +527,48 @@ function initCobraToolbox()
     if ENV_VARS.printLevel
         clearvars
     end
+        
+    
+end
+
+
+function adaptMacPath()
```
laurentheirendt: May you put this function as a separate function in `base`?

tpfau:
Thanks for the comments. Is this what you wanted?

tpfau:
So all tests pass, but codecov fails and thus the default?

tpfau:
Close/Reopen to retrigger the CI

PR #1062, opened 2017-12-14
Batch Addition of Reactions/Metabolites/Genes

tpfau:
This pr introduces batch addition of reactions/metabolites and genes.
The provided functions can potentially speed up the process of adding multiple reactions quite a bit as it is no longer necessary to adapt fields multiple times when reactions are added.
The functions all provide a very simplistic method to provide information for additional fields, by just matching the fields and the name/value pairs in `varargin` (see the examples in the functions).

The pr also adds information on the type of each field, thus allowing fields to be properly initialised by the new `createEmptyField` method.
Further `getModelFieldsForType` will now properly return empty fields which can clearly be associated with one of the basic model fields (rxns, mets, genes) if it starts with the the field name without s.


tpfau:
close/reopen to retrigger CI (seems to be stuck somewhere)

laurentheirendt:
Closing temporarily. @tpfau, as discussed, may you re-open this PR after the maintenance period? Thanks! 🎄

PR #1068, opened 2017-12-20
Speedup generateRules()

laurentheirendt:
Speedup by a factor of 3-4, depending on the model. This is a serial speedup - no parallelization.

For instance, for the model `modelR204` loaded from `Recon2.v04.mat`:
- **Existing version:**
```Matlab
>> tic; newModel = generateRules(modelR204); toc;
Elapsed time is 14.444188 seconds
```

- **New version:**
```Matlab

>> tic; newModel = generateRules(modelR204); toc;
Elapsed time is 4.584043 seconds.
```


src/base/parseGPR.m
```diff
@@ -8,51 +8,61 @@
...
 % whitespace character or a opening or closing bracket, respectively. Gene
 % Names can also be at the beginning or the end of the string.
-% 
-% 
+%
+%
 % USAGE:
 %
-%    [newGeneList,totalGeneList,ruleString] = generateRules(grRuleString,currentGenes)
+%    [newGeneList,totalGeneList,ruleString] = generateRules(grRuleString,totalGeneList)
```
tpfau: The header still says `currentGenes`

laurentheirendt: true 😄

src/base/parseGPR.m
```diff
@@ -8,51 +8,61 @@
...
 %
 % INPUT:
 %    grRuleString:     The rule string in textual format.
```
tpfau: Also this is no longer true. If its a cell array now, it expects a different input, maybe this should be marked with an additional flag?

laurentheirendt: What exactly is no longer true?

tpfau: Essentially I'm wondering about line 40:
`if ~iscell..`
Why is this check necessary?

laurentheirendt: This check is necessary because of lines 20 and 31 of `generateRules`.

syarra:
Nice speedup! Thanks @laurentheirendt 🍰

PR #1073, opened 2017-12-21
Adding Preparsed Flag for further speedup

tpfau:
Further speed increase (albeit less pronounced), and explicitly allowing preparsed grRules strings (i.e. those with replace and/ors to be used by parseGPR. This was done before by generateRules but not explicitly specified in parseGPR, which could lead to problems in the future if code of parseGPR gets adjusted according to the documentation. 


syarra:
Thanks @tpfau for this PR! 🎄

PR #1074, opened 2017-12-21
Integration of CNA

laurentheirendt:
This PR includes wrappers to CNA as well as tests.

cc: @SteffKlamt, @susanGhaderi


tpfau:
To note it here:
The way CNA and the toolbox handle SBMLS is unfortunately incompatible. Since the binaries are OS specific, we have to make sure that the libsbml `OutputSBML` and `TranslateSBML` .m files are not present on the matlab path, otherwise the folder they are in will be checked for the binaries and it will, obviously, not be found. 
Thus, with CNA supplying its own set of libSBML binaries, we run into the problem that those binaries are used and it seems like they are not the latest version (i.e. not 5.16) and thus incompatible with the COBRA toolbox IO.
The easiest would be for CNA to update their binaries BUT: This would lead to older Macs potentially being unable to use SBML (compatability problems), which we handle in the toolbox by also providing the binaries for older mac OS versions.

test/testAll.m
```diff
@@ -44,6 +44,9 @@
 
 if ~isempty(strfind(getenv('HOME'), 'jenkins')) || ~isempty(strfind(getenv('USERPROFILE'), 'jenkins'))
     WAITBAR_TYPE = 0;
+
+    % add CellNetAnalyzer for testing purposes
+    addpath(genpath(getenv('CNA_PATH')));
```
tpfau: You would need a checkCNAInstallation here as well.
Otherwise libSBML usage depends on the order of the tests.

laurentheirendt: 👍

PR #1081, opened 2018-01-08
thermoModel subspace edits

rmtfleming:


src/analysis/exploration/printLabeledData.m
```diff
@@ -17,7 +17,8 @@ function printLabeledData(labels, data, nonzeroFlag, sortCol, fileName, headerRo
 % .. Authors: Markus Herrgard 6/9/06
 
 tol = 1e-9;
-format = '%g\t';
+%format = '%g\t';
```
laurentheirendt: We need to have a check here then, as this format fails on Windows (the `%g` is cross platform).

tpfau:
Not sure whether the format change is good if the fid is not standard out. 
Currently the output is a valid tab separated value format, after the change, the first column no longer fits to this scheme and all ids will have spaces added when read in.
Maybe a format switch would be good here.
If fid is an actual file: output it as tsv without padding.
If fid is std out: output it with padding (potentially dependent on the maximal size of identifier column?)

Also: The header column should have the same format as the individual columns. Otherwise the header is not indented correctly.

laurentheirendt:
@tpfau, may you branch off and commit your suggestion to see? Thanks.

tpfau:
please have a look at my [pr1081](https://github.com/tpfau/cobratoolbox/tree/pr1081) branch

laurentheirendt:
Great! 🎉

PR #1086, opened 2018-01-12
Incorporation of LRS and update to Extreme Pathway Tests

tpfau:
This PR adds the lrs binaries in the COBRA.binary repository and updates the Extreme Pathway tests.
Reuse of the model from ExtremePathways to ExtremePools


src/base/install/setupCOBRASystemPath.m
```diff
@@ -0,0 +1,31 @@
+function setupCOBRASystemPath()
```
laurentheirendt: Wouldn't it be better to rename this function to `setupCOBRAbinaryPath()` or similar?

src/analysis/topology/extremeRays/lrs/extremePathways.m
```diff
@@ -99,20 +99,16 @@
...
-        % call lrs and wait until extreme pathways have been calculated
-        systemCallText = ['lrs ' pwd filesep filename '_' suffix '.ine > ' pwd filesep filename '_' suffix '.ext'];
-        [status, result] = system(systemCallText);
-        if status == 1
-            error(['lsr failed on file ', pwd filesep filename '_' suffix '.ine']);
-        end
-    else
-        error('lrs is not installed on your machine or not in the path. Refer to http://cgm.cs.mcgill.ca/~avis/C/lrs.html to install lrs.');
+[status, result] = system('which lrs');
+if ~isempty(strfind(result, '/lrs'))
```
laurentheirendt: May you use here `filesep`?

src/base/install/setupCOBRASystemPath.m
tpfau: probably best would be "addCOBRABinaryPathToSystemPath" (even though thats quite lengthy), but would actually say what it does.

src/analysis/topology/extremeRays/lrs/extremePathways.m
tpfau: Nope. 
`which` (also on windows) returns / as file separator. I was puzzled as well, but see the comments in the other files where this is looked up.

src/base/install/setupCOBRASystemPath.m
tpfau: Would that be fine with you?

src/analysis/topology/extremeRays/lrs/extremePathways.m
laurentheirendt: Interesting ... may you put that as a note?

src/base/install/setupCOBRASystemPath.m
laurentheirendt: Sounds good 👍

laurentheirendt:
Thanks @tpfau 👍

PR #1092, opened 2018-01-25
Adjustment to FVA

tpfau:
This PR ensures that:
A: FVA is also tested for non parallel processing
B: when checking solver versions the matlab path is not reset.


laurentheirendt:
Thanks @tpfau 🎉

PR #1096, opened 2018-01-29
Update testDeletionStudy

tpfau:
Update to testDeletionStudy.
The test now includes use of the uniqueGene parameter, along with actual tests for multiple genes in the double Gene deletion study (which was completely missed in the test).


laurentheirendt:
Thanks @tpfau 🌵

PR #1097, opened 2018-01-29
gMCS Test Update

tpfau:
Adding test for some optional parameters.
Also correcting a missing field if there are no related genes (likely to be of big impact, but otherwise the test model fails). 


rmtfleming:
@fplanes

laurentheirendt:
Thanks @tpfau 👍

PR #1100, opened 2018-02-01
BugFix for dual Report from dqq

tpfau:
dqq considers the objective as an additional constraint and adds a dual value to it. 
This is "concealed" in optimizeCbModel (which strips surplus dual variables) but thereby returns the wrong duals as it strips the last and not the first dual variable. 
Fixed this in solveCobraLP.

### DIFF
--- a/src/base/solvers/solveCobraLP.m
+++ b/src/base/solvers/solveCobraLP.m
@@ -444,7 +444,10 @@ switch solver
         %        sol.y               m vector: dual variables for Ax - s = 0.
         x = sol.x;
         f = c'* x;
-        y = sol.y;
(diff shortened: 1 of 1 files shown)


laurentheirendt:
Is there any extra test we can add here?

tpfau:
Actually, is testSolveCobraLP would run all solvers through the whole test this would show up (incidentially this is why I noticed it, when I changed the test to the new format that I proposed for the test suite). 
So I don't think an additional extra test is needed, just a "adaption" of the existing test (but that will come with the new test suite (for which I need this to be in as it otherwise fails.)

laurentheirendt:
Thanks @tpfau 👍

PR #1101, opened 2018-02-01
make fluxVariability() Fail gracefully if unbounded fluxes

cladelpino:
Inspired by this post: https://groups.google.com/forum/#!topic/cobra-toolbox/CJ8YAfueV58

fluxVariability() checks the quality of the solution of the first FBA, but not of the two other "previous" problems that it solves (max/min sum of all fluxes). If there are only some unbounded (lb=-Inf or ub=Inf) fluxes, so that the first FBA is optimal, the function will fail when trying to access the solution of these other problems with an unspecific error. 


cladelpino:
I don't know why it added another commit (nothing to do with mine which was just a few lines in fluxVariability())... sorry for my poor git usage.

src/analysis/FVA/fluxVariability.m
```diff
@@ -255,7 +255,12 @@
 else
     sol = solveCobraLP(QuickProblem);
 end
+if sol.stat~=1
+    error('Maximization of the sum of all fluxes failed, even though the first FBA was optimal. Possible cause of error: unbounded fluxes');
+end
+
```
tpfau: Hmmm... We don't really need to fail at all if we have unbounded fluxes, or infinite bounds.
Essentially, these fluxes just have to be excluded from the initial problem and handled specially in the individual flux checks. i.e. `fluxVariability` would return infinite if the solution for a flux with infinite bounds is unbounded. 

The only problem is if `Vmin` and `Vmax` are requested. The question is what we return then. I personally would return a `NaN` vector.

cladelpino: You're right, I took the easy way out. I'll try to implement your suggestion.

tpfau: 👍

cladelpino: I am finding a couple of bugs, like:

An obvious one (typo, should be LPproblem): 
https://github.com/opencobra/cobratoolbox/blob/7b92e46b476cfaf1cd377cc8099b759ce222dd57/src/analysis/FVA/fluxVariability.m#L391

And this I understand is a bug (ignores even if argin is method='fba'): 
https://github.com/opencobra/cobratoolbox/blob/7b92e46b476cfaf1cd377cc8099b759ce222dd57/src/analysis/FVA/fluxVariability.m#L117-L121

How should I proceed ?

tpfau: The first one: good for spotting that. Just fix it. 

The second one: That is not a bug. 
If neither Vmin nor Vmax are requested (i.e. if nargout < 3), there is no point in calculating them, except for wasting time. Method will never change the actual min/max values. Only the solution that corresponds to it.

cladelpino: ok, I was just confused by minNorm being the name for pointing that Vmin/Vmax were required, instead of: 
https://github.com/opencobra/cobratoolbox/blob/7b92e46b476cfaf1cd377cc8099b759ce222dd57/src/analysis/FVA/fluxVariability.m#L104-L108

Which is actually unused.

tpfau: Err well... 
This function is around now for a long time and has had multiple mergers.. 
Looking at it:
could you replace the 'OutputMatrix' variables in that statement, by `minNorm` and remove these lines:

https://github.com/opencobra/cobratoolbox/blob/7b92e46b476cfaf1cd377cc8099b759ce222dd57/src/analysis/FVA/fluxVariability.m#L117-L121

as they are in an entirely unrelated if clause.... (which I assume is due to some file merge).

cladelpino: Ok, but just because you ask me :) I actually think that outputMatrix better reflects what the variable is supposed to flag.

src/analysis/FVA/fluxVariability.m
```diff
@@ -74,12 +75,14 @@
 end
 if nargin < 7
     method = '2-norm';
+elseif ~any(strcmp(method,{'FBA','0-norm','1-norm','2-norm','minOrigSol'}))
+    error('Argument ''method'' must be one of {FBA,0-norm,1-norm,2-norm,minOrigSol}');
```
tpfau: I would also put the arguments in single quotes

src/analysis/FVA/fluxVariability.m
```diff
@@ -281,19 +295,28 @@
...
+Vmax=zeros(nRxns,length(rxnNameList));
+Vmin=zeros(nRxns,length(rxnNameList));
 
 if ~PCT_status && (~exist('parpool') || poolsize == 0)  %aka nothing is active
     if minNorm
         for i = 1:length(rxnNameList)
-        
             LPproblem.osense = 1;
-            [minFlux(i),Vmin(:,i)] = calcSolForEntry(model,rxnNameList,i,LPproblem,0, method, allowLoops,printLevel,minNorm,cpxControl,preCompMinSols{i});
+            if(~isNegUnb(i))
```
tpfau: The calculation should be done in all instances (an unbounded reaction does not necessarily have unbounded) fluxes.  (Imagine A -> B -> C ->D , if A -> B has a maximal bound of 5 all others have a max bound of 5 as well, even if their ub is infinite.)

src/analysis/FVA/fluxVariability.m
```diff
@@ -281,19 +295,28 @@
     preCompMinSols((model.lb(Order(Presence)) == relSol)) = {sol};
 end
 %Restrict the reactions to test only those which are not at their boundariestestFv.
-rxnListMin = rxnNameList(~minSolved);
-rxnListMax = rxnNameList(~maxSolved);
+%isNegUnb/isPosUnb were row vectors to match origPresence
+rxnListMin = rxnNameList(~minSolved & ~isNegUnb);
```
tpfau: Don't exclude them here. 
as mentioned below, they can still have bounded fluxes.

src/analysis/FVA/fluxVariability.m
```diff
@@ -305,62 +328,72 @@
...
             changeCobraSolver(qpsolver,'QP',0,1);
             changeCobraSolver(lpsolver,'LP',0,1);
-            changeCobraSolver(milpsolver,'MILP',0,1);                
-            parLPproblem = LPproblem;        
+            changeCobraSolver(milpsolver,'MILP',0,1);
+            
+            parLPproblem = LPproblem;
             parLPproblem.osense = 1;
-            [minFlux(i),Vmin(:,i)] = calcSolForEntry(model,rxnNameList,i,parLPproblem,1, method, allowLoops,printLevel,minNorm,cpxControl,preCompMinSols{i});
+            if(~isNegUnb(i))
```
tpfau: see above.

src/analysis/FVA/fluxVariability.m
```diff
@@ -370,73 +403,72 @@
...
-        Flux = getObjectiveFlux(LPsolution, LPproblem);        
+        LPsolution = solveCobraMILP(addLoopLawConstraints(LPproblem, model));
     end
-    % minimise the Euclidean norm of the optimal flux vector to remove loops -Ronan
-    if minNorm == 1
-        V = getMinNorm(LPproblem, LPsolution, nRxns, Flux, model, method);
+    % take the maximum flux from the flux vector, not from the obj -Ronan
+    % A solution is possible, so the only problem should be if its
+    % unbounded and if it is unbounded, the max flux is infinity.
+    if LPsolution.stat == 2
```
tpfau: You would have to check here, whether the flux is actually unbounded. 
If its not, then this is an error. Only if it is unbounded (in this direction), it is likely that this is a infinite solution.

tpfau:
As a suggestion:
Write a small test case with infinite bounds:
```
R1: A <-> ; lb: -inf; ub: 10
R2: A <-> B ; lb: -inf; ub: inf;
R3: B <-> C ; lb: -inf; ub : inf;
R4: C <-> D ; lb: -20; ub : inf;
R5: C <->    ; lb: - inf ; ub : 100;
R6: D <->   ; lb: -inf ; ub : inf;
```
[minFlux, maxFlux] should be:
```-inf , 10
-inf , 10;
-inf , 10;
-20, inf;
-inf, 100;
-100, inf;
```
Adding this to the fluxVaribility Test would ensure that your code is actually tested, as the current tests don't cover this at all.

src/analysis/FVA/fluxVariability.m
cladelpino: yes, understood! 👍

tpfau:
eeks... small mistake on that Example..
``R1: A <-> ; lb: -10; ub: inf
R2: A <-> B ; lb: -inf; ub: inf;
R3: B <-> C ; lb: -inf; ub : inf;
R4: C <-> D ; lb: -20; ub : inf;
R5: C <->    ; lb: - inf ; ub : inf;
R6: D <->   ; lb: -inf ; ub : 100;
```
And min/max : 
```
R1: -10 , inf;
R2: -inf , 10;
R3: -inf, 10;
R4: -20, 100;
R5: -inf, inf;
R6: -20, 100;
```
(I hope I didn't make a mistake again... :/)

laurentheirendt:
@cladelpino, your branch `patch-1`is derived from `master`. That is why you are having [this commit](7b92e46b476cfaf1cd377cc8099b759ce222dd57). 

May you branch-off from `develop` and open a new PR (or push with force to this PR branch)? Thanks.

cladelpino:
Oh, thanks!, I think a new PR would be the best way since with the help of
@tpfau the scope has changed.

El 6/2/2018 08:47, "Laurent Heirendt" <notifications@github.com> escribió:

@cladelpino <https://github.com/cladelpino>, your branch patch-1is derived
from master. That is why you are having this commit
<http://7b92e46b476cfaf1cd377cc8099b759ce222dd57>.

May you branch-off from develop and open a new PR (or push with force to
this PR branch)? Thanks.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/opencobra/cobratoolbox/pull/1101#issuecomment-363338244>,
or mute the thread

laurentheirendt:
Thanks @cladelpino 👍 You may simply cherry-pick the comments from this branch. Let me know if you need help.

PR #1104, opened 2018-02-05
Adding key to known hosts - installDevTools

laurentheirendt:
- adding key to known host when installing [MATLAB.devTools](https://github.com/opencobra/MATLAB.devTools)
- various bug fixing


src/base/install/addKeyToKnownHosts.m
```diff
@@ -0,0 +1,61 @@
...
+    if ispc
+        homeDir = getenv('userprofile');
+    else
+        homeDir = getenv('HOME');
+    end
+
+    if status_keyscan == 1 && ~isempty(strfind(result_keyscan, 'usage:'))
+
+        % touch the file first
+        system(['touch ', homeDir, filesep, '.ssh', filesep, 'known_hosts']);
```
tpfau: Is this the same on windows and unix os?
Also, it is not necessarily present on a windows system (thus you would have to create the folder first).
Not sure what the normal "search path" for known ssh key on a windows system is...

src/base/install/addKeyToKnownHosts.m
```diff
@@ -0,0 +1,61 @@
...
+                fprintf(result_kh);
+                error([' > ', siteName, ' could not be added to the known hosts file in ~/.ssh/known_hosts. \n']);
+            end
+        else
+            fprintf([' > ', siteName, ' is already a known host.\n']);
+            keyAdded = true;
+        end
+    else
+        fprintf(result_keyscan);
+        error(' > ssh-keyscan is not installed.');
```
tpfau: Would a link on how to install it be useful? (e.g. to gitbash?)

test/verifiedTests/base/testInstall/testAddKeyToKnownHosts.m
```diff
@@ -0,0 +1,58 @@
...
+    [~, result_grep] = system(['grep "^github.com " ', homeDir, filesep, '.ssh', filesep, 'known_hosts']);
+
+    % test whether the resulting string is not empty (site has been found)
+    assert(~strcmp(result_grep, ''))
+
+    % test if the host is already known
+    assert(addKeyToKnownHosts('github.com'));
+
+    % remove the old host file
+    delete([homeDir filesep '.ssh' filesep 'known_hosts.old']);
```
tpfau: where is this file created?
And shouldn't this test reset the old hosts file? Like this, the test modifies the system config permanently....

src/base/install/addKeyToKnownHosts.m
laurentheirendt: It is the same on windows and unix. I tested on both.

As it my not be present, the file is `touched` before. The path is the same on all platforms when using `gitBash`.

src/base/install/addKeyToKnownHosts.m
laurentheirendt: `ssh-keyscan` is part of the tools installed with `gitBash`.

test/verifiedTests/base/testInstall/testAddKeyToKnownHosts.m
laurentheirendt: This file is created when running `ssh-keyscan -R`.

src/base/install/addKeyToKnownHosts.m
tpfau: The path is not present on the Win7 I'm using for testing, which also has git-bash installed. is it possible, that this is again only present if ssh was used for git checkout?

laurentheirendt: It is for sure present when having cloned with `git@github.com`. When the directory `.ssh` is not present and the file `known_hosts` does not exist, it is created when running:
```
touch ~/.ssh/known_hosts
```

laurentheirendt: On Win7, we may need to create the folder first.

tpfau: The reason for my comment seems to stem from an odd system configuration, where the .ssh folder is not in `%userprofile%` but a different folder. 
Not sure, but maybe it would be necessary to check where the .ssh folder actually is to properly set it up.

src/base/install/addKeyToKnownHosts.m
```diff
@@ -34,6 +34,14 @@
 
     if status_keyscan == 1 && ~isempty(strfind(result_keyscan, 'usage:'))
 
+        % try to create the directory
+        try
+            w = warning ('off','all');
+            mkdir([homeDir filesep '.ssh']);
+            w = warning ('on','all');
```
syarra: To silence the warning of `mkdir` in MATLAB, you can use this one-liner:
```
[status, msg, msgID] = mkdir([homeDir filesep '.ssh']);
```

PR #1109, opened 2018-02-08
Allows gene merging genes to be optional in createMultipleSpeciesModel

almut-heinken:
I changed the function createMultipleSpeciesModel to make the merging of genes optional. It causes the function to take much longer and is usually unnecessary.


laurentheirendt:
Thanks @almut-heinken 👍 

May you have a look at the error?

```
18:30:20 ================================================================================
18:30:20 Error occurred in testCreateMultipleSpeciesModel/testCreateMultipleSpeciesModel and it did not run to completion.
18:30:20 
18:30:20     ---------
18:30:20     Error ID:
18:30:20     ---------
18:30:20     'MATLAB:assertion:failed'
18:30:20 
18:30:20     --------------
18:30:20     Error Details:
18:30:20     --------------
18:30:20     Error using testCreateMultipleSpeciesModel (line 135)
18:30:20     Assertion failed.
18:30:20 ================================================================================
```

laurentheirendt:
@almut-heinken, I removed your last commit, as there were issues with the submodules (don't add them). 

May you only add the updated test function? Thanks.

almut-heinken:
Hi Laurent, sorry, that was by accident. How do I remove a commit afterwards? It is still in my fork.

almut-heinken:
@tpfau  That currently creates the problem though that the new keywords can only be set if a host model is entered. I will rewrite it to work around that.

almut-heinken:
@laurentheirendt @tpfau  I have opened a new pull request where the gene flag is at the end of the list. Please close this one. Sorry for the inconvenience. I think there shouldn't be problems with other scripts using this function now.

tpfau:
> @tpfau That currently creates the problem though that the new keywords can only be set if a host model is entered. I will rewrite it to work around that.

Just a comment: Thats why I personally prefer  parameter/value pair parameters for anything with more than 1 optional argument. Param/Value allows you to later add additional options and make code better readable (since you can see which parameters were set to which value).

PR #1113, opened 2018-02-13
Small fix for the parameter names in writeCbModel documentation

shjchan:
The keywords for compartment symbols and compartment names should be `compSymbols` and `compNames` instead of `compSymbolList` and `compNameList` as in line 116 to 117. `testWriteCbModel` also uses these keywords.


laurentheirendt:
Thanks @shjchan 👍

Issue #756, opened 2017-07-12
RelaxedFBA tutorial bugs


shjchan:
@fromseto and I have tried to run the latest version (as of July 12, 2017) of `tutorial_relaxedFBA.mlx` from the `develop` branch. We found that:
1. The .mlx and .m files do not read the same (.mlx is more recently updated).
2. In the .mlx file, there are bugs in the first code block loading the 'Recon2.0model':
```
filename2='Recon2.0model';
if exist('Recon2.0model.mat','file')==2
      model = readCbModel(fileName);
end
```
`fileName`, `filename2` do not match and 'readCbModel' is unable to handle filename with '.' inside. This can be fixed easily by using `load` instead.
 

3. But the default setting in the tutorial for Recon2.0 model results in an error in both of our trials. The code leading to the error in the tutorial is:
```
relaxOption.internalRelax = 0;
relaxOption.exchangeRelax = 2;
relaxOption.steadyStateRelax = 0;
feasTol = getCobraSolverParams('LP', 'feasTol');
relaxOption.epsilon = feasTol/100;%*100;
relaxOption.gamma  = 10;
relaxOption.lambda = 10;  
tic;
solution = relaxedFBA(model,relaxOption);
```
Then the following error occurred:
```
Warning: solveCobraLP solution status is 0 
> In relaxFBA_cappedL1>relaxFBA_cappedL1_solveSubProblem (line 348)
  In relaxFBA_cappedL1 (line 170)
  In relaxedFBA (line 211) 
Error using relaxFBA_cappedL1 (line 179)
Problem infeasible !
Error in relaxedFBA (line 211)
solution = relaxFBA_cappedL1(model,relaxOption);
``` 
But if we set `relaxOption.internalRelax = 2` , then relaxFBA finishes without error.
We are not sure whether it needs to be fixed, or it is the problem of the particular model.

**I hereby confirm that I have:**

- [X] Tried to solve the issue on my own
- [X] Retried to run my code with the latest version of The COBRA Toolbox
- [X] Checked that a similar issue has not already been opened

*(Note: You may replace [] with [X] to check the box)*

marouenbg:
Indeed, relaxedFBA doesn´t handle lb < 1 which was the case for Recon2 bounds, this is now fixed in #786 . A workaround would have been to multiply the bounds by 1e9. Thanks for reporting!

laurentheirendt:
@shjchan, thanks for reporting this. 

[...]

Issue #551, opened 2017-06-15
relaxFBA doesn't give minimal relaxation


tpfau:
I tried to test relaxFBA on a small model (basically:
```
-> A -> B -> C ->
        |    ^
        v  /
        D
```
setting the lower bound of B -> D  to 2, the lower bound of B -> C to 2 and the upper bound of C -> to 2 as well, (which gives an infeasible solution).
Now, relaxFBA indicates, that it minimizes the number of alterations needed to make this problem feasible. 
However, when running relaxFBA, I get an increase of the upper bound of C-> by 998 and decreases of the lower bound of B->D and B -> C by 1002 as minimal solution which is (quite obviously) not the minimal number of interventions (which would be just increasing the upper bound of C -> to something above 4 or decreasing either of the lower bounds to something <= 0. 

Anyone with a better grasp on relaxFBA an idea what is going wrong here?

**I hereby confirm that I have:**

- [X] Tried to solve the issue on my own
- [X] Retried to run my code with the latest version of The COBRA Toolbox
- [X] Checked that a similar issue has not already been opened

rmtfleming:
please post here your script

tpfau:
```Matlab
initCobraToolbox
changeCobraSolver('gurobi','all')
rxnForms = {' -> A','A -> B','B -> C', 'B -> D','D -> C','C ->'}
rxnNames = {'R1','R2','R3','R4','R5', 'R6'};
model = createModel(rxnNames, rxnNames,rxnForms)
model.lb(3) = 1;
model.lb(4) = 2;
model.ub(6) = 2;
res = relaxFBA(model)
```

rmtfleming:
>> model = createModel(rxnNames, rxnNames,rxnForms)
Warning: Metabolite A[c] not in model - added to the model 
> In addReaction (line 276)
  In createModel (line 139) 
Error using xlsread (line 121)
File name must be specified.
Error in getDefinedFieldProperties (line 131)
    [~,~,raw] = xlsread(fileName,'Programatic Specification');
Error in getRelevantModelFields (line 98)
    fields = getDefinedFieldProperties();
Error in updateRelevantModelFields (line 57)
fields = getRelevantModelFields(model, type, originalSize);
Error in addMetabolite (line 211)
newmodel = updateRelevantModelFields(model,'mets');
Error in addReaction (line 277)
        model = addMetabolite(model,metaboliteList{i},metaboliteList{i});

Error in createModel (line 139)
[...]

PR #1116, opened 2018-02-21
feasopt

marouenbg:
feasOpt is IBM CPLEX routine to relax an infeasible model through minimal 0-norm, 1-norm, and 2-norm.

### DIFF
--- a/src/analysis/feasOpt/feasOpt.m
+++ b/src/analysis/feasOpt/feasOpt.m
@@ -1,0 +1,182 @@
+function solution = feasOpt(LPProblem,lhs,rhs,lb,ub,mode,tol,Rxnscore)
+%
+% feasOpt is IBM CPLEX routine to relax an infeasible model.
+% It returns a feasible/optimal solution and a minimal relaxation. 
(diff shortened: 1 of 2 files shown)


rmtfleming:
Different algorithmic implementations of the same function should use the same function interface. Please work it into https://github.com/opencobra/cobratoolbox/blob/master/src/analysis/relaxedFBA/relaxedFBA.m
instead.

marouenbg:
Hi Ronan,
yes I thought about it in the beginning, but relaxedFBA has long-standing issues, 
https://github.com/opencobra/cobratoolbox/issues/756
https://github.com/opencobra/cobratoolbox/issues/551
so I thought about having a separate function until it is fixed, in order to have functional code in the master branch.

rmtfleming:
Yes, they are minor and on my list of priorities, but not at the top. I have any way released them before publication.

Different algorithmic implementations of the same function should use the same function interface. Please work it into https://github.com/opencobra/cobratoolbox/blob/master/src/analysis/relaxedFBA/relaxedFBA.m
instead.

PR #1122, opened 2018-02-26
Update install devTools

laurentheirendt:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -113,9 +113,9 @@ within MATLAB by typing:
 
 |bulb| Check out the `MATLAB.devTools
 <https://github.com/opencobra/MATLAB.devTools>`__ - and contribute the smart
-way! The official documentation is `here <https://opencobra.github.com/MATLAB.devTools>`__.
+way! The **official documentation** is `here <https://opencobra.github.com/MATLAB.devTools>`__.
 
-Contribute to the `opencobra/cobratoolbox` repository by following `these
+|thumbsup| Contribute to the ``opencobra/cobratoolbox`` repository by following `these
 instructions
 <https://opencobra.github.io/MATLAB.devTools/stable/contribute.html#the-cobra-toolbox>`__:
 
@@ -123,7 +123,7 @@ instructions
 
     >> contribute('opencobra/cobratoolbox');
 
-Contribute to the `opencobra/COBRA.tutorials` repository by following `these
+|thumbsup| Contribute to the ``opencobra/COBRA.tutorials`` repository by following `these
 instructions
 <https://opencobra.github.io/MATLAB.devTools/stable/contribute.html#cobra-tutorials>`__:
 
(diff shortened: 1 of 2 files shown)


src/base/install/installDevTools.m
```diff
@@ -1,15 +1,17 @@
...
+function installDevTools(repoName)
 % checks if the MATLAB.devTools are already present and installs the
 % MATLAB.devTools next to the cloned folder cobratoolbox (if not present yet)
 %
 % USAGE:
-%    installDevTools()
+%    installDevTools(repoName)
+%
+% INPUT:
+%    repoName:     Name of the repository (default: 'opencobra/cobratoolbox')
```
tpfau: If there is a default, then this is an optional input.

README.rst
```diff
@@ -111,7 +111,25 @@ within MATLAB by typing:
 
     >> installDevTools()
 
-|bulb| Check out the `MATLAB.devTools <https://github.com/opencobra/MATLAB.devTools>`__ - and contribute the smart way!
+|bulb| Check out the `MATLAB.devTools
+<https://github.com/opencobra/MATLAB.devTools>`__ - and contribute the smart
+way! The **official documentation** is `here <https://opencobra.github.com/MATLAB.devTools>`__.
```
tpfau: this link <https://opencobra.github.com/MATLAB.devTools> (at least at the moment) does not work. 
I guess it should be:
in any case this should be (I think) https://opencobra.github.io/MATLAB.devTools, but even that link does not work.

src/base/install/installDevTools.m
```diff
@@ -99,7 +106,7 @@ function installDevTools()
             % removing .git folder
             rmpath(genpath([devToolsDir filesep '.git']));
 
-            fprintf([' > In order to start contributing, you can now run:\n\n >> contribute\n\n']);
+            fprintf([' > In order to start contributing, you can now run:\n\n >> contribute(''' repoName '''); \n\n']);
```
tpfau: One question:
Will it be necessary to always indicate the repoName when using contribute? Or will this (by default) be detected automatically based on the git information in the current folder?

src/base/install/installDevTools.m
laurentheirendt: True.

README.rst
laurentheirendt: Once the regular merge on the MATLAB.devTools has been made, these links will work. We are aware of that.

src/base/install/installDevTools.m
laurentheirendt: It is defined as follows:
```
>> contribute
```
is the same as
```
>> contribute('opencobra/cobratoolbox')
```
When contributing to any other repo:
```
>> contribute('orgName/repoName')
```

syarra:
Great! Thanks @laurentheirendt 🌵

PR #1132, opened 2018-03-14
Allowing files to fail gracefully

akaviaLab:
Under certain conditions of running fastcore, it reaches a situation where it prints a warning (using fprintf) and returns, since the algorithm can't continue. At that point, it causes an error if fastcore was called returning an output, such as

newModel = fastcore(model, core, epsilon, printlevel)

because the output (tissueModel in fastcore.m) hasn't been defined. I replaced it with an error, so the error message will propagate upwards. It allows running fastcore with try, catch, and understanding why fastcore failed.

I went over the COBRA toolbox and changed instances of using return to have either predefined outputs, or that they return an error. It is possible I made the wrong decisions, but I'd like to raise this issue.

addLoopLawConstraints.m has a comment, since I'm not sure what would make sense.

I ran the tests on my computer, and they worked. I'm going to see if the tests continue working on other OS/Matlab version combinations, and if not, I will attempt to update this branch.

### DIFF
--- a/src/analysis/sparseFBA/sparseFBA.m
+++ b/src/analysis/sparseFBA/sparseFBA.m
@@ -286,8 +286,7 @@ end
 
 %Double check that all inputs are valid:
 if ~(verifyCobraProblem(LPproblem, [], [], false) == 1)
-    warning('invalid problem');
-    return;
+    error('invalid problem');
 end
 
 if nnz(model.c)~=0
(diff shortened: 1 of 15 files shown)


tpfau:
Looks good to me.
Might have an effect on some existing code that relies on these functions to "not" error. But I would agree that a sensible fail is better than a requirement to check whether the outputs are valid, and a function that relies on this can easily be fixed by a try/catch block.

laurentheirendt:
Thanks @akaviaLab 👍

Issue #1130, opened 2018-03-14
removeGitIgnoredNewFiles seems to crash on my Mac


akaviaLab:
When running testAll, removeGitIgnoredNewFIles crashes
Error using cell/setdiff>cellsetdiffR2012a (line 291)
Input A of class cell and input B of class struct must be cell arrays of character
vectors, unless one is a character vector.

Error in cell/setdiff (line 83)
    [varargout{1:nlhs}] = cellsetdiffR2012a(varargin{:});

Error in removeGitIgnoredNewFiles (line 23)
diffContent = setdiff(newContent, oldcontent);

Error in testAll (line 241)
removeGitIgnoredNewFiles(testDirPath,testDirContent);

When trying to follow this error, it turns out that testAll defines oldContent as the output of rdir (which returns "a structure with the same fields as Matlab DIR output"). 
removeGitIgnoredNewFiles line 19 uses system('git ls-files') to get the files and the git command returns a char array.
At that point, trying to compare a struct and a char gives an error.

It seems that removeGItIgnoredNewFIles will only work if testDirContent is defined as a char, perhaps by using 'system('ls')' or if removeGitIgnoredNewFIles will call rdir. I'm not sure what would be preferable.

**I hereby confirm that I have:**

- [X] Tried to solve the issue on my own
- [X] Retried to run my code with the latest version of The COBRA Toolbox
- [x] Checked that a similar issue has not already been opened

*(Note: You may replace [] with [X] to check the box)*

rmtfleming:
Thanks a lot for this detailed investigation of an important issue.

tpfau:
Thanks for the report.
The issue was/is, that this fail happened after the tests were run. on the CI testAll returns before this cleanup is called. 
There currently seems to be some issue with our Windows CI machines, but we are working on it and this should be fixed in the develop branch soon.

tpfau:
Ok, bug on the Windows CI machines was caused due to some text being in a file which should not be there... My bad.

tpfau:
This should be fixed with #1138

laurentheirendt:
Thanks!

PR #1133, opened 2018-03-15
Test all fix

tpfau:
Addressing issue #1130

Simple exchange of rdir by git ls-files. It wasn't caught by the CI, since the removal of temporary files is never reached on the CI, where `testAll` exits before.

I moved the cleanup calls so that they are also executed on the CI.


tpfau:
Closing again... saw a mistake.

tpfau:
Reopening.
1. Speed up for checks for files to delete when on the cobra-folders (using git).
2. changeCobraSolver now does not fail if validation is requested and we are not on a cobra directory

src/base/install/getFilesInDir.m
```diff
@@ -0,0 +1,140 @@
...
+gitType = lower(parser.Results.gitTypeFlag);
+
+%test, whether the folder is git controlled
+[gitStatus,~] = system('git status');
+if gitStatus ~= 0 && ~strcmpi(gitType,'cobraignored')
+    gitType = 'all';
+end
+if gitStatus == 0 && strcmpi(gitType,'cobraignored')
+    [~,repos] = system('git remote -v');
+    if any(cellfun(@(x) ~isempty(strfind(x,'opencobra/cobratoolbox.git')),strsplit(repos,'\n')))
```
laurentheirendt: This condition must be `strfind(x, 'cobratoolbox.git')` for any forks.

src/base/install/getFilesInDir.m
```diff
@@ -0,0 +1,140 @@
...
+        end
+            
+    case 'tracked'
+        [status, files] = system('git ls-files');
+        files = strsplit(strtrim(files), '\n');
+    case 'untracked'
+        %Files, which are not tracked but are not ignored.
+        [status, files] = system('git ls-files -o -exclude-standard');
+        files = strsplit(strtrim(files), '\n');
+    case 'ignored'
```
laurentheirendt: What is the difference between `'ignored'` and `'cobraignored'`?

src/base/install/removeTempFiles.m
```diff
@@ -0,0 +1,49 @@
...
+%                                                (Default: true)
+%                          checkSubFolders -     Whether to check subFolder (Default: true)
+%
+
+parser = inputParser();
+parser.addParamValue('COBRAGitIgnoredOnly',true,@(x) islogical(x) || isnumeric(x) && (x == 0 || x == 1));
+parser.addParamValue('checkSubFolders',true,@(x) islogical(x) || isnumeric(x) && (x == 0 || x == 1));
+
+parser.parse(varargin{:});
+COBRAGitIgnoredOnly = parser.Results.COBRAGitIgnoredOnly;
```
laurentheirendt: May these variable names follow the `camelToe` format?

src/base/install/getFilesInDir.m
tpfau: Cobraignored will read the COBRA ignored files file, while just ignored relies on the .gitignore of whichever project this is in, (and if the folder is not under git control, it doesn't work).

tpfau:
Ok. no idea what went wrong with codecov here...

src/base/install/getFilesInDir.m
laurentheirendt: Still not sure I understand ... Isn't this the same for this repo? Or is there a plan to use this function in another repo?

tpfau: Yes and no...
If you run this function in e.g. tutorials, he tutorials .gitignore would be used if you run it with `'ignore'`. 
Also `'ignore'`, is much fast than `'cobraignore'`, since the git functions are much faster than manually parsing the gitignore file and regexp checking each file...
However 'ignore' only works if you are in a git controlled folder AND it will use whichever .gitignore is valid for that folder.
So if you call e.g. changeCobraSolver from some non cobra-repository location, it would use the "wrong" or no .gitignore file. We could (of course) add cd's to changeCobraSolver, but I'm not sure, whether we should do that...
Thats why I implemented the two options. For `changecobraSolver` we need the explicit cobraIgnore option. for initCobraToolbox and testAll we can work with the faster git option (as we cd).

laurentheirendt: May you add such a note to the header?

tpfau:
I'll retrigger the CI.

tpfau:
reopening, after forced update.

PR #1145, opened 2018-03-23
Fix pip

Fede-edef:
*Some changes implemented as suggested by @tpfau *

### DIFF
--- a/src/analysis/multiSpecies/microbiomeModelingToolbox/mgPipe/README.md
+++ b/src/analysis/multiSpecies/microbiomeModelingToolbox/mgPipe/README.md
@@ -3,9 +3,9 @@
 ## Introduction
 
 MgPipe is a MATLAB based pipeline to integrate microbial abundances (coming
-from metagenomic data) with constraint based modeling, creating
+from metagenomic data) with constraint-based modeling, creating
 individuals' personalized models.  The pipeline is divided in 3 parts: [PART 1]
-Analysis of individuals' specific microbes abundances is computed. Individuals?
+Analysis of individuals' specific microbes abundances is computed. Individuals'
 metabolic diversity in relation to microbiota size and disease presence as well
 as Classical multidimensional scaling (PCoA) on individuals' reaction
 repertoire are examples.  [PART 2]: 1 Constructing a global metabolic model
@@ -16,13 +16,13 @@ function.  [PART 3] Simulations under different diet regimes. Set of standard
 analysis to apply to the personalized models. PCA of computed MNPCs of
 individuals as for example.
 
-**WARNING :**Please take into consideration only the files listed in this
+**WARNING:** Please take into consideration only the files listed in this
 document. Everything present in the folder but not listed and explained in this
 document is to be considered not relevant or obsolete.
 
 ## Requirements
 
-MgPipe requires Matrix Laboratory as well as the Cobra Toolbox to be installed
+MgPipe requires Matrix Laboratory, the Parallel Computing Toolbox, as well as, the Cobra Toolbox to be installed
 with ILOG CPLEX as solver.  All toolboxes/solvers/files should be in the MATLAB
 path.  MgPipe was created (and tested) for AGORA 1.0 please first download
 AGORA version 1.0 from https://vmh.uni.lu/#downloadview and place the mat files
@@ -30,7 +30,7 @@ into a dedicated folder.
 
 ## Main Folder Structure and Files
 
-The following files are supplied
+The following files are essential for the usage of the pipeline and are supplied in the current folder or at `COBRA.papers/2018_microbiomeModelingToolbox`
 
 | Filename                                       | Purpose                                                                |
 | -----------------------------------------------|------------------------------------------------------------------------|
@@ -42,7 +42,7 @@ The following files are supplied
 | detectOutput.m                                 | *function to check if a specific file was already created and saved*   |
 | getIndividualSizeName.m                        | *get information on number and ID of organisms and individuals*|
 | addMicrobeCommunityBiomass.m                   | *function to add community biomass*                                    |
-| MgPipe.m                                       | *Pipeline*                                                             |
+| MgPipe.m                                       | *pipeline*                                                             |
 | parsave.m                                      | *function to allow object saving in parallel loops*                    |
 | getMappingInfo.m                               | *function to extract information from the mapping*             |
 | plotMappingInfo.m                              | *function plot extracted information from the mapping*                 |
@@ -51,11 +51,10 @@ The following files are supplied
 | makeDummyModel.m                               | *function to create a dummy model*                                     |
 | mgSimResCollect.m                              | *function to collect and output simulation results*                    |
 | README.md                                      | *this file*                                                            |
-| **DietImplementation**                         | *Folder containing diets and related scripts*                          |
 | useDiet.m                                      | *function to impose a specific diet and add essential elements to microbiota models*|
 | adaptVMHDietToAGORA.m                          | *function to convert a specific diet from VMH into an AGORA compatible one*|
 | **examples**                                   | *folder containing necessary files to replicate the pipeline tutorial* |
-| StartMgPipeHMP.m                               | *driver, containing all the input variables of the HMP study*         |
+| StartMgPipe.m                                  | *driver, containing all the input variables to start MgPipe*         |
 | ***compfile***                                 | *Results subfolder: contains objects saved in open format*             |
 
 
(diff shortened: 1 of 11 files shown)


src/analysis/multiSpecies/microbiomeModelingToolbox/mgPipe/README.md
```diff
@@ -16,33 +16,36 @@ function.  [PART 3] Simulations under different diet regimes. Set of standard
...
 ## Requirements
 
-MgPipe requires Matrix Laboratory as well as the Cobra Toolbox to be installed
-with ILOG CPLEX as solver.  All toolboxes/solvers/files should be in the MATLAB
-path.  MgPipe was created (and tested) for AGORA 1.0 please first download
-AGORA version 1.0 from https://vmh.uni.lu/#downloadview and place the mat files
-into a dedicated folder.
+MgPipe requires `Matrix Laboratory`, the `Parallel Computing Toolbox`, as well
+as, the Toolbox to be installed with the `ILOG CPLEX` solver. Pleas refer to
+the [installation instructions](https://opencobra.github.io/cobratoolbox/stable/installation.html).
```
tpfau: Just as a question:
Apart from reproducibility or numeric stability:
Why do you need CPLEX? Shouldn't any other solver do the same job?

src/analysis/multiSpecies/microbiomeModelingToolbox/mgPipe/microbiotaModelSimulator.m
```diff
@@ -77,7 +77,7 @@
 % Starting personalized simulations
 for k = startIter:(patNumb + 1)
  idInfo = cell2mat(sampName((k - 1), 1))
- load(strcat('microbiota_model_samp_', idInfo, '.mat'))
+ load(strcat('microbiota_model_samp_', idInfo, '.mat'));
```
tpfau: redCbModel ? Or is there a specific reason for load?

src/analysis/multiSpecies/microbiomeModelingToolbox/mgPipe/README.md
Fede-edef: Hi Thomas,
The reason for asking for CPLEX is to be able to use fastFVA function which, at the moment, is not compatible with other solvers. As matter of fact, it is possible to run the simulations also with the fluxVariability function and this works with every solver, but it is much slower....

src/analysis/multiSpecies/microbiomeModelingToolbox/mgPipe/microbiotaModelSimulator.m
Fede-edef: Hi Thomas,
I tried to use readCbModel but I got errors. I think it is because the structure of the microbiota models is a bit different from the usual...  (additional fields for example )

tpfau: Additional fields should not be an issue.
The problem is that this is then likely to fail on anything but exactly those models, and can lead to problems in the future, if things change for whatever reason. 
I'll have a look at this and come back to you.

Fede-edef: Thanks for your help Thomas :)

src/analysis/multiSpecies/microbiomeModelingToolbox/mgPipe/README.md
tpfau: Then I would suggest to indicate that, and only say that it is suggested to use CPLEX due to fastFVA not being available for other solvers and thus slower. Even though from some quick tests, on recon2, they are not really far apart speedwise (which might be due to a lot of reactions in recon 2 having their ub as max and lb as min flux, which speeds up fluxVariability a lot).

src/analysis/multiSpecies/microbiomeModelingToolbox/mgPipe/microbiotaModelSimulator.m
tpfau: Do you have some script with supplemental  files that I could use to test this?
Some kind of test that can be run on the toolbox (potentially with a smaller subset of models) that reproduces your issues with readCbModel ?
We could base a test of mgPipe on that test once we have sorted out the problems.

Fede-edef: Hi Thomas,
https://github.com/opencobra/COBRA.tutorials/tree/develop/additionalTutorials/microbiomeModelingToolbox
Here, in the tutorial,  you have all the inputs (except then the agora models to be downloaded from vmh.life)  and the steps to reproduce the all Pipeline for a really small example (it should take around 10 minutes).
Otherwise, if you prefer to have directly the models let me know I can try to send them to you....
thanks
fede

tpfau: running mgPipeTutorial leads to errors:
1. solver is not defined in the driver (i set it by `solver = 'ibm_cplex'`)
2. the initMgPipe call does not work:
```
Error using initMgPipe
Too many input arguments.

Error in mgPipeTutorial (line 93)
[init,modPath,toolboxPath,resPath,dietFilePath,abunFilePath,objre,figForm,solver,numWorkers,autoFix,compMod,patStat,rDiet,extSolve,fvaType,autorun]= initMgPipe(modPath, toolboxPath, resPath, dietFilePath, abunFilePath, objre, figForm, solver, numWorkers, autoFix, compMod, patStat, rDiet,extSolve,fvaType,autorun);
```
And admittedly: This call is way to long (isn't there any way to provide defaults of parameter value pairs)? its hard to say what is "wrong".

Fede-edef: Hi Thomas,
Yes, you are right. The 'solver' input got deprecated as it is set automatically by initCobra toolbox. 
I guess the problem with the tutorial is that am guilty of having forgotten to remove the deprecated 'solver' parameter from the call to the initMgPipe function. I will fix this. If you still want to try to run it you can easily fix it by removing solver from the call of the function. Everything else should work. Sorry for the inconvenience. 

Concerning the number of parameters, I am not sure...

tpfau: Ok...
This is very likely an issue with coupling constraints.

src/analysis/multiSpecies/microbiomeModelingToolbox/mgPipe/fastSetupCreator.m
```diff
@@ -38,18 +38,19 @@
...
 for j = 1:size(models, 1)
-    model = models{j, 1};
-    model = findSExRxnInd(model);
-    mod_exch=model.rxns(model.ExchRxnBool);
-    mod_exch=strrep(mod_exch, 'EX_', '');
-    mod_exch=strrep(mod_exch, '(e)', '[e]');
-    exch = union(exch,mod_exch);
+     model = models{j, 1};
+     exch = union(exch, model.mets(find(sum(model.S(:, strncmp('EX_', model.rxns, 3)), 2) ~= 0)));
+%      model = findSExRxnInd(model);
```
tpfau: I see what went wrong. you essentially got the biomass reaction in as well and added reactions instead of metabolites.
This should do what you want:
```
if ~isempty(host)  % Get list of all exchanged metabolites
    exStruct = findSExRxnInd(host);
    exch = findMetsFromRxns(host,host.rxns(exStruct.ExchRxnBool & ~exStruct.biomassBool));
else
    exch = {};
end
for j = 1:size(models, 1)
    model = models{j, 1};
    exStruct = findSExRxnInd(model);
    new_exch = findMetsFromRxns(model,model.rxns(exStruct.ExchRxnBool & ~exStruct.biomassBool));
    exch = union(exch,new_exch);
end
```

tpfau:
As for the issues with `readCbModel`:
One cause is likely to the `makeDummyModel` function.
Is there a reason why you don't use `createModel()` along with `addReaction/addMetabolite` ?
If this is too slow, we just added  `addMultipleReactions` and a `addMultipleMetabolites` which are much fast and intended exactly for use in these kinds of circumstances.

The MakeDummeModel actually creates invalid fields, and fields that were changed quite a bit (e.g. its metKEGGID and rxnKEGGID etc. 
Using createModel + add functions would ensure that you create a valid model.

src/analysis/multiSpecies/microbiomeModelingToolbox/mgPipe/fastSetupCreator.m
Fede-edef: Yes, the problem seemed to be that, enabling the function back and changing the call as you proposed worked....
Thanks

Fede-edef: @laurentheirendt

laurentheirendt:
@Fede-edef, now, only tests are missing. The issue with `makeDummyModel` pointed out by @tpfau can be solved in a next PR.

Thanks 👍

PR #1147, opened 2018-03-26
Return of duals and reduced costs from solveCobraQP

laurentheirendt:
- return of duals and reduced costs from solveCobraQP for `ibm_cplex` and `tomlab_cplex` in `solveCobraQP` 
- additional test to make sure that the signs of the duals and reduced costs are the same when returned from `solveCobraLP` and `solveCobraQP`.

For `mosek`:
- fixed sign of objective value
- tolerance for optimality condition in `solveCobraQP` temporarily set to `1e-2`

cc: @rmtfleming 


rmtfleming:
:)

PR #1151, opened 2018-03-29
Fix TheoretMaxProd Test

tpfau:
The TheoretMaxProd test was solver dependent and, since the solutions were not unique could fail if a solver did not yield the expected solution.
This was corrected in the new test and the function documentation was updated to better explain the options.


laurentheirendt:
Thanks @tpfau 👍

PR #1152, opened 2018-03-29
Fixed poropagation of arguments into findElementaryMoietyVectors

tpfau:
`computeMetFormula` passed on the whole `varargin` array which might not be a valid struct for `findElementaryMoietyVectors`.
This PR removes the first element from the varargin vector if it is a struct before passing on additional arguments.


laurentheirendt:
Thanks @tpfau 👍

PR #1158, opened 2018-04-05
Parse sbml notes improvements

akaviaLab:
I added the ability to parse case-insensitive notes, specifically for Confidence Level (Is not all caps in RECON 2.2). I also did some general tidying. I tried separating into multiple commits, since if there is a specific reason to run regexp twice, it is a separate commit and can be ignored.

### DIFF
--- a/src/base/io/utilities/parseSBMLNotesField.m
+++ b/src/base/io/utilities/parseSBMLNotesField.m
@@ -46,11 +46,10 @@ else
 end
 
 
-[tmp,fieldList] = regexp(notesField,['<' tag '>.*?</' tag '>'],'tokens','match');
+[fieldTmp,fieldList] = regexp(notesField,['<' tag '>(.*?)</' tag '>'],'tokens','match');
 
 for i = 1:length(fieldList)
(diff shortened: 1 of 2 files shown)


src/base/io/utilities/parseSBMLNotesField.m
```diff
@@ -46,44 +46,41 @@
...
-        formula = strtrim(strjoin(strfields(2:end),':'));
-    elseif strcmp(strfields{1},'CHARGE') || strcmp(strfields{1},'Charge')
-        charge = str2num(strtrim(strjoin(strfields(2:end),':')));
+    elseif strcmpi(strfields{1},'EC Number') || strcmpi(strfields{1},'EC_Number')
+        ecNumber = valueStr;
+    elseif strcmpi(strfields{1},'FORMULA')
+        formula = valueStr;
+    elseif strcmpi(strfields{1},'CHARGE')
+        charge = str2num(valueStr);
     elseif strcmp(strfields{1},'AUTHORS')
```
tpfau: Would it make sense to change this to `strcmpi` as well?

laurentheirendt:
Thanks @akaviaLab 👍

PR #1162, opened 2018-04-11
Cross-platform support for CPLEX 12.8

laurentheirendt:


tpfau:
the `isCompatible` test is flawed at the moment:
if the solver can be changed, this indicates, that `isCompatible` should return either `1` (compatible statement) or `2` (undefined). at the moment, the test asserts `compatibleStatuts == solverOK`. 
This should be changed to `compatibleStatuts && solverOK`

PR #1167, opened 2018-04-16
Add an option to exclude Solvers for Tests

tpfau:
Some tests work with all but a few solvers, specifying solvers to exclude is therefore sometimes easier than specifying which solvers are possible. 
Added a new option to `prepareTest` to reflect that.


laurentheirendt:
Thanks @tpfau 👍

PR #1173, opened 2018-04-19
Checking requirements

tpfau:
Skip the test, if the Parallel toolbox is not installed.
Currently, the test indicates a fail if no parallel toolbox is installed. now it will be skipped due to missing requirements.


laurentheirendt:
Nice spotted! 👍

PR #1176, opened 2018-04-19
Fix blocked reactions

marouenbg:
cc @Fede-edef 

### DIFF
--- a/src/analysis/exploration/findBlockedReaction.m
+++ b/src/analysis/exploration/findBlockedReaction.m
@@ -22,6 +22,7 @@ function blockedReactions = findBlockedReaction(model, method)
 
 blockedReactions = cellstr('');
 qptol = getCobraSolverParams('QP','intTol');
+[m,n]=size(model.S);
 if (nargin < 2 || isequal(method, 'FVA'))
     tol = 1e-10;
     [minMax(:, 1), minMax(:, 2)] = fluxVariability(model, 0);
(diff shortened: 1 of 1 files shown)


src/analysis/exploration/findBlockedReaction.m
```diff
@@ -33,7 +33,7 @@
     end
 else
     solution = solveCobraLPCPLEX(model, 0, 0, 0, [], 1e-6);
-    blockedReactions = model.rxns(find(~solution.full))';
+    blockedReactions = model.rxns(find(abs(solution.full) < 1e-11))';
```
tpfau: shouldn't this be smaller than the feasibility tolerance, whatever that is? 
Also, you don't need the find any more if you have a logical array anyways.

marouenbg: the test fails using the feasTol, and 1e-11 is more stringent in any case.

src/analysis/exploration/findBlockedReaction.m
```diff
@@ -33,7 +33,7 @@
     end
 else
     solution = solveCobraLPCPLEX(model, 0, 0, 0, [], 1e-6);
-    blockedReactions = model.rxns(find(abs(solution.full) < 1e-11))';
+    blockedReactions = model.rxns(abs(solution.full) < 1e-11)';
```
tpfau: I would still argue, that your "null criterion" should depend on the requested feasibility Tolerance used. 
if you have a feasibility tolerance of only 1e-4 than a reaction carrying a flux of 1e-5 can still be blocked. 
Or, since, it is also a fixed number above, at least the same tolerance (1e-10) should be used.

marouenbg:
Thanks @tpfau for the tip.

PR #1177, opened 2018-04-19
Update to restore Environment

tpfau:
An update to allow specific resetting of the globals only (which is much faster than also resetting the path).
Should speed up testFVA and particularily testpFBA.


laurentheirendt:
Thanks @tpfau 👍

PR #1182, opened 2018-04-23
Update BiGG Test to skip if no connection to BiGG possible

tpfau:
Skipping the BiGG test, if we cannot read from the BiGG database, i.e. if webread fails. 


tpfau:
Well. Under normal circumstances the 5 second timeout should be fine... and if it is not reachable any higher timeout only slows things down, so I prefer sticking with 5 seconds.

Wrt to the rebase: I don't think its necessary, as the failing test is quite obviously not related to your code.

laurentheirendt:
Thanks @tpfau 👍

PR #1184, opened 2018-04-23
Q not positive semidefinite for maximisation problems.

huldash:

### DIFF
--- a/src/base/solvers/solveCobraQP.m
+++ b/src/base/solvers/solveCobraQP.m
@@ -573,7 +573,7 @@ switch solver
             osense = 1;
        end
 
-        QPproblem.Q = 0.5*sparse(QPproblem.F);
+        QPproblem.Q = osense*0.5*sparse(QPproblem.F);
         QPproblem.modelsense = QPproblem.osense;
         [QPproblem.A,QPproblem.rhs,QPproblem.obj,QPproblem.sense] = deal(sparse(QPproblem.A),QPproblem.b,double(QPproblem.c),QPproblem.csense);
         resultgurobi = gurobi(QPproblem,params);


src/base/solvers/solveCobraQP.m
```diff
@@ -573,7 +573,7 @@
             osense = 1;
        end
 
-        QPproblem.Q = 0.5*sparse(QPproblem.F);
+        QPproblem.Q = osense*0.5*sparse(QPproblem.F);
```
tpfau: I'm not sure why solveCobraQP should change the objective. 
The respective code has to ensure, that the F matrix is either positive semidefinite or negative semidefinite depending on the objective. 
This change would break a lot of code that considers this and formulates its F matrix accordingly. 
In addition, changing this solely on gurobi would make gurobi return results different to any other solver, but as I said, I would not change this at all, as it is changing the objective. I.e. it is impossible to maximise the sum of quadratic fluxes. But with this change, it would actually minimise them (valid), which is not what was intended.

tpfau:
Hmmm.
Ok, this is really odd and very inconsistent in between solvers. 
I also get the impression, that the optimality criterion is not yet entirely correct. 
Overall I think any adaption needs much more than just one addition of a `osense` in order to keep this consistent in between solvers. And I personally would suggest to:  
A: Allow min/max as objective as indicated in parts of the documentation (instead of enforcing min for F)
B: Add a maximisation example to the QP test.

rmtfleming:
I think only osense*c*v + 0.5*v'*F*v is valid, when F is positive semidefinite. If F is negative semidefinite, one could make a case for osense*(c*v + 0.5*v'*F*v), only if osense is +1, but this would be unusual.

tpfau:
`osense*c + 0.5 x'*F*x` looks odd to me. 
`osense` should be indicative of overall maximization/minimization, and not just apply to some part of the objective.
As such, I would argue, that if you want to use `osense == -1`, your `F` has to be negative semidefinite, and if `osense == 1` it has to be positive semidefinite. But applying `osense` only to parts of the objective is, at least to me, very counter intuitive.

huldash:
Hi Thomas, good points! I will close the pull request.

tpfau:
I think its good you pointed this out. As we will have to address this somehow.
Could you raise an issue referencing this PR?

PR #1185, opened 2018-04-24
Adaption of FVA to unbounded systems

tpfau:
FVA could fail if a system had unbounded reactions. 
This PR fixes this issue and also adds a graceful termination, if solvers report that a solution is not optimal in FVA (this should normally not happen but it can happen if solvers cannot properly handle unboundedness).
Finally gurobi now returns the correct status if the original status is infeasible or unbounded by checking, if a system without objective can achieve an optimal flux (at which point the original solution is unbounded, as feasibility is not influenced by the objective)


laurentheirendt:
Thanks @tpfau 👍

PR #1186, opened 2018-04-24
Correction for verifyFunctionError

tpfau:
`verifyCobraFunctionError` always returned true if inputs were defined, as those were not present in the verification function. This did not show up, as all errors should give errors and no message tests were performed.
Updated the functionality and corrected all concerned tests.


laurentheirendt:
Thanks @tpfau 👍

PR #1204, opened 2018-05-16
Fixing bug with removal of reaction prior consistency check

arichelle:
I just saw this mistake in the MBA code.

Modification of the MBA code to ensure that the random reaction selected is removed only if it the system is still solvable after removal and therefore ensure that the consistency check of the model is performed on a "solvable" model. Indeed, in the previous version the random reaction selected was removed in anyway while it could lead to inconsistent model.


syarra:
Thanks @arichelle!

PR #1210, opened 2018-05-20
Ftest

Fede-edef:
*Adding tests for mgPipe*

- [x ] Tested my code on my own machine

### DIFF
--- a/test/verifiedTests/analysis/testMultiSpeciesModelling/testMgPipe.m
+++ b/test/verifiedTests/analysis/testMultiSpeciesModelling/testMgPipe.m
@@ -167,35 +167,33 @@ setup=fastSetupCreator(models, organisms, {},objre);
  
 assert(exist('setup', 'var') == 1)
 assert(strcmp(class(setup),'struct') == 1)
-assert((length(setup.S) == length(setup.rxns)) == 1)
-assert((length(setup.lb) == length(setup.rxns)) == 1)  
-assert((length(setup.c) == length(setup.rxns)) == 1) 
-assert((length(setup.mets) == length(setup.metNames)) == 1)  
+assert(size(setup.S,2) == length(setup.rxns))
+assert(length(setup.lb) == length(setup.rxns))
+assert(length(setup.c) == length(setup.rxns))
+assert(length(setup.mets) == length(setup.metNames)) 
  
-assert(length(setup.rxns(strmatch('EX',setup.rxns))) /2 == length(setup.rxns(strmatch('DUt',setup.rxns))) == 1) 
-assert(length(setup.rxns(strmatch('EX',setup.rxns))) /2 == length(setup.rxns(strmatch('UFEt',setup.rxns))) == 1) 
-assert(length(setup.rxns(strmatch('DUt',setup.rxns)))== length(setup.rxns(strmatch('UFEt',setup.rxns))) == 1)
+assert(length(setup.rxns(strmatch('EX',setup.rxns))) /2 == length(setup.rxns(strmatch('DUt',setup.rxns)))) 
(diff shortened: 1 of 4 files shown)


test/verifiedTests/analysis/testMultiSpeciesModelling/testMgPipe.m
```diff
@@ -134,3 +139,78 @@
...
+
+% test createdModels
+[createdModels]=createPersonalizedModel(abunFilePath,resPath,setup,sampName,organisms,indNumb);
+assert(length(createdModels)== 5)
+assert(strcmp(createdModels(2,1),'Test1') == 1)
+assert(strcmp(createdModels(5,1),'Test4') == 1)
+
+microbiota_model=load(strcat(resPath,'microbiota_model_samp_Test1')); 
+microbiota_model=microbiota_model.microbiota_model;
+assert(length(microbiota_model.A)> length(microbiota_model.S))
```
tpfau: I would recommend not testing this (A/csense/S), as it is likely to become invalid in the future. 
In particular, as to the model definitions, length(csense) should be size(S,1). 

Also I would suggest to use an explicit size(S/A,1/2) to refer to the respective dimensions, as length can be either dimension and if someone else tries to figure out what a certain assert in a test means, this makes it really hard to detect potential problems.

test/verifiedTests/analysis/testMultiSpeciesModelling/testMgPipe.m
```diff
@@ -134,3 +139,78 @@
...
+assert(length(microbiota_model.csense) == length(microbiota_model.A))
+assert((length(strmatch('communityBiomass',microbiota_model.rxns)) > 0) == 1)
+assert((length(strmatch('UFEt_microbeBiomass',microbiota_model.rxns)) > 0) == 1)
+assert((length(strmatch('EX_microbeBiomass[fe]',microbiota_model.rxns)) > 0) == 1)
```
tpfau: May I suggest:
`any(ismember('EX_microbeBiomass[fe]',microbiota_model.rxns))` ?
It is equivalent, but from local tests, `any`+`ismember` is faster and (at least personally) I find it easier to understand what is tested.

test/verifiedTests/analysis/testMultiSpeciesModelling/testMgPipe.m
```diff
@@ -134,3 +139,78 @@
...
+assert(length(microbiota_model.rxns(strmatch('DUt',microbiota_model.rxns)))== length(microbiota_model.rxns(strmatch('UFEt',microbiota_model.rxns)))-1 == 1)
```
tpfau: This assertion is odd (imo). 
It tests that the following is true:
`'DUt'` must be in `microbiota_model.rxns` and `'UFEt'` must be in `microbiota_model.rxns`
But the way this is tested is really odd...

test/verifiedTests/analysis/testMultiSpeciesModelling/testMgPipe.m
```diff
@@ -134,3 +139,78 @@
...
+assert(exist('fixVec', 'var') == 1)
+assert(length(organisms) == 5)
+
+% test fastSetupCreator
+
+setup=fastSetupCreator(models, organisms, {},objre);
+ 
+assert(exist('setup', 'var') == 1)
+assert(strcmp(class(setup),'struct') == 1)
+assert((length(setup.S) == length(setup.rxns)) == 1)
```
tpfau: As further down:
could you please use `size()` for matrices?
Also:
Could `verifyCobraModel` make sense here?

laurentheirendt:
Thanks @Fede-edef 👍

PR #1221, opened 2018-05-30
Speedup for Pairwise Model generation

tpfau:
This PR speeds up the generation of pairwise models by adding all reactions to the models simultaneously and removing the printout from the calls.


laurentheirendt:
Thanks @tpfau 👍

PR #1222, opened 2018-05-31
Fixing issues on addMultipleReactions when the model is empty

tpfau:
When the model was empty, `addMultipleReactions` would fail if fields referencing rxns where added as optional arguments.


tpfau:
Need to check some things here.

tpfau:
Reopening after fixing issues

laurentheirendt:
Thanks @tpfau 👍

PR #1239, opened 2018-06-07
Reaction addition improvements

tpfau:
- This PR addresses a bug in `addMultipleReactions` where trying to set GPR rules resulted in an invalid model. 
- It also introduces a `printLevel` parameter to the `addMultipleReactions` functions that allows this function to also print all reactions added, if requested. Also added testing code for it.
- Further updated the `addDemandRxn` function to use `addMultipleReactions` speeding up the function.
- Finally, `addExchangeRxn` is corrected to allow the addition of exchangers for non existing metabolites (adding the metabolite), which was impossible after the change to addMultipleReactions.


syarra:
Thanks @tpfau for these improvements.

Issue #1247, opened 2018-06-08
SBO terms get redundantly stored as notes


BenjaSanchez:
After defining SBO terms for my model and saving it with COBRA, they are being stored 2 times: once in the proper field `sboTerm` but also as an additional note. For instance, for metabolites ideally it should be:
```html
      <species metaid="s_0001__91__ce__93__" sboTerm="SBO:0000247" id=...
```
but instead I get:
```html
      <species metaid="s_0001__91__ce__93__" sboTerm="SBO:0000247" id=...
        <notes>
          <body xmlns="http://www.w3.org/1999/xhtml">
            <p>metSBOTerms:SBO:0000247</p>
          </body>
        </notes>
```
Which is redundant but also conflicting, as it gets appended to `model.metNotes` when reading back and in the next I/O rounds it keeps on growing:
```html
      <species metaid="s_0001__91__ce__93__" sboTerm="SBO:0000247" id=...
        <notes>
          <body xmlns="http://www.w3.org/1999/xhtml">
            <p>metSBOTerms:SBO:0000247</p>
            <p>metSBOTerms:SBO:0000247</p>
          </body>
        </notes>
```
Same happens with `rxnSBOTerms`. I tried with and without the leading `SBO:` and the results are the same.

**Reproducing these results:**

Using [yeast-GEM](https://github.com/SysBioChalmers/yeast-GEM):
```matlab
model.metSBOTerms    = cell(size(model.mets));
model.metSBOTerms(:) = {'SBO:0000247'};
writeCbModel(model,'sbml','yeastGEM.xml');
```

**I hereby confirm that I have:**

- [X] Tried to solve the issue on my own
- [X] Retried to run my code with the latest version of The COBRA Toolbox
- [X] Checked that a similar issue has not already been opened

tpfau:
Thanks for raising the issue.
This was due to a wrong regexp for SBO Term IDs in the fieldDefinitions. 
However, this information wil still be stored in a redundant way (both as SBOTerm and as a `hasProperty` annotation), as  we can't be sure how it got into the model, and other tools might or might not read the SBOTerm and/or annotation fields. 
However it should no longer start to multiply.

BenjaSanchez:
Hi, thanks for the quick reply! Looks like `rxnSBOTerms` has the same problem in `fieldDefinitions`, as that regexp also lacks the `:` char, so you might want to fix that as well.
No problem if there is a little redundancy for the reason you point out :)

[...]

PR #1248, opened 2018-06-11
fixing SBOTerm property definition

tpfau:
This PR fixes the issue reported here:
 #1247

The SBO Term regexp missed a `:` and it had the wrong bioql relation. An SBO identifier is a property of an element (i.e. `hasProperty`) not a `is` relation. 
This information will be duplicated in the output SBML file, as different tools might ignore either the annotation or the SBOterm attribute. 


laurentheirendt:
Thanks @tpfau 👍

PR #1250, opened 2018-06-14
Fixing bug with caps mismatch with printLevel

akaviaLab:
The print level was defined as printLevel in most of the code, but fastcore (for example) would check for the existence of printLevel, and then set printlevel, which resulted in this variable having no effect.


laurentheirendt:
Thanks @akaviaLab 👍

PR #1260, opened 2018-06-20
Reading Model Annotation from SBML files

tpfau:
Update to the SBML IO to properly read and write the model annotations.


laurentheirendt:
Is there any way that we can add a few tests for these additions?

tpfau:
Yes,
I'll have a look.

tpfau:
I'm working on that now. But to make this a sensible check I will have to build a few functions for annotations (which can be reused when writing an SBML).

tpfau:
Closing temporarily

PR #1249, opened 2018-06-12
Adding an option to check for coupled metabolites

tpfau:
Adding an option to `biomassPrecursorCheck` to check for coupled metabolites as suggested in #1238 

**I hereby confirm that I have:**

- [X] Tested my code on my own machine
- [X] Followed the guidelines in the [Contributing Guide](https://opencobra.github.io/cobratoolbox/docs/contributing.html)
- [X] Selected `develop` as a target branch (top left drop-down menu)

*(Note: You may replace [ ] with [X] to check the box)*

PR #1238, opened 2018-06-06
Fix biomassPrecursorCheck

shjchan:
*Please include a short description of enhancement here*
- When no missing metabolite is found, the function output `missingMets` is not initialized, leading to an error. The same for `presentMets`. Fixed this.
- Correct the function name in the documentation
- Add `testBiomassPrecursorCheck.m` for `biomassPrecursorCheck.m`
(This function has not been updated for quite long time. Is there another function for the same purpose? I have kept using this.)

**I hereby confirm that I have:**

- [X] Tested my code on my own machine
- [X] Followed the guidelines in the [Contributing Guide](https://opencobra.github.io/cobratoolbox/docs/contributing.html)
- [X] Selected `develop` as a target branch (top left drop-down menu)

*(Note: You may replace [ ] with [X] to check the box)*

PR #1264, opened 2018-06-21
Check cofactor pairs in biomassPrecursorCheck

shjchan:
This PR continues the efforts in #1238 and #1249 to improve `biomassPrecursorCheck`.
In some models some cofactors are consumed during biomass production (e.g. AA-charged-tRNA -> unchaged-tRNA, fatty-acyl-ACP -> ACP). The AA-charged tRNA and the uncharged tRNA must be consumed and produced as a pair because the uncharged tRNA is not synthesized in the network (conserved moiety defined by the network structure). The current `biomassPrecursorCheck` would identify these metabolites as unable to be synthesized. This PR adds the functionality to check these cofactor pairs by looking at the conserved moieties defined by the network structure.

### DIFF
--- a/docs/prepareTutorials.sh
+++ b/docs/prepareTutorials.sh
@@ -96,7 +96,7 @@ createLocalVariables(){
     htmlHyperlink="https://prince.lcsb.uni.lu/cobratoolbox/tutorials/$tutorialFolder/iframe_$tutorialName.html"
     mlxHyperlink="https://github.com/opencobra/COBRA.tutorials/raw/master/$tutorialFolder/$tutorialName.mlx"
     mHyperlink="https://github.com/opencobra/COBRA.tutorials/raw/master/$tutorialFolder/$tutorialName.m"
-    dirHyperLink="https://github.com/opencobra/COBRA.tutorials/tree/master/$tutorialFolder"
+    dirHyperlink="https://github.com/opencobra/COBRA.tutorials/tree/master/$tutorialFolder"
 
     previousSection=""
     if [[ -n $section ]]; then
--- a/src/analysis/exploration/biomassPrecursorCheck.m
+++ b/src/analysis/exploration/biomassPrecursorCheck.m
@@ -1,7 +1,7 @@
-function [missingMets, presentMets,coupledMets] = biomassPrecursorCheck(model,checkCoupling)
+function [missingMets, presentMets,coupledMets, missingCofs, presentCofs] = biomassPrecursorCheck(model,checkCoupling)
 % Checks if biomass precursors are able to be synthesized.
 %
-% [missingMets, presentMets, coupledMets] = biomassPrecursorCheck(model)
+% [missingMets, presentMets, coupledMets] = biomassPrecursorCheck(model, checkCoupling)
 %
 % INPUT:
 %    model:             COBRA model structure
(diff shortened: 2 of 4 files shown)


src/analysis/exploration/biomassPrecursorCheck.m
```diff
@@ -16,37 +16,39 @@
 %    presentMets:    List of biomass precursors that are able to be synthesized
 %    coupledMets:    List of metabolites which need an exchange reaction for at least one other
 %                    biomass component because their production is coupled to it.
-%                    
+%    missingCofs:    List of cofactor pairs (defined by the network conserved moieties) 
+%                    that are not able to be synthesized
+%    presentCofs:    List of cofactor pairs that are able to be synthesized
 %
 % .. Authors: - Pep Charusanti & Richard Que (July 2010)
-% May identify metabolites that are typically recycled within the
```
tpfau: Could you add this under a NOTE?
i.e. 
```
% NOTE:
%    May identify metabolites that are typically recycled within the
%    such as ATP, NAD, NADPH, ACCOA. Turn on checkCoupling to check them.
```

src/analysis/exploration/biomassPrecursorCheck.m
```diff
@@ -1,7 +1,7 @@
-function [missingMets, presentMets,coupledMets] = biomassPrecursorCheck(model,checkCoupling)
+function [missingMets, presentMets,coupledMets, missingCofs, presentCofs] = biomassPrecursorCheck(model,checkCoupling)
```
tpfau: Given that #1249 change is very recent, I think we can savely change the header here. 
And, given that this is another option, I personally would suggest parameters for
`checkForCoupling`  and `checkForConservedQuantities`
which by default are false but can be turned on.
And probably change the outputs to `argout`, explaining the order of the outputs in the documentation.
Whyt do you think?

shjchan: @tpfau I think for general users when they call the function, they would like everything to be checked, so it seems more convenient to me to activate all additional functionalities using only one flag. When you added the `checkCoupling` flag, I guess it was more for backward compatibility, i.e., to keep the same function outputs when it is not turned on, correct?

shjchan:
@tpfau Is that what you were thinking? I have also deleted the test using the iJO1366 model, which is not quite necessary now and a bit time consuming.

tpfau:
Yep looks good.

docs/prepareTutorials.sh
```diff
@@ -96,7 +96,7 @@ createLocalVariables(){
     htmlHyperlink="https://prince.lcsb.uni.lu/cobratoolbox/tutorials/$tutorialFolder/iframe_$tutorialName.html"
     mlxHyperlink="https://github.com/opencobra/COBRA.tutorials/raw/master/$tutorialFolder/$tutorialName.mlx"
     mHyperlink="https://github.com/opencobra/COBRA.tutorials/raw/master/$tutorialFolder/$tutorialName.m"
-    dirHyperLink="https://github.com/opencobra/COBRA.tutorials/tree/master/$tutorialFolder"
+    dirHyperlink="https://github.com/opencobra/COBRA.tutorials/tree/master/$tutorialFolder"
```
laurentheirendt: 👍

src/analysis/exploration/biomassPrecursorCheck.m
```diff
@@ -1,7 +1,11 @@
-function [missingMets, presentMets,coupledMets] = biomassPrecursorCheck(model,checkCoupling)
+function varargout = biomassPrecursorCheck(model, checkCoupling, checkConservedQuantities)
```
laurentheirendt: May you change here to this format:
```
[missingMets, presentMets, coupledMets, missingCofs, presentCofs] = ...
```
in order to be consistent with all other functions?

src/analysis/exploration/biomassPrecursorCheck.m
```diff
@@ -79,6 +99,87 @@
...
+        
+    end
+    presentCofs = cofactorPair(producible);
+    missingCofs = cofactorPair(~producible);
+    metCofs = [cofactorPair{:}];
+    % Exclude those metabolites in cofactor pairs from missingMets
+    missingMets = missingMets(~ismember(missingMets, metCofs(:)));
+end
+
+% assign output argument
```
laurentheirendt: This then changes ...

test/verifiedTests/analysis/testBiomassPrecursorCheck/testBiomassPrecursorCheck.m
```diff
@@ -6,53 +6,131 @@
 % Authors:
 %    - Siu Hung Joshua Chan June 2018
 
-% save the current path
+% Save the current path
```
laurentheirendt: We generally try to keep the first word in the comment lowercase.

src/analysis/exploration/biomassPrecursorCheck.m
tpfau: Well... with varargout, you could (potentially) allow getting only `missingCofs` and `presentCofs` but not `coupledMets`.  Which you can't if you have fixed values. his was actually a request by me. 
The problem for me is, that you have two different things that you might want to check, and I would love to keep them separate. But yes, this makes the function inconsistent with other functions :(

shjchan:
Does it look better now?

laurentheirendt:
Thanks @shjchan 👍

PR #1265, opened 2018-06-22
Fixing bug with mCADRE creating non-unique values, which 

akaviaLab:
confuse addReactions
mCADRE would create repeating values of precursorMets (when running with checkFunctionality=1). This would end up confusing the COBRA toolbox when trying to add demand reactions (in checkModelFunction.m) and result in an error.

I've modified the code to have no loops, and no repeating values.

### DIFF
--- a/external/mCADRE/mCADRE.m
+++ b/external/mCADRE/mCADRE.m
@@ -71,30 +71,26 @@ function [tissueModel, coreRxn, nonCoreRxn,	zeroExpRxns, pruneTime, cRes] = mCAD
     end
 
 
-    display('Processing inputs and ranking reactions...')
+    disp('Processing inputs and ranking reactions...')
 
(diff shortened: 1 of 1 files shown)


laurentheirendt:
Is there any way to add a test for this functionality?

akaviaLab:
Can you be more specific about which functionality you are referring to?
addDemandReaction will call addMultipleReactions.m, which will fail if given the same reactions (line 76). If you call addDemandReaction(model, mets) with repeating mets (which is the previous case), addMultipleReactions will fail.

I guess you could add a test for mCADRE in general, since mCADRE will fail before my changes, but that seems overkill.
If you are talking about other functionality, can you tell me what you mean?

laurentheirendt:
Yes, a test for mCADRE in general would be nice. Let's keep that in mind. 👍

laurentheirendt:
Thanks @akaviaLab 👍

PR #1268, opened 2018-06-25
Fixing bug in generateRules caused by #1234

akaviaLab:
PR #1234 created a bug, where if genes needed to be added it is model.genes = addGenes(model, newGenes). However, addGenes returns a COBRA model full structure, so I fixed that bug.

There was also one linked bug
generateRules could deal with model.genes being empty, but not with it being missing.
addGenes would crash if model.genes was empty, but could deal with it being missing.

I've fixed that.

Right now, the tests will fail due to changes I've made to testGenerateRules, since I added checking both cases. Unfortunately, when model.genes is empty or missing, it will be repopulated in sorted order, which means that the rules will not match up.
I'm having it fail to ask the maintainers how to deal with this - I don't know how to add genes in a non-sorted manner.
One option is to compare model3 and model4 (and NOT compare them to the original model). Another option is not to check this subcases, and just leave testGenerateRules as it was before.

What do you think?

### DIFF
--- a/src/reconstruction/refinement/generateRules.m
+++ b/src/reconstruction/refinement/generateRules.m
@@ -1,4 +1,4 @@
-function [model] = generateRules(model)
+function [model] = generateRules(model, printLevel)
 % If a model does not have a model.rules field but has a model.grRules
 % field, can be regenerated using this script
 %
@@ -7,7 +7,9 @@ function [model] = generateRules(model)
 %    [model] = generateRules(model)
 %
 % INPUT:
-%    model:     COBRA model with model.grRules
+%    model:        COBRA model with model.grRules
+%    printLevel:   optional variable to print out all new genes 
+%                  (default = TRUE), can be zero if not needed
 %
 % OUTPUT:
 %    model:     same model but with model.rules added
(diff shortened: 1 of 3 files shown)


tpfau:
What I would do for the sorted model is to check for a few rules, whether those rules fit. I wouldn't test the full models since this will necessarily take a long time (converting all rules back and forth). 

Otherwise:
Thanks for spotting this.

src/reconstruction/refinement/generateRules.m
```diff
@@ -17,13 +19,22 @@
 %            - Diana El Assal 30/8/2017
 %            - Laurent Heirendt December 2017, speedup
 
+    if (nargin < 2)
```
laurentheirendt: `if  ~exist('printLevel', 'var')`

akaviaLab:
Fixed according to review
@tpfau I test 6 models to test generateRules, and it takes less than 1 minute on my laptop, even testing all rules. I'm leaving this as is, and not doing a subselection of rules, unless you think I should be testing all distributed models.

tpfau:
No its fine. Leave it as it is.

laurentheirendt:
Thanks @akaviaLab 👍

PR #1271, opened 2018-06-26
Adding a flag to allow a toolbox update to be optional 

tpfau:
Introducing a flag to `initCobraToolbox` that allows skipping the toolbox update. 
Changed the behaviour of all functions within the toolbox that call `initCobraToolbox` to not update the toolbox.


laurentheirendt:
All tutorials etc. have to be updated. 

Should it not be `initCobraToolbox(false)`, rather than `initCobraToolbox(0)`?

laurentheirendt:
Thanks @tpfau 👍

PR #1272, opened 2018-06-26
Updating the solveCobraXYZ input structure

tpfau:
This PR tests a changed input into the solveCobraXYZ functions.
Documentation will be updated when this is found to work.
At the moment it has a reduced test suite to allow quicker checks to see whether the changes work on all tested solvers.

### DIFF
--- a/src/base/solvers/getCobraSolverParams.m
+++ b/src/base/solvers/getCobraSolverParams.m
@@ -61,7 +61,8 @@ if isempty(valDef)
     valDef.saveInput = [];
     valDef.PbName = [solverType 'problem'];
     valDef.debug = 0;
-    
+    valDef.lifting = 0;
+       
     % GUROBI parameters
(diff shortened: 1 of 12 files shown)


rmtfleming:
Is this streamlining, or has the solveCobraXYZ interface changed functionality? 

e.g. is it still possible to pass in parameters specific to an instance of a particular type of solver, e.g., parameters directly to, and only for, gurobi?

tpfau:
This is mainly streamlining and clarifying what can go where and which order of parameter settings is being used, if there are contradicting settings. 
In essence the following order is being used (descending order of precedence):
1. Solver specific parameters provided as a solver specific struct - where applicable.
2. Cobra parameters provided in the same struct (which are then removed from the struct).
3. Cobra parameters passed as parameter/value pairs into the function
4. parameters set by setCobraSolverParams

This is how it is (mostly) handled at the moment, but there are some odd situations where things are not done in that order for some parameters

laurentheirendt:
Closing in favor of PR https://github.com/opencobra/cobratoolbox/pull/1282.

PR #1274, opened 2018-06-26
Adding a option to check for a website to prepareTest

tpfau:
Adding an option to check for access to a website to `prepareTest`.
This allows e.g. testReconMap to determine, if the vmh server is available and skip instead of fail, if it is not available. If something is wrong with the code, it will still fail, but it wont fail if the website is down.
The same for loadBiGGModel.


laurentheirendt:
Awesome! 👍

laurentheirendt:
Thanks @tpfau 👍 🎉

PR #1277, opened 2018-06-27
Updating Installation info wrt libssl

tpfau:
Some Matlab versions ship broken system libraries (in particular `libssl.so.1.0.0`) which need to be replaced by working ones in order to use webread without crashing matlab.

### DIFF
--- a/docs/source/installation/requirements.rst
+++ b/docs/source/installation/requirements.rst
@@ -62,14 +62,14 @@ Linux (Ubuntu or Debian)
 
     $ sudo apt-get install git-all curl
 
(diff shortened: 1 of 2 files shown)


tpfau:
We can put it in both places, but the problem is, that there is no error message (for the normal user). Matlab simply crashes without any comment.
If you are on a console, then you will get an error message, but not if you use the GUI.

laurentheirendt:
Ok, but then the crash behavior should be mentioned in the FAQ. 

You can include a link in the `requirements` section to the `FAQ`.

tpfau:
closing temporarily

tpfau:
Reopening

laurentheirendt:
Closing. See #1278.

PR #1294, opened 2018-07-07
Simplifying findRxnFromCompartment and printRxnFormula

akaviaLab:
I simplified findRxnFromCompartment and I think it sped up a bit.
I tried simplifying printRxnFormula - the tests will fail, since I can't match up the space composition with the reference data. It seems quirks due to strcat not including spaces at the end of a string, and similar issues.
Would it be possible to change the reference data? If not, can someone help me match the results to the reference data.

### DIFF
--- a/src/analysis/exploration/printRxnFormula.m
+++ b/src/analysis/exploration/printRxnFormula.m
@@ -75,15 +75,15 @@ end
 
 parser = inputParser();
 parser.addRequired('model',@isstruct) % we only check, whether its a struct, no details for speed
-parser.addParameter('rxnAbbrList',model.rxns,@(x) iscell(x) || ischar(x))
-parser.addParameter('printFlag',true,@(x) isnumeric(x) || islogical(x))
-parser.addParameter('lineChangeFlag',true,@(x) isnumeric(x) || islogical(x));
-parser.addParameter('metNameFlag',false,@(x) isnumeric(x) || islogical(x));
-parser.addParameter('fid',1, @isnumeric);
-parser.addParameter('directionFlag',false,@(x) isnumeric(x) || islogical(x));
-parser.addParameter('gprFlag',false,@(x) isnumeric(x) || islogical(x));
-parser.addParameter('proteinFlag',false,@(x) isnumeric(x) || islogical(x));
-parser.addParameter('printBounds',false,@(x) isnumeric(x) || islogical(x));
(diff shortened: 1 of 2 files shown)


src/analysis/exploration/printRxnFormula.m
```diff
@@ -75,15 +75,15 @@
...
-parser.addParamValue('rxnAbbrList',model.rxns,@(x) iscell(x) || ischar(x))
-parser.addParamValue('printFlag',true,@(x) isnumeric(x) || islogical(x))
-parser.addParamValue('lineChangeFlag',true,@(x) isnumeric(x) || islogical(x));
-parser.addParamValue('metNameFlag',false,@(x) isnumeric(x) || islogical(x));
-parser.addParamValue('fid',1, @isnumeric);
-parser.addParamValue('directionFlag',false,@(x) isnumeric(x) || islogical(x));
-parser.addParamValue('gprFlag',false,@(x) isnumeric(x) || islogical(x));
-parser.addParamValue('proteinFlag',false,@(x) isnumeric(x) || islogical(x));
-parser.addParamValue('printBounds',false,@(x) isnumeric(x) || islogical(x));
+parser.addParameter('rxnAbbrList',model.rxns,@(x) iscell(x) || ischar(x))
```
laurentheirendt: May you leave here `addParamValue` in order to keep the compatibility with MATLAB versions older than `R2013b`? Thanks.

src/analysis/exploration/printRxnFormula.m
```diff
@@ -158,105 +153,69 @@
         formulaStr = '';
         for j = 1:length(reactMets)
             if (j > 1)
-                if (printFlag)
-                    fprintf(fid, '+ ');
-                end
-                formulaStr = [formulaStr '+ '];
+                formulaStr = sprintf('%s+ ', formulaStr);
```
laurentheirendt: `formulaStr` should only be built if `printFlag` is set.

akaviaLab: This comment doesn't seem to make a lot of sense. In the original, fomulaStr was built always, since it is one of the return values.
In my program, it is built, and used for both printing and return, to reduce code duplication.

laurentheirendt: Sorry, I didn't see that `formulaStr` is actually returned in `formulas`. 😄

laurentheirendt:
Thanks @akaviaLab 👍

PR #1301, opened 2018-07-23
Adding src files of essentialRxns4ModelTutorial

laurentheirendt:
cc: @mapoliveira


laurentheirendt:
Thanks, @mapoliveira. 

Merging, but these functions need testing. May you submit another PR with tests? Thanks. 👍

PR #1306, opened 2018-07-27
Updated minerva redirect and response fix

bertonoronha:
With the new MINERVA update, there were 2 issues in need of solving

- Redirect `MapViewer` not working at the moment so replace with `minerva` in all calls
- Response format now different when overlay has been sent, so the test was failing - parsing user of the submitter to recognize that the overlay is successfully sent


laurentheirendt:
Great, @bertonoronha ! 🎉

PR #1320, opened 2018-08-28
Moiety code update

huldash:
Added new outputs and simplified code. Should be easier to read and maintain. Needed to update the test data with a new incidence matrix for the moiety graph (Lambda0) because the new version orders the nodes and edges slightly differently. The graph is still the same:

lambda0_1ddb37a

lambda0_9b75a4a


laurentheirendt:
Very nice PR, @huldash 👍 🎉

PR #1334, opened 2018-09-24
:bug: Default Values could be wrong

tpfau:
Fix to actually check the format of empty field default values for cell arrays (this caused a problem e.g. for subSystems, which was generated in an invalid way.


tpfau:
restarting CI

laurentheirendt:
Thanks @tpfau 👍

PR #1337, opened 2018-09-26
Small fix in writeCbmodel

tpfau:
Currently when the old style syntax is used and no filename is given, `writeCbmodel` errors with a missing argument exception.
This pr fixes the behaviour and makes the function request a name, if none is provided.


laurentheirendt:
Great @tpfau 👍

PR #1339, opened 2018-09-26
Allowed manual input of reaction and metabolite database in rBioNet

almut-heinken:
Enables the manual input of reaction and metabolite database instead of through GUI.

### DIFF
--- a/src/reconstruction/rBioNet/data2model.m
+++ b/src/reconstruction/rBioNet/data2model.m
@@ -148,7 +148,7 @@ model.metCharges = {};
 %
 
 % if reaction and metabolite database is input manually instead of through rBioNet GUI
-if isfield(database,'reactions') && isfield(database,'metabolites')
+if exist('database','var') && isfield(database,'reactions') && isfield(database,'metabolites')
     reactions=database.reactions;
     metabolites=database.metabolites;
 else
--- a/src/reconstruction/rBioNet/data2model.m
+++ b/src/reconstruction/rBioNet/data2model.m
@@ -1,4 +1,4 @@
-function model_out = data2model(data,description)
+function model_out = data2model(data,description,database)
 % Takes in data from `rbionet` to create a complete reconstruction
 %
 % USAGE:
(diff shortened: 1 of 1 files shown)


src/reconstruction/rBioNet/data2model.m
```diff
@@ -9,10 +9,18 @@
 %    data:          data from `rbionet`
 %    description:   model description
 %
+% OPTIONAL INPUT
```
laurentheirendt: This needs proper documentation formatting.

src/reconstruction/rBioNet/data2model.m
```diff
@@ -138,8 +146,14 @@
 % max_msgbox = 0;
 %
 
-reactions   = rBioNetSaveLoad('load','rxn');
-metabolites = rBioNetSaveLoad('load','met');
+% if reaction and metabolite database is input manually instead of through rBioNet GUI
+if nargin>2
```
laurentheirendt: May you use `if isfield(...)`?

almut-heinken: Is that really such a good idea? Then there is a "not enough input arguments" error when database is not entered and every script using this function would need to be rewritten.

laurentheirendt: Not sure I understand ...

almut-heinken: Never mind,...that should be isstruct, not isfield.

laurentheirendt: Sorry, I wanted to say: `if ~exist('database', 'var')` :smile:

src/reconstruction/rBioNet/data2model.m
```diff
@@ -147,7 +148,7 @@
 %
 
 % if reaction and metabolite database is input manually instead of through rBioNet GUI
-if nargin>2
+if isfield(database,'reactions') && isfield(database,'metabolites')
```
tpfau: You will still need to test, whether database actually exists
i.e.
```
if exist('database','var') && isfield(database,'reactions') && isfield(database,'metabolites')
...
```

almut-heinken: Ok, fine.

laurentheirendt:
Build on ARTENOLIS is failing because of an unrelated test.

laurentheirendt:
Thanks @almut-heinken 👍

PR #1342, opened 2018-09-28
Introduction of Additional Constraints and Variables

tpfau:
This PR introduces additional constraints and variables in a consistent manner.
Current coupling approaches are converted to use this approach.
It also adjusts several methods allowing them to take these constraints/variables into account.


tpfau:
I created a PR for the tutorial on COBRA.tutorials and will update the submodules in this PR once the tutorial is merged. 
I also added additional convenience functions for Variables/constraints in order to allow an easier updating/manipulation and added a test function (based on the tutorial).

tpfau:
I've seen that. This is nothing that my code is touching as it is a pure cplex methodology. 
I'll restart the CI to see if this error keeps popping up. If not, I think we will have to change this test.

laurentheirendt:
Nice job, @tpfau 👍 ⚽️

almut-heinken:
Hi @tpfau @laurentheirendt ,
the change in the coupleRxnList2Rxn function results in different model structures. Currently, the fastFVA function is still looking for the A matrix with coupling constraints, so we are unable to use it. Could you please also adapt fastFVA? 
Best,
Almut and @Fede-edef

PR #1350, opened 2018-10-11
Several small fixes related to glpk/minimal Installations

tpfau:
On a system without any advanced solvers, several tests fail due to glpk using feasibility limits and finding solutions different to e.g. cplex/gurobi. 
e.g. testCycleFreeFlux, assumed, that the solvers always produced a solution with a specific cycle, which is not necessarily true.
testBiomassPrecursorCheck had a hard comparison with 0, while it should compare with the optimality/feasibility tolerance instead of 0

Finally, for any not actively supported solver, changeCobraSolver simply did not set the corresponding global due to a condition of the category of the solver matching active for it to be set.
If this happens now, the solver will simply be set to the corresponding passively supported solver.

### DIFF
--- a/initCobraToolbox.m
+++ b/initCobraToolbox.m
@@ -517,7 +517,7 @@ function initCobraToolbox(updateToolbox)
             end
             k = 1;
             for j = 1:length(catSolverNames.(OPT_PROB_TYPES{i}))
-                if SOLVERS.(catSolverNames.(OPT_PROB_TYPES{i}){j}).installed
+                if SOLVERS.(catSolverNames.(OPT_PROB_TYPES{i}){j}).working
                     if k == 1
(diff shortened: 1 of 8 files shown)


initCobraToolbox.m
```diff
@@ -517,7 +517,7 @@ function initCobraToolbox(updateToolbox)
             end
             k = 1;
             for j = 1:length(catSolverNames.(OPT_PROB_TYPES{i}))
-                if SOLVERS.(catSolverNames.(OPT_PROB_TYPES{i}){j}).installed
+                if SOLVERS.(catSolverNames.(OPT_PROB_TYPES{i}){j}).working
```
laurentheirendt: Nice catch 👍

PR #1362, opened 2018-10-25
Storing invalid charges in Notes

tpfau:
Until the new FBC package finally provides key/value pairs, we will put invalid charges/formulas into the notes section. We read this part anyways and therefore don't loose model information during IO, and do not need to adapt code in the future as it will stay for compatability anyways.

This PR also adds a simplification for writeCbModel which allows its use as:
writeCbModel(model,'FileName') 


tpfau:
restarting CI... problem with 2015b parfor

laurentheirendt:
Thanks @tpfau 👍

PR #1366, opened 2018-10-26
addSinkReactions() argument checking

cladelpino:
🎨 Argument checking, added error and warning messages

### DIFF
--- a/src/reconstruction/refinement/addSinkReactions.m
+++ b/src/reconstruction/refinement/addSinkReactions.m
@@ -21,6 +21,14 @@ function [model, rxnsInModel] = addSinkReactions(model, metabolites, lb, ub)
 %
 % .. Author: - Ines Thiele 05/06/08
 
+if ~iscell(metabolites)
+    error('Metabolites must be in a cell array, even if it is only one');
+end
+if any(~ismember(metabolites,model.mets))
+    notFound = metabolites(~ismember(metabolites,model.mets)); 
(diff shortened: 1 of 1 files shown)


src/reconstruction/refinement/addSinkReactions.m
```diff
@@ -21,6 +21,14 @@
 %
 % .. Author: - Ines Thiele 05/06/08
 
+if ~iscell(metabolites)
+    error('Metabolites must be in a cell array, even if it is only one');
+end
```
tpfau: Just as a question, wouldn't it be more convenient if a char would also be acceptable? 
I.e. instead of throwing an error if its not a cell, just to encapsulate it in a cell?

cladelpino: Yes. I didn't want to modify the original behaviour besides issuing the warning.. something like ?:

if ~iscell(metabolites) & ischar(metabolites) 
 metabolites = {metabolites}
end

Which assumes it is a single dimensional char with only one metabolite.

tpfau: Looks good to me.
In the end, the only change is that we allow additional inputs so any old code will still work.

laurentheirendt:
@cladelpino, may you add this small change? Thanks. 👍

laurentheirendt:
Thanks @cladelpino 👍

PR #1374, opened 2018-11-06
Test calculate LOD function

loicmarx:
A modification of the function has been made and the function has been tested.

I hereby confirm that I have:

### DIFF
--- a/src/dataIntegration/metabotools/calculateLODs.m
+++ b/src/dataIntegration/metabotools/calculateLODs.m
@@ -14,28 +14,22 @@ function [lod_mM] = calculateLODs(theo_mass,lod_ngmL)
 %   lod_mM:            Detection limits in mM
 %
 % .. Author: - Maike K. Aurich 27/05/15
-%    Modified by Loic Marx, November 2018
+%            - Modified by Loic Marx, November 2018
 
 % input checking
-if length(lod_ngmL) == 1
-    
+if size(lod_ngmL) == 1
     lod_ngmL = repmat(lod_ngmL, 1, length(theo_mass)); 
-    
-end    
+    end    
 
-if numel(lod_ngmL) ~= numel(theo_mass)
-    
+if size(lod_ngmL) ~= size(theo_mass)
     error('The number of elements in the input vectors do not match. They have to be either the same size, or lod_ngmL has to be a single value which is used for all elements');
 end
 
 lod_gL = lod_ngmL * 1e-6; 
-
 lod_mM = zeros(length(lod_gL), 1);
 
-for i=1 : length(lod_gL)
-
+for i = 1 : length(lod_gL)
     lod_mM(i) = lod_gL(i)/theo_mass(i)*1000;
-
 end
 
 end
--- a/src/dataIntegration/metabotools/calculateLODs.m
+++ b/src/dataIntegration/metabotools/calculateLODs.m
@@ -14,26 +14,30 @@ function [lod_mM] = calculateLODs(theo_mass,lod_ngmL)
 %   lod_mM:            Detection limits in mM
 %
 % .. Author: - Maike K. Aurich 27/05/15
-%  Modified by Loic Marx
+%  Modified by Loic Marx, November 2018
 
 % input checking
 if length(lod_ngmL) == 1
+    
     lod_ngmL = repmat(lod_ngmL, 1, length(theo_mass)); 
+    
     % throw a warning
-    warning('The inputs have different size');
+    warning('The inputs have different size but we managed to fix this problem');
+    
 end    
 
 if length(lod_ngmL) ~= length(theo_mass)
-    error('both inputs do not have the same size');
+    
+    error('The length of the inputs are not the same');
 end
 
-lod_gL = lod_ngmL* 0.000001;
+lod_gL = lod_ngmL * 0.000001;
 
 lod_mM = zeros(length(lod_gL), 1);
 
-for i=1:length(lod_gL)
+for i=1 : length(lod_gL)
 
-    lod_mM(i)= lod_gL(i)/theo_mass(i)*1000;
+    lod_mM(i) = lod_gL(i)/theo_mass(i)*1000;
 
 end
 
--- a/src/dataIntegration/metabotools/calculateLODs.m
+++ b/src/dataIntegration/metabotools/calculateLODs.m
@@ -21,17 +21,14 @@ if length(lod_ngmL) == 1
     
     lod_ngmL = repmat(lod_ngmL, 1, length(theo_mass)); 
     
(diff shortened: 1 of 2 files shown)


src/dataIntegration/metabotools/calculateLODs.m
```diff
@@ -14,14 +14,30 @@
...
+    
+end    
 
-lodmM =[];
+if length(lod_ngmL) ~= length(theo_mass)
+    
+    error('The length of the inputs are not the same');
+end
+
+lod_gL = lod_ngmL * 0.000001;
```
tpfau: I would suggest to change this to:
```
lod_gL = lod_ngmL * 1e-6
```
for readability.

src/dataIntegration/metabotools/calculateLODs.m
```diff
@@ -14,14 +14,30 @@
...
+%  Modified by Loic Marx, November 2018
 
-lod_gL = lod_ngmL* 0.000001;
+% input checking
+if length(lod_ngmL) == 1
+    
+    lod_ngmL = repmat(lod_ngmL, 1, length(theo_mass)); 
+    
+    % throw a warning
+    warning('The inputs have different size but we managed to fix this problem');
```
tpfau: I'd suggest to remove the warning and update the documentation, that the inputs can either be a single value which is used for every lod or a vector of individual lods.

src/dataIntegration/metabotools/calculateLODs.m
```diff
@@ -14,14 +14,30 @@
...
+    
+end    
 
-lodmM =[];
+if length(lod_ngmL) ~= length(theo_mass)
```
tpfau: what you are looking for here is not `length` but `numel`. 
`[2,2;2,2]` and `[2,1]` have the same length but are still invalid inputs.

test/verifiedTests/dataIntegration/testMetaboTools/testCalculateLODs.m
```diff
@@ -0,0 +1,53 @@
...
+calculateLOD = calculateLODs(theo_mass,lod_ngmL);
+
+% Comparison between calculate value and ref value
+
+assert(isequal(calculateLOD, v_ref))
+
+% Throwing an error if the dimensions of both inputs are not equal
+lod_ngmL_unequal = [1, 2, 3];
+
+assert(verifyCobraFunctionError('calculateLODs', 'inputs', {theo_mass,lod_ngmL_unequal'}, 'testMessage', 'both inputs do not have the same size'));
```
tpfau: our tested message is not the same as the error message thrown. 
This should be  `'The length of the inputs are not the same'`, and I would actually suggest to improve that message:
`'The number of elements in the input vectors do not match. They have to be either the same size, or lod_ngmL has to be a single value which is used for all elements'`

src/dataIntegration/metabotools/calculateLODs.m
```diff
@@ -14,26 +14,30 @@
 %   lod_mM:            Detection limits in mM
 %
 % .. Author: - Maike K. Aurich 27/05/15
-%  Modified by Loic Marx
+%  Modified by Loic Marx, November 2018
```
laurentheirendt: This is not properly formatted.

src/dataIntegration/metabotools/calculateLODs.m
laurentheirendt: Sounds good. @loicmarx, may you do that? 👍

test/verifiedTests/dataIntegration/testMetaboTools/testCalculateLODs.m
```diff
@@ -9,34 +9,45 @@
...
 currentDir = pwd;
 
-% initialize the test
+% Initialize the test
 fileDir = fileparts(which('testcalculateLODs'));
+
 cd(fileDir);
 
-% calculate reference value
+% Defining the inputs
```
laurentheirendt: `% Define the inputs`

test/verifiedTests/dataIntegration/testMetaboTools/testCalculateLODs.m
```diff
@@ -9,34 +9,45 @@
...
 v_ref = 1e-3 * [1; 0.5];
 
-%calculate inputs
-lod_ngmL = 1;
+% Calculate the actual value by using the function
 
 calculateLOD = calculateLODs(theo_mass,lod_ngmL);
-% Comparison to ref value
+
+% Comparison between calculate value and ref value
```
laurentheirendt: `% Compare between the calculated value and the reference value`

test/verifiedTests/dataIntegration/testMetaboTools/testCalculateLODs.m
```diff
@@ -9,34 +9,45 @@
...
 
 assert(isequal(calculateLOD, v_ref))
-% throwing an error if the dimensions are not equal
+
+% Throwing an error if the dimensions of both inputs are not equal
```
laurentheirendt: `Throw ...`

test/verifiedTests/dataIntegration/testMetaboTools/testCalculateLODs.m
```diff
@@ -9,34 +9,45 @@
...
 lod_ngmL_unequal = [1, 2, 3];
 
 assert(verifyCobraFunctionError('calculateLODs', 'inputs', {theo_mass,lod_ngmL_unequal'}, 'testMessage', 'both inputs do not have the same size'));
-% Testing the case where lod_ngmL is a vector of the same size as theo_mass
 
+% If both inputs are two vectors of the same size
```
laurentheirendt: This comment is not very clear.

test/verifiedTests/dataIntegration/testMetaboTools/testCalculateLODs.m
```diff
@@ -9,34 +9,45 @@
...
 lod_ngmL_vector = [1, 2];
+
 v_ref_vector = 1e-3 * [1; 1];
-% Comparison to the ref value
+
+% Comparison between the values calculated when the inputs have the same size and the reference value
```
laurentheirendt: `Compare ... `

src/dataIntegration/metabotools/calculateLODs.m
laurentheirendt: I prefer `size()`...

tpfau: Yes, but.
Size looks into both dimensions, at which point you have to either make a complicated statement or you only allow input that is either a column or a row vector.

laurentheirendt:
Thanks @loicmarx 👍

PR #1380, opened 2018-11-14
Test exportSetToGAMS

loicmarx:
The function exportSetToGAMS has been updated and a test has been created.

I hereby confirm that I have:


tpfau:
Do you have any GAMS reference data, or anything that you could compare the outputs with?
Worst case, you could run the function once on a problem/model and then write a test, that those results are reproduced. 
But at the moment, the test doesn't really test anything but that the error message is thrown if the wrong number of inputs is given.

laurentheirendt:
@loicmarx ?

laurentheirendt:
Thanks @loicmarx 👍

PR #1388, opened 2018-11-25
Taking account of numeric version of genes

arichelle:
I modify these two codes to take into account he possibility that gene field is numeric


arichelle:
You completely right. I think I took the bad habit to transform it in a numeric array to handle it more easily in Matlab. I guess that it would be more efficient to just transform the part of "code_preprocessing" in Cobra_paper to perform a proper transformation from HGNC to entrez ID rather than adding little spec ;)
I close this PR.

PR #1390, opened 2018-11-26
Test function for verifyProblem

tpfau:
Test function for verifyProblem

### DIFF
--- a/src/base/solvers/verifyCobraProblem.m
+++ b/src/base/solvers/verifyCobraProblem.m
@@ -49,7 +49,7 @@ statusOK = 1;
 
 %Check A
 if ~isfield(XPproblem, 'A')
-    disp('Required field A not found');
+    warning('Required field A not found');
     statusOK = -1;
     return;
 elseif any(isnan(XPproblem.A(:)))
@@ -58,7 +58,7 @@ elseif any(isnan(XPproblem.A(:)))
     for i=1:length(r)
         strCoords = [strCoords ' ' num2str(r(i)) ',' num2str(c(i))];
     end
-    disp(['NaN present in A matrix at' strCoords '.']);
+    warning(['NaN present in A matrix at' strCoords '.']);
     statusOK = -1;
     return;
 end
@@ -67,7 +67,7 @@ end
 
 %Check b
 if ~isfield(XPproblem, 'b')
-    disp('Required field b not found');
+    warning('Required field b not found');
     statusOK = -1;
     return;
 elseif any(isnan(XPproblem.b))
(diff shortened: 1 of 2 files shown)


tpfau:
reopening to trigger CI

tpfau:
Trying a restart (failing tests don't seem to have anything to do with the current PR.

src/base/solvers/verifyCobraProblem.m
```diff
@@ -146,7 +154,7 @@
 
 %check ub
 if ~isfield(XPproblem,'ub')
-    diplay('Required field ub not found');
+    display('Required field ub not found');
```
laurentheirendt: Why are you using sometimes `display` and sometimes `disp`?

test/verifiedTests/base/testSolvers/testVerifyCobraProblem.m
```diff
@@ -0,0 +1,242 @@
...
+problem.c = [0;1];
+res = verifyCobraProblem(problem);
+assert(res == 1);
+% remove A
+problem = rmfield(problem,'A');
+res = verifyCobraProblem(problem);
+assert(res == -1);
+%Add invalid A
+problem.A = [1,0;NaN,1];
+res = verifyCobraProblem(problem);
```
laurentheirendt: Instead of testing for `== -1`, would it make sense to test for the error itself?

test/verifiedTests/base/testSolvers/testVerifyCobraProblem.m
```diff
@@ -0,0 +1,242 @@
...
+problem.F = [1;1];
+res = verifyCobraProblem(problem);
+assert(res == -1);
+
+%Add valid F - this now is finally a valid Problem.
+problem.F = [1,0;0,1];
+res = verifyCobraProblem(problem);
+assert(res == 1);
+
+%Add invalid vartype
```
laurentheirendt: Would it be possible to eventually write this test differently, i.e. with a loop?

src/base/solvers/verifyCobraProblem.m
tpfau: because personally, I write "disp", but the erronous diplay was in the code and I just corrected it.

test/verifiedTests/base/testSolvers/testVerifyCobraProblem.m
tpfau: if it would be thrown as an error: sure. 
But since It's just a display output, not really, except having loads of "diary" commands, which I would like to avoid.

test/verifiedTests/base/testSolvers/testVerifyCobraProblem.m
tpfau: not really, as all the different settings are not necessarily the same for the individual fields.
Some are matricies, some are vectors, some need chars some need doubles etc pp...

src/base/solvers/verifyCobraProblem.m
tpfau: Also, the code originally had a mix of it as well.

test/verifiedTests/base/testSolvers/testVerifyCobraProblem.m
laurentheirendt: Yes, I meant to actually through an error ... Maybe a `soft` and a `hard` mode?

test/verifiedTests/base/testSolvers/testVerifyCobraProblem.m
laurentheirendt: ok ... I just thought as the tests look quite repetitive.

test/verifiedTests/base/testSolvers/testVerifyCobraProblem.m
tpfau: I changed them to warnings, which can be validated via lastwarn. 
A hard mode is not really an option, as it could break existing code that relies on a soft return value instead of an error.

laurentheirendt:
Thanks @tpfau 👍

PR #1395, opened 2018-11-28
Function and test to determine the binary form of S and connectivities

laurentheirendt:


src/analysis/subspaces/determineBinaryMatrix.m
```diff
@@ -0,0 +1,35 @@
...
+% Determine the binary form of the stoichiometric matrix S
+%
+% INPUT:
+%
+%    S:            Stoichiometric matrix of size m x n (m rows, n columns)
+%
+% OUTPUTS:
+%
+%    Shat:         Binary form of the stoichiometric matrix
+%    Shatabs:      Absolute value of Shat
```
tpfau: From the name of the function, I would actually expect it to only return Shatabs, or even just a logical sparse with `Sbin = S~=0`

src/analysis/subspaces/determineBinaryMatrix.m
```diff
@@ -0,0 +1,35 @@
...
+% .. Author: - Laurent Heirendt, June 2018
+
+    Shat = S;  % initialize the binary matrix
+    Shat(S > 0) = 1;  % products, produced -> kin
+    Shat(S < 0) = -1;  % reactants, consumed -> kout
+
+    % determine the row connectivity
+    Shatabs = abs(Shat);
+    if nargout > 2
+        mconnect = sort(sum(Shatabs, 2), 'descend');
```
tpfau: sorting looks very odd here, I would expect the outcome to be linked to each respective row, while by sorting, you actually change that order.

src/analysis/subspaces/determineBinaryMatrix.m
laurentheirendt: What other name do you suggest?

src/analysis/subspaces/determineBinaryMatrix.m
laurentheirendt: I need them to be sorted, but I can specify that in the header.

tpfau: But why put that in this function? Couldn't you sort them afterwards?

src/analysis/subspaces/determineBinaryMatrix.m
tpfau: `determineSignMatrix`?

src/analysis/subspaces/determineBinaryMatrix.m
tpfau: And why are the other outputs not sorted? This is confusing (imo)

src/analysis/subspaces/determineBinaryMatrix.m
```diff
@@ -0,0 +1,35 @@
...
+%    mconnect:     Compound connectivity (connectivity number; sum of the number of non-zero entries in a row)
+%    nconnect:     Reaction participation number (sum of the number of non-zero entries in a column)
+%    mconnectin:   Produced compound connectivity
+%    mconnectout:  Consumed compound connectivity
+%
+% .. Author: - Laurent Heirendt, June 2018
+
+    Shat = S;  % initialize the binary matrix
+    Shat(S > 0) = 1;  % products, produced -> kin
+    Shat(S < 0) = -1;  % reactants, consumed -> kout
```
tpfau: isnt this just:
Shat = sign(S) (which is about 50% faster)

src/analysis/subspaces/determineBinaryMatrix.m
laurentheirendt: Ok I can change that.

src/analysis/subspaces/determineBinaryMatrix.m
tpfau: I think a flag would make sense (but (imo) that flag should then sort everything, becuase currently this is somewhat of an inconsistency in the function.

laurentheirendt: Yes, I will implement such a flag.

laurentheirendt:
Thanks for the review, @tpfau 👍

rmtfleming:
Binary is only 0,1. With -1, 0, 1 it is a sign pattern matrix, or sign for short.

PR #1396, opened 2018-11-29
Adding metComps and utilities for Annotations

tpfau:
This PR adds further utility functions for annotations and updates the existing functions (i.e. add/getMIRIAMAnnotation) to be more flexible with respect to outputs (potentially allowing this to be used during IO).
It also introduces the metComps field and adjusts several things required for this change. 
I'm not sure, whether I caught all fields which rely on the metabolite names, but I actually haven't found many that do. And those that did were adjusted.
We will still need to "handle" compartment IDs in id names, if we want to find the "basic" metabolite for them (or rely on external ids for that purpose) but this PR should address at least some of it.

- [X] Selected `integrationRAVEN` as a target branch (top left drop-down menu)


rmtfleming:
@ithiele this may affect Harvey computations. 
@laurentheirendt It needs checked before we can accept this PR.

laurentheirendt:
@tpfau, is there any particular test that ensures backward compatibility?

tpfau:
It passes all existing tests with only minimal modifications which are mainly due to the new fields being added and equality comparisons between existing models and manually manipulated models, as well as ensuring that the new fields are covered by the tests.
Not sure what kind of other backward compatability tests could be used for otherwise untested functionality.

laurentheirendt:
@tpfau, may you do a rebase here? Thanks 👍

laurentheirendt:
@tpfau, may you do a rebase here? Thanks.

laurentheirendt:
This can actually be safely merged to `integrationRAVEN`. Further tests will be needed once that branch will be merged to `develop`.

PR #1400, opened 2018-12-03
rMTA

lvalcarcel:
Implementation of a algorithm that we are submitting to Bioinformatics: 
**_"rMTA: Robust Metabolic Transformation Analysis"_**

> Motivation:
 The development of computational tools exploiting -omics data and high-
quality genome-scale metabolic networks for the identification of novel drug targets is a 
relevant topic in Systems Medicine. Metabolic Transformation Algorithm (MTA) is one 
of  these  tools,  which  aims  to  identify  targets  that  transform  a  disease  metabolic  state 
back  into  a  healthy  state,  with  potential  application  in  any  disease  where  a  clear 
metabolic alteration is observed.

This implementation includes auxiliary functions as well as the base algorithm _MTA_.


src/analysis/rMTA/rMTA.m
```diff
@@ -0,0 +1,726 @@
...
+%
+% .. Authors:
+%       - Luis V. Valcarcel, 03/06/2015, University of Navarra, CIMA & TECNUN School of Engineering.
+% .. Revisions:
+%       - Luis V. Valcarcel, 26/10/2018, University of Navarra, CIMA & TECNUN School of Engineering.
+%       - Francisco J. Planes, 26/10/2018, University of Navarra, TECNUN School of Engineering.
+
+
+
+%% Check solver to be used
```
tpfau: see my comment above wrt optional input arguments.

src/analysis/rMTA/rMTA.m
```diff
@@ -0,0 +1,726 @@
+function [TSscore, deletedGenes, Vres] = rMTA(model, rxnFBS, Vref, alpha, epsilon, options)
+% Calculate robust Metabolic Transformation Analysis (rMTA) using the
+% solver CPLEX.
+% Code was prepared to be able to be stopped and be launched again.
+% After every run it is neccesary to eliminate the file "temp_rMTA.mat"
```
tpfau: Would it make sense to just clean it in the code itself when finishing?

src/analysis/rMTA/rMTA.m
```diff
@@ -0,0 +1,726 @@
...
+%
+% .. Authors:
+%       - Luis V. Valcarcel, 03/06/2015, University of Navarra, CIMA & TECNUN School of Engineering.
+% .. Revisions:
+%       - Luis V. Valcarcel, 26/10/2018, University of Navarra, CIMA & TECNUN School of Engineering.
+
+
+%% --- check the inputs ---
+
+if nargin<3
```
tpfau: as above optional arguments.

laurentheirendt:
@lvalcarcel, very nice PR! 👍 

I see that there are many internal functions - would it be possible to split these into separate functions/files? This makes the code overall easier to maintain.

Are you planning to add tests? This would be a requirement to ensure high quality. More information on how to write a test is [here](https://opencobra.github.io/cobratoolbox/stable/contributing.html#guide-for-writing-a-test).

src/analysis/rMTA/MTA.m
```diff
@@ -0,0 +1,531 @@
...
+else
+    if isfield(options,'rxnKO') rxnKO = options.rxnKO; else rxnKO = false;   end
+    if isfield(options,'timelimit') timelimit = options.timelimit; else timelimit = inf;   end
+    if isfield(options,'separate_transcript') separate_transcript = options.separate_transcript; else separate_transcript = '';   end
+    if isfield(options,'numWorkers') numWorkers = options.numWorkers; else numWorkers = 0;   end
+    if isfield(options,'printLevel') printLevel = options.printLevel; else printLevel = 1;   end
+end
+
+
+fprintf('===================================\n');
```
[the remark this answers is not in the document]
lvalcarcel: Done!

src/analysis/rMTA/MTA.m
```diff
@@ -0,0 +1,531 @@
...
+%
+% .. Authors:
+%       - Luis V. Valcarcel, 03/06/2015, University of Navarra, CIMA & TECNUN School of Engineering.
+% .. Revisions:
+%       - Luis V. Valcarcel, 26/10/2018, University of Navarra, CIMA & TECNUN School of Engineering.
+%       - Francisco J. Planes, 26/10/2018, University of Navarra, TECNUN School of Engineering.
+
+
+
+%% Check solver to be used
```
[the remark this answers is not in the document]
lvalcarcel: I have implemented the method for other solvers, although the function to solve the MIQP still is able to detect CPLEX and use it. 
I have not programmed for other solvers as I only have IBM Cplex

src/analysis/rMTA/rMTA.m
```diff
@@ -225,28 +225,17 @@
...
-cplex_model_moma.lb = model.lb;
-cplex_model_moma.ub = model.ub;
-cplex_model_moma.lhs = zeros(size(model.mets));
-cplex_model_moma.rhs = zeros(size(model.mets));
-cplex_model_moma.obj = c;
-cplex_model_moma.Q = Q;
-cplex_model_moma.sense = 'minimize';
-cplex_model_moma.ctype = ctype;
-fprintf('\tcplex model for MOMA built\n');
+QPproblem = struct();
```
tpfau: What about using `buildLPProblemFromModel`?
That way you will Cover C and E matricies without worring about them. 
In this instance it would look like:
```
[nRxns,nMets] = size(model.S);
if isfield(model,E)
 nVars = size(model.E,2);
else 
 nVars = 0;
end
if isfield(model,E)
 nCtrs = size(model.C,1);
else 
 nCtrs = 0;
end
QPproblem = buildLPProblemFromModel(model);
QPproblem.c(1:nRxns) = -2*Vref;
QPproblem.F = [2*speye(nRxns),sparse(nRxns,nVars),sparse(nVars,nRxns+nVars)
QPproblem.osense = 1;
```
Also, is it intentional, that you reset any potential settings made in model.b /model.csense ?

lvalcarcel: Hi! 
It seems prety interesting, but I am just performing MOMA analysis to obtain the resulting flux distribution after deleting some reactions. 
Furthermore, I have not taken into account C and E matrix for the MIQP problems, so it would be not accurate to use them in the MOMA formulation.

On the other hand, I have looked at the help of `buildLPProblemFromModel` and it is not clear to me how to use this tool. I suppose that E matrix are additional variables that are after the flux variables, that C matrix does not include S matrix, and that the order of the constrains must be S, C, E and D. As I am not sure how to use it correctly I prefer not to use it.

tpfau: Hi,
There is a tutorial for these additional Constraints/Variables (https://github.com/opencobra/COBRA.tutorials/blob/master/reconstruction/constrainingModels/tutorial_AdditionalConstraintsAndVariables.m)
`buildLPProblemFromModel` essentially builds an LP from the given model with
```
A = [S , E;...
       C , D];
```
and the appropriate associated fields (bounds, constraint senses etc). Overall its just a helper function to make code less dependent on the internal model structure.

rmtfleming: All one can do is suggest such changes, they are not required. In this case, the code to build the problem structure does the same thing, so we should not delay this PR because of it.

lvalcarcel: Thanks @tpfau !

I was not able to find the tutorial for the `buildLPProblemFromModel ` function. I will try to use it with the next upgrade to the code.

tpfau: The reason I'm suggesting it is to avoid getting new functions in which do not support the whole COBRA structure, i.e. do calculations which might miss properties set in the model.

lvalcarcel: I have uploaded a commit using `buildLPproblemFromModel`  function for the moma part of the algorithm. I have also changed the name of one function, from `MTA_model` to  `buildMTAproblemFromModel`, to have some similarities in the nomenclature of the code

src/analysis/rMTA/MTA_MIQP.m
```diff
@@ -0,0 +1,141 @@
...
+% .. Revisions:
+%       - Luis V. Valcarcel, 26/10/2018, University of Navarra, CIMA & TECNUN School of Engineering.
+
+% Check the input information
+p = inputParser;
+% check requiered arguments
+addRequired(p, 'OptimizationModel');
+addRequired(p, 'KOrxn');
+% Check optional arguments
+addOptional(p, 'numWorkers', 0);
```
tpfau: Do you want them as optional arguments (requires the right positions) or as parameters?
Personally, I always prefer parameters, as they make later modifications of the header much easier without breaking backward compatability, and, especially with a larger number of arguments, they make (imo) reading the calls easier, as it is clearer, what each value means.

src/analysis/rMTA/MTA.m
```diff
@@ -0,0 +1,219 @@
...
+addRequired(p, 'Vref', @isnumeric);
+% Check optional arguments
+addOptional(p, 'alpha', 0.66, @isnumeric);
+addOptional(p, 'epsilon', 0, @isnumeric);
+% Add optional name-value pair argument
+addParameter(p, 'rxnKO', false);
+addParameter(p, 'timelimit', inf);
+addParameter(p, 'SeparateTranscript', '');
+addParameter(p, 'numWorkers', 0);
+addParameter(p, 'printLevel', 1);
```
tpfau: You could also add validation functions (similar to the optional arguments) here, to give more precise errors, when something wrong is supplied.

src/analysis/rMTA/calculateEPSILON.m
```diff
@@ -0,0 +1,62 @@
...
+%    minimum:           Minimun value for epsilon requiered (default=1e-3)
+%
+% OUTPUT:
+%    epsilon            Numeric value or array with the epsilon for the different reactions
+% 
+% .. Authors:
+%       - Luis V. Valcarcel, 06/07/2015, University of Navarra, CIMA & TECNUN School of Engineering.
+% .. Revisions:
+%       - Luis V. Valcarcel, 26/10/2018, University of Navarra, CIMA & TECNUN School of Engineering.
+%       - Francisco J. Planes, 26/10/2018, University of Navarra, TECNUN School of Engineering.
```
tpfau: This function seems to be never called in your tests. 
Would it make sense to write an explicit test Function for it, or modify tthe tests to also run this function?

test/verifiedTests/analysis/testMTA/testMTA.m
```diff
@@ -0,0 +1,84 @@
...
+% Purpose:
+%     - Check the functions MTA and rMTA using different solvers
+%
+% Authors:
+%     - Luis V. Valcarcel 2018-12-17
+%
+
+global CBTDIR
+% prepareTest('needsMIQP', true, 'useSolversIfAvailable', {'tomlab_cplex', 'ibm_cplex', 'gurobi'});
+prepareTest('needsQP', true, 'useSolversIfAvailable', {'tomlab_cplex', 'ibm_cplex', 'gurobi'});
```
tpfau: this should be `solverPkgs = prepareTest('needsQP', true, 'useSolversIfAvailable', {'tomlab_cplex', 'ibm_cplex', 'gurobi'});`
and later use the respective solverPackage.QP /LP etc fields.

lvalcarcel:
I have included the test for the remaining functions, updated the solver search and the input parser of arguments

laurentheirendt:
Thanks @lvalcarcel and @tpfau 👍

lvalcarcel:
Thanks @tpfau @laurentheirendt and @rmtfleming

PR #1401, opened 2018-12-03
Cplex miqp

tpfau:
Adding a MIQP interface for ibm_cplex. 
Removing the MIQP flag from cplex_direct (its not present in the solveCobraMIQP file).

### DIFF
--- a/src/base/solvers/solveCobraMIQP.m
+++ b/src/base/solvers/solveCobraMIQP.m
@@ -211,6 +211,8 @@ switch solver
            stat = -1; % Solution not optimal or solver problem
         end
         solStat = stat;
+        
+        
     case 'gurobi'
      % Free academic licenses for the Gurobi solver can be obtained from
         % http://www.gurobi.com/html/academic.html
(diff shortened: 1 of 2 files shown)


src/base/solvers/solveCobraMIQP.m
```diff
@@ -278,6 +280,87 @@
...
+            b_L = MILPproblem.b_L;
+            b_U = MILPproblem.b_U;
+        else
+            b_L = b;
+            b_U = b;
+        end
+        intVars = (vartype == 'B') | (vartype == 'I');
+        % intVars
+        % pause;
+        cplexlp.Model.A = A;
```
laurentheirendt: Is this code from `solveCobraLP`? Why not reuse it?

tpfau: Its essentially code from solveCobraMILP.
But yes, makes sense to make a "buildCplexLP" function. I'll have a look at that.

laurentheirendt: Yes, that would be nice - that will make the code even more maintainable.

tpfau:
closing for various updates.

PR #1402, opened 2018-12-03
Testing alternative to MOCov

tpfau:
This is a test to try an alternative to MOCov since the matlab profiler sometimes looses Code Coverage.
It writes a valid cverage.json file which should be sufficient for CodeCov. This is just to test, whether this code cov calculation works.


laurentheirendt:
Why not modify MoCov instead? https://github.com/uni-lu/MoCov

tpfau:
Because from what I've seen when playing around a bit, is that the profiler sometimes "looses" code coverage data. 
Therefore, this modification extracts the profiling info after each individual test, and updates the coverage accordingly, which is something that the MoCov concept would not allow (running once). 
Also, this piece of code does not provide all facilities provided by MoCov (e.g. html output).

laurentheirendt:
I prefer this code being either part of MoCov or in a separate repo so it can be reused in other repos.

tpfau:
Well... I can try to move the whole testSuite part into a separate repo, because it is interconnected and can't really be separated.

laurentheirendt:
What do you mean by `the whole testSuite part`?

tpfau:
runTestSuite, runScript etc. This codecov is woven into it and not separate.

laurentheirendt:
... hence my reasoning to actually modify MoCov. Moving the whole testSuite out is not a good idea.

tpfau:
But as I said:
Modifying MoCov is not really an option, as it would not address the issue I want to address (profiler loosing info) at all.

laurentheirendt:
But could we add this functionality then to MoCov?

PR #1403, opened 2018-12-04
Fix regulatory fba

tpfau:
When looking into the `solveBooleanRegModel` function, I noticed an odd call to `optimizeCbModel`:
`fbasolDR = optimizeCbModel(modelDR,'max',true);`
This call looks like it was originally intended to pass some flag as the third argument, but the third argument of optimizeCbModel is minNorm, and true (which matlab will interpret as a value of 1) will lead to a solution that, in addition to `osense*c` will minimize the quadratic fluxes of the solution with a weight of 1 for the quadratic fluxes. I think this is not what is supposed to happen here, where actually, the effect of the knockdown of several genes is the desired target.
I corrected this (in y opinion) obvious bug and adapted the test data accordingly.


laurentheirendt:
Why did you completely remove `test/verifiedTests/analysis/testrFBA/refData_dynamicRFBA.mat`?

tpfau:
its not completely removed. its only modified. 
otherwise that test would fail.

laurentheirendt:
Ok, weird, Github indicates a 100% negative change, but maybe that's because it is a binary file. All good 👍

laurentheirendt:
Thanks @tpfau 👍

PR #1405, opened 2018-12-05
Exclude qpng

tpfau:
Exclude the qpng solver from the MOMA test.
qpng does not work with normal models due to them commonly not being full rank. 
We might need to find a way to make those models fit to the qpng scheme.

### DIFF
--- a/test/verifiedTests/analysis/testMOMA/testMOMA.m
+++ b/test/verifiedTests/analysis/testMOMA/testMOMA.m
@@ -30,7 +30,7 @@ cd(fileDir);
 model = getDistributedModel('ecoli_core_model.mat');
 
 % test solver packages
-solverPkgs = prepareTest('needsLP', true, 'needsQP', true, 'excludeSolvers', 'pdco');
+solverPkgs = prepareTest('needsLP', true, 'needsQP', true, 'excludeSolvers', {'qpng','pdco'});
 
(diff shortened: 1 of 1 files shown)


rmtfleming:
If qpng is called, you could check the row rank deficiency, and if it is greater than zero, invoke row reduction
https://opencobra.github.io/cobratoolbox/stable/modules/analysis/subspaces/rowReduce/index.html?highlight=rowreduce

tpfau:
Could you implement that please? 
As of now, qpng can't be used for the test specified in testMOMA, which makes that test fail on the extensive branch.

laurentheirendt:
@rmtfleming, we need to have this fix in, as the jobs on the CI are failing.

tpfau:
For now, I modified our qpng interface to convert Equality constraints to two inequality constraints, which don't have the rank deficiency restriction), as a quick try with `rowReduce` led to the returned Matrix being empty.

laurentheirendt:
@tpfau, may you remove [e7fc7ea](
https://github.com/opencobra/cobratoolbox/pull/1405/commits/e7fc7ea7188528a923374a1c9838ec7b49ce2df6) and open another PR for that? Thanks.

tpfau:
done

laurentheirendt:
Great, thanks @tpfau 👍

PR #1406, opened 2018-12-05
Convert all qpng equalities to inequalities -> no row rank restriction

tpfau:
Update to the QPNG interface to avoid the rank deficiency issue with equality constraints.
Changing equality constraints to two equivalent inequality constraints which don't have the restriction. 


tpfau:
hmmm... For some reason qpng does not stick to the constraints set...

PR #1407, opened 2018-12-05
Utility functions

tpfau:
Adding a function that determines the basic model field sizes. 
Allows to avoid lengthy code to determine `nRxns,nMets,nCtrs,nVars` which is often necessary when modifying COBRA Problem structures derived from a model. 


laurentheirendt:
Thanks @tpfau 👍

PR #1412, opened 2018-12-17
Small bug in field reordering

cladelpino:
There seems to be an error in the concatenation of the reordered fields. At least, it affects me when running one model instance with ibm_cplex with allowLoops = false


laurentheirendt:
Thanks @cladelpino 👍

PR #1417, opened 2019-01-15
Bug fixes in Microbiome Modeling Toolbox functions

almut-heinken:
-Changed inputs to the path of the file with abundance data-spreadsheets may not be read properly otherwise.
-Allowed creation of pan-models through database which will be uploaded to cobratoolbox/papers.


laurentheirendt:
Thanks @almut-heinken 👍

PR #1421, opened 2019-02-04
added rxns as input for printFluxBounds

cjos238:
I have only added a "rxns" argument so anyone can only print the flux bounds for those reactions that they need rather than all the reactions. I couldn't find a test for this function, so left it as is. However, will write one if needed.


cjos238:
@laurentheirendt am I missing some steps here?

laurentheirendt:
Thanks @cjos238 👍

PR #1438, opened 2019-03-03
add QFCA

mtefagh:
*Please read the full details in https://mtefagh.github.io/qfca/


laurentheirendt:
@mtefagh , great! 👍 

Eventually, it might be best to either have your code as a submodule in `external`, or to completely integrate the code (documentation included) and then deprecate https://github.com/mtefagh/qfca. 

What do you think?

mtefagh:
Thanks a lot @laurentheirendt for the prompt response! I appreciate it.

Either way is fine with me. Maybe we can have this one as a submodule in `external` as you suggested, and I have another piece of code which I will probably pull tomorrow, and we can integrate that one completely with the full documentation included. The other one's paper is still under review and not published yet, so I can even change the reference in the article to the corresponding *COBRA* documentation webpage. If you want to take a lot in [here](https://mtefagh.github.io/swiftcore/), it is just an even faster version of `fascore` and `fastcc` (**10x** speedup). What do you think?

mtefagh:
@laurentheirendt the latest commit moves QFCA to `external` as per your request above.

laurentheirendt:
@mtefagh, may you have a look? Thanks.
```
20:04:34 The following tests failed:
20:04:34 testQFCA.m
20:04:34 
20:04:34 The reasons were as follows:
20:04:34 ------------------------------------------------
20:04:34 testQFCA.m:
20:04:34 Undefined function 'vecnorm' for input arguments of type 'double'.
20:04:34 Error in blockedReac (line 30)
20:04:34     blocked(blocked == 0 & rev == 1) = vecnorm(Z(rev(blocked == 0) == 1, :), ...
20:04:34 Error in QFCA (line 70)
20:04:34     [S, rev, blocked] = blockedReac(S, rev, solver);
20:04:34 Error in testQFCA (line 31)
20:04:34 [~, ~, fctableD, blockedD] = QFCA(model.S, model.lb < 0, false);
20:04:34 ------------------------------------------------
```

mtefagh:
@laurentheirendt, according to https://www.mathworks.com/help/matlab/ref/vecnorm.html the function `vecnorm` was introduced in R2017b, and I guess the test was running on an older version of Matlab. Can you confirm that?

laurentheirendt:
@mtefagh, may you use a more conventional function to determine the norm of a vector that is backward compatible with versions older than R2018b?

tpfau:
We aim for a compatability for at least the last 3 b versions of Matlab (i.e. the currently oldest supported plattform would be 2016b, so any newer features should only be used if they are absolutely necessary.

mtefagh:
@laurentheirendt and @tpfau, `vecnorm` is replaced. I have also changed the input output format.

laurentheirendt:
@mtefagh, may you have a look at why your test is failing on MATLAB versions older than R2018b?
```
18:42:29 The reasons were as follows:
18:42:29 ------------------------------------------------
18:42:29 testQFCA.m:
18:42:29 Error using all
18:42:29 Dimension argument must be a positive integer scalar within indexing range.
18:42:29 Error in testQFCA (line 34)
18:42:29 assert(all(fctableD == fctable, 'all'));
18:42:29 ------------------------------------------------
```

laurentheirendt:
@mtefagh, actually, you put all the source now in `external/`. It would be better if you really then integrated the code as a submodule properly speaking, or that you move the code under `src`. Would that be possible?

mtefagh:
@laurentheirendt, yes I can move the code to the `src` directory. But apart from that, is there anything else I can do? I don't know of any functions in COBRA which need flux coupling analysis, and hence QFCA can be integrated with them.

PR #1439, opened 2019-03-04
add swiftcore and swiftcc

mtefagh:
Please read the full details in https://mtefagh.github.io/swiftcore/

### DIFF
--- a/src/dataIntegration/transcriptomics/SWIFTCORE/blocked.m
+++ b/src/dataIntegration/transcriptomics/SWIFTCORE/blocked.m
@@ -1,0 +1,86 @@
+function result = blocked(S, rev, solver)
+% blocked finds all the irreversible blocked reactions and is utilized in swiftcc
+%
+% USAGE:
+%
+%    result = blocked(S, rev, solver)
+%   
+% INPUTS:
+%    S:         the associated sparse stoichiometric matrix
+%    rev:       the 0-1 vector with 1's corresponding to the reversible reactions
+%    solver:    the LP solver to be used; the currently available options are
+%               'gurobi', 'linprog', and 'cplex' with the default value of 
+%               'linprog'. It fallbacks to the COBRA LP solver interface if 
+%               another supported solver is called.
+%
+% OUTPUT:
+%    result:    the result returned by the LP solver; among the last n entries, 
+%               all the -1 entries are blocked, and the other entries are zero.
+%               The first m entries are its fictitious metabolite certificate.
+%
+% .. Authors:
+%       - Mojtaba Tefagh, Stephen P. Boyd, 2019, Stanford University
+
+    [m, n] = size(S);
+    irev = m + find(rev == 0);
+    model.obj = zeros(m+n, 1);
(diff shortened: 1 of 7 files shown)


rmtfleming:
It would be good to
a) include swiftcore as an option within the function: createTissueSpecificModel
b) make swiftcore future proof since we are moving to support more general linear inequality constraints, as indicated in: buildLPproblemFromModel
These would increase the amount of users of swiftcore and make it more flexible for more than basic COBRA models.

mtefagh:
Thanks a lot @rmtfleming for the prompt response! I appreciate it.

a) This should be straightforward. I will commit the necessary changes shortly.
b) This one takes a little more time, and maybe I'll do it in the near future but I can't make it before my article is published.

About your first suggestion, I also noted that `fastGapFill` calls both `fastcc` and `fastcore` and these subroutine calls constitute the majority of the runtime. I don't know if your licensing policy allows me to just copy and paste the source code and change those subroutine calls to their swift version and then name it `swiftGapFill`, and of course, I will include the original authors in the header of the file. Is it okay to do so?

rmtfleming:
Regarding gap filling, what is really needed is a generic algorithm independent interface. i.e. gapFill.m that would have multiple options, one to call fastcc and fastcore, another to call swiftcc and swiftcore. That could be done by making copies, so long as there is still that generic interface.

mtefagh:
Okay, to conclude I will do the following by the end of tomorrow:

1. add `swiftcore` to the options of `createTissueSpecificModel`
2. add `swiftcc` as the default method to `findFluxConsistentSubset`
3. add `swiftcore` to the options of fasGapFill and rewrite it as `gapFill`

Please confirm.

tpfau:
One question:

Why are the different versions (gurobi/cplex/linprog) needed?
Essentially they are already implemented in soolveCobraLP which is
called otherwise.
The reason we have this general approach is to keep solver specific code
as concentrated as possible so that it is easier to debug in thefuture
instead of distributing it over multiple different functions.

As for 2: swiftcc/swiftcore is not an exact replacement for
fastcc/fastcore because it does not integrate model boundaries (lb/ub/c
etc and does not allow to use them).
While this makes it likely faster (because the solvers are good in
exploiting unboundedness of individual variables), it will make it
impossible to try to extract models under certain conditions, so I would
not recommend to modify defaults of any algorithm, as results would
likely change, and thus create backward compatability issues.

As for 3:

No, it is more about creating a general gapFill.m file that calls the
other file, similar to createTissueModel for iMAT, GIMME, fastcore etc.

mtefagh:
Thank @tpfau for your comments.

1) Actually, I developed `swiftcore` as a stand-alone package with zero-dependancies, and then one of the reviewers of my article suggested integrating it with COBRA. In my future works, I will probably start developing in COBRA from the beginning.

2) Fine, I have reverted all the default functions to `fastcc` in my latest commit.

3) Okay, if I understand correctly you want both `swiftGapFill.m` and `gapFill.m`. Maybe I will do it in the future, but I skip this one for now.

PR #1455, opened 2019-04-01
Fix columnVector for correctly manipulating empty row and column vectors

shjchan:
`columnVector(zeros(0, 1))` and `columnVector(zeros(1, 0))` both return `1x0 empty row vector`, which is not exactly correct. And this causes, for example, `convertOldStyleModel(modelWithA)` not passing `verifyModel( *, 'simpleCheck', true)` if modelWithA has *.A equal to *.S (though unusual) because `convertOldStyleModel` will add empty row vectors for `*.ctrs`, `*.dsense` and `*.d`.

This pull request makes `columnVector(zeros(0, 1))` and `columnVector(zeros(1, 0))` both return `0x1 empty column vector`. Tests are added accordingly.

### DIFF
--- a/src/base/columnVector.m
+++ b/src/base/columnVector.m
@@ -17,7 +17,7 @@ function vecT = columnVector(vec)
 
 [n, m] = size(vec);
 
-if n < m
+if (m ~= 1 && n < m) || n == 1
     vecT = vec';
 else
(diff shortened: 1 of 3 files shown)


laurentheirendt:
Thanks @shjchan 👍

PR #1460, opened 2019-04-12
Exclude matlab and lp_solve from testLP9

laurentheirendt:
cc: @WegrzynAB


WegrzynAB:
Ah, I forgot about these. Sorry. Will remember for the future! Thank you @laurentheirendt  for taking care of it!

PR #1462, opened 2019-04-22
Fixing errors when creating missing fields

akaviaLab:
Trying to add reactions to a model that has genes, but no gene rules leads to errors (invalid models), as did adding exchange reactions to a model (problem with subSystem definition). I've corrected the error and modified the test to fail without my corrections.


akaviaLab:
I've been looking into why testExtractConditionSpecificModel fails
- The reference model has Exchange reactions in a different manner {'Exchange'} than the result {{'Exchange'}}. I've fixed that by changing the reference model (will be a separate commit)
- isSameCobraModel does not check the subSystems, because they are cells within cells. Should this be opened as a separate bug?

laurentheirendt:
@akaviaLab, is there any way that you can fix that?

akaviaLab:
Hi Laurent,
I've added a different version of the reference model, and the tests pass.

As for isSameCobraModel not checking subSystems, should that be a separate issue? Should it even be fixed (a lot of scripts are using isequal, which seems to work)?
One way of fixing it is changing subSystems not to be cells within cell, but I'm not sure why that is even the case. Is this so reactions can have more than one subSystem?

laurentheirendt:
Thanks @akaviaLab 👍

tpfau:
> 
> 
> Hi Laurent,
> I've added a different version of the reference model, and the tests pass.
> 
> As for isSameCobraModel not checking subSystems, should that be a separate issue? Should it even be fixed (a lot of scripts are using isequal, which seems to work)?
> One way of fixing it is changing subSystems not to be cells within cell, but I'm not sure why that is even the case. Is this so reactions can have more than one subSystem?

That's exactly it.
By convention, e.g. in SBML subsystems are represented by the groups extension. And thus a reaction can be part of multiple subsystem "groups".
It also makes a lot of sense, as e.g. we currently have things like "Glycolysis/Gluconeogenesis" as a single pathway, while this are in fact different pathways which have multiple reactions in common, but also have reactions which separate them (something thats not well represented by current models).

PR #1465, opened 2019-04-25
Adding support for table fields in the model struct

tpfau:
Currently the dynamic field adaption does not handle table fields since they can't properly be modified with `subsasgn`. This PR adds support for table fields.


laurentheirendt:
Great! Thanks!

Btw, your commit email is not recognized by Github.

PR #1466, opened 2019-04-30
Update verify rules

tpfau:
Updating VerifyModel to:
1. Have better verbosity when testing model fields for verbosity
2. Explicity testing rule syntax (i.e. whether rules are actually evaluateable)


src/reconstruction/modelGeneration/modelVerification/verifyRuleSyntax.m
```diff
@@ -0,0 +1,44 @@
...
+%    ruleString:    The string in rule format (i.e. x(1) | x(2) & x(3) ...)
+%um
+% OUTPUT:
+%    tf:            Whether the given string is evaluateable (i.e. the
+%                   syntax is correct
+persistent x
+
+if isempty(x)
+    % 10000 genes should be large enough for most models.
+    x = true(10000,1);
```
laurentheirendt: Would you be able to base this number on the model size?

src/reconstruction/modelGeneration/modelVerification/verifyRuleSyntax.m
```diff
@@ -0,0 +1,44 @@
+function tf = verifyRuleSyntax(ruleString)
+% try to verify if a given rule string is syntactically correct
```
laurentheirendt: This header needs proper formatting.

src/reconstruction/modelGeneration/modelVerification/verifyRuleSyntax.m
```diff
@@ -0,0 +1,44 @@
...
+            x = true(max(vals),1);
+            temp = eval(ruleString);
+            % if we reach this point it is fine
+            tf = true;
+        catch ME
+            tf = false;
+        end
+    end
+end
+
```
laurentheirendt: There are many empty lines here.

src/reconstruction/modelGeneration/modelVerification/verifyRuleSyntax.m
```diff
@@ -0,0 +1,44 @@
...
+    x = true(10000,1); 
+end
+
+if isempty(ruleString)
+    tf = true;
+else
+    try
+        temp = eval(ruleString);
+        % if we reach this point it is fine
+        tf = true;
```
laurentheirendt: `tf = true` could be set as default, no?

src/reconstruction/modelGeneration/modelVerification/verifyRuleSyntax.m
tpfau: I would have to pass a model into the function, and the function as such does not need one. So I would rather not do so.

src/reconstruction/modelGeneration/modelVerification/verifyRuleSyntax.m
tpfau: No, rather false.
But I just noticed, that this is still bugged. Need to fix the second try/catch block..

laurentheirendt: Yeah, doesn't matter 😄

tpfau:
I'm not quite sure, what to do about the invalid rule in Recon2.04.mat.
There is a rule (2173) which has ' ... | ()' which is definitely invalid, and can't be evaluated, and this is also present in grRules, and thus can't be fixed by `generateRules`
@rmtfleming would it be possible to fix this rule in the Recon2.04.mat file on vmh, as thats the place this model is obtained from?

tpfau:
Ok, I updated the convertOldStyle model function to explcicitly correct that one particular rule... This is ugly, but since there is no response, I don't see how else to make these updated work

tpfau:
Lets retry this, as it works on my local machine and on different CI machines

tpfau:
@laurentheirendt Any idea why the windows CI can't process the checkout?

tpfau:
restarting again

tpfau:
another restart. lets see if the win7 machine is broken or can recover.

tpfau:
Any further changes needed or do I miss something?

tpfau:
@laurentheirendt :
Can this be merged?

PR #1467, opened 2019-05-02
Added support in sampler for C,d inequality constraints

Bounciness:
Added support in the CHRR sampler for inequality constraints that are present in the COBRA model fields model.C and model.d. The sense of the inequalities should be in the field model.dsense. If either of the fields model.C or model.d are missing, the behavior of the new function is equivalent to the current function in the COBRA toolbox.


rmtfleming:
Thanks Ben.

laurentheirendt:
Thanks @Bounciness 👍

rmtfleming:
Hi Ben,
this function is the one we use to make an LP problem from a general model structure.
https://github.com/opencobra/cobratoolbox/blob/51e487bfcb959950e6212b6dfc4f8bf1ed3f81c5/src/base/solvers/buildLPproblemFromModel.m
Consider to use that then take the output to form a sampling problem. It will be more future proof.

PR #1472, opened 2019-05-13
Modifications of MATLAB version determination

laurentheirendt:
Modification as suggested by Marco Severgnini:

https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/cobra-toolbox/uMSv27Lfi4U

### DIFF
--- a/src/base/install/getFilesInDir.m
+++ b/src/base/install/getFilesInDir.m
@@ -140,7 +140,7 @@ if ~parser.Results.checkSubFolders
     files = files(cellfun(@(x) isempty(regexp(x,regexptranslate('escape',filesep))),files));
 end
 
-files = strcat(absPath, filesep, files);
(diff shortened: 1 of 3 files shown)


src/base/install/getFilesInDir.m
```diff
@@ -140,7 +140,7 @@
     files = files(cellfun(@(x) isempty(regexp(x,regexptranslate('escape',filesep))),files));
 end
 
-files = strcat(absPath, filesep, files);
+files = strcat(repmat({absPath}, 1, length(files)), repmat({filesep}, 1, length(files)),files);
```
tpfau: In which matlab version does this fail?
Just out of curiosity

laurentheirendt: This came from the forum. As this is equivalent, I though of adding it.

PR #1475, opened 2019-05-16
adding old version of testFVA

FaridZare17:
Adding the older version of testFVA.

This function is not written or modified by me.


laurentheirendt:
@FaridZare17, may you integrate the missing functionality (if there is any) into the existing test? Thanks.

FaridZare17:
@laurentheirendt 
Since it's the older version,  it doesn't have any further functionality.
There is only further simplicity in some lines.
Thanks for your response,
Farid

PR #1481, opened 2019-05-31
HotFix: Adapt MIRIAM to new registry.org API

tpfau:
The registry.org API is changing due to a change in their setup. This means we have to adapt the automated fetch script to obtain the specifications for registry.org databases which are not explicitly specified in the Toolbox. 
This PR adapts the respective function.


laurentheirendt:
Thanks @tpfau 👍

PR #1482, opened 2019-05-31
Adding removeZeroFluxRxns Function

FaridZare17:
Adding removeZeroFluxRxns function to the develop brunch.

This functions finds maximum and minimum flux for each reaction that can be carried out through the
network, and removes those reactions that are incapable of having any fluxes due to that functional circumstances.

### DIFF
--- a/src/analysis/deletion/removeZeroFluxRxns.m
+++ b/src/analysis/deletion/removeZeroFluxRxns.m
@@ -10,32 +10,23 @@ function modelOut = removeZeroFluxRxns(model, rxns)
 %    new_Model = removeZeroFluxRxns(model, reactions)
 %
 % INPUTS:
-%    model:                 COBRA model structure (must define `.rxns`, `.lb`, `.ub`)
+%    model:                      COBRA model structure (must define `.c`, `.lb`, `.ub`, `.S`, `.rxns`, `.mets`)
 %
 %OPTIONAL INPUT
-%    reactions:             Cell array of reaction names in the model
+%    reactions:                Cell array of reaction names in the model
 %
 % OUTPUT:
 %    new_model:             Curated model, the model with removed reactions
-%                           incapable of having a flux value, and corrected
-%                           upper and lower bounds.
+%                                     incapable of having a flux value, and corrected
+%                                     upper and lower bounds.
 %
 % .. Author: - Farid Zare  5/31/2019
+%...Further modifications suggested by: -Thomas Pfau 6/1/2019
 
-if ~isstruct(model)
-    error('Input model must be structure');
-end
+checkModel = verifyModel(model,'FBAOnly',true,'simpleCheck', true);
 
-if ~isfield(model, 'rxns')
-    error('Input model should contain `.rxns` field');
-end
-
-if ~isfield(model, 'lb')
-    error('Input model should contain `.lb` field');
-end
-
-if ~isfield(model, 'ub')
-    error('Input model should contain `.ub` field');
+if ~checkModel
+    error('The input model is invalid. Please check the model with ''verifyModel(model)'' and correct all indicated errors');
 end
 
 if nargin < 2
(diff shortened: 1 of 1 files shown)


src/analysis/deletion/removeZeroFluxRxns.m
```diff
@@ -0,0 +1,76 @@
...
+%    reactions:             Cell array of reaction names in the model
+%
+% OUTPUT:
+%    new_model:             Curated model, the model with removed reactions
+%                           incapable of having a flux value, and corrected
+%                           upper and lower bounds.
+%
+% .. Author: - Farid Zare  5/31/2019
+
+if ~isstruct(model)
```
tpfau: If you test for some properties, why not for the existence of `model.S` which is necessary t run optimizeCbModel.
Alternatively, you could use `verifyModel(model,'FBAOnly',true)` to check all fields that are relevant for FBA calculations. The result is a struct that is empty if there are no errors and contains error information if there are any errors. Admittedly, a printout function for the error struct is currently still missing, but we can add that if you need it. If you just want to check a yes/no you could call: `verifyModel(model,'FBAOnly',true,'simpleCheck', true)` and if this is false e.g. exist with an error like:
"The input model is invalid. Please check the model with verifyModel(model) and correct all indicated errors"

src/analysis/deletion/removeZeroFluxRxns.m
```diff
@@ -0,0 +1,76 @@
...
+    solMin = optimizeCbModel(model, 'min');
+    solMax = optimizeCbModel(model, 'max');
+    
+    %Set bounds to their feasible values in the network
+    rxnID = findRxnIDs(model, rxns(i) );
+    modelOut.lb(rxnID) = solMin.f;
+    modelOut.ub(rxnID) = solMax.f;
+    
+end
+
```
tpfau: You could simply run:
```[mins,maxs] = fluxVarability(model);
modelOut.lb = mins;
modelOut.ub = maxs;
```

src/analysis/deletion/removeZeroFluxRxns.m
```diff
@@ -0,0 +1,76 @@
...
+index = modelOut.ub ==0 & modelOut.lb == 0;
```
tpfau: I would suggest a feasibility threshold instead of direct comparison to 0.
(e.g.
```
feasTol = getCobraSolverParams('LP','feasTol')
index = abs(modelOut.ub) < feasTol & abs(modelOut.lb) < feasTol
 ```

src/analysis/deletion/removeZeroFluxRxns.m
FaridZare17: Dear  @tpfau  
Thank you so much for your valuable advice, I'm going through this comments and apply the changes !
I'm going to use [min, max] = fluxVariability(model , 0) to get the same result, and so there would be no for loop and I also have to terminate the waiting bar.
how about adding a waiting bar to fluxVariability itself? since it is usually time consuming, having an estimation of runtime would be helpful.

tpfau: There is a progress bar for some instances of fluxVariability. However its currently not implemented for parallel fluxVariability (which is the default mode, as this is not as trivial as it sounds since you need an external monitor. Maybe it will be implemented at a later stage.

rmtfleming:
This functionality needs to be integrated as an option to use fva via the
generic interface for this sort of function: findFluxConsistentSubset

See
in
https://opencobra.github.io/cobratoolbox/stable/modules/reconstruction/modelGeneration/fluxConsistency/index.html

laurentheirendt:
@FaridZare17, as this is a new function, it also needs to be tested. Guidelines are [here](https://opencobra.github.io/cobratoolbox/stable/contributing.html#guide-for-writing-a-test).

FaridZare17:
Dear @rmtfleming, Dear @laurentheirendt 
Thank you very much for your responses, I'm going to write the test, and try to add this functions as an option to findFluxConsistentSubset function.

FaridZare17:
Hi,
sorry for the delay,
Unfortunately I've faced a problem. When I run FVA with following options
[minFlux, maxFlux] = fluxVariability(model, 0, 'max', rxns)
I get different answers when I enter different reaction name list, 
I need to know the maximum and minimum flux capacity for each reaction, and to this aim I had used FBA while setting each reaction as objective and minimized and maximized the flux, the I replaced the FBA with FVA according to the comments, and now I think the replaced lines have different functionality.
May I back to use FBA again in the code ?

rmtfleming:
Test with the new FVA function, then reopen if necessary.

PR #1496, opened 2019-08-04
Making warning of missing genes more explicit

tpfau:
Updating findRxnsFromGenes making the warning message when searching for missing genes more explicit, as it is often unclear, which genes where not found in the model.


laurentheirendt:
Thanks @tpfau 👍

PR #1499, opened 2019-08-17
Allowing https to work with cobrapy 015 and up

akaviaLab:
Cobrapy outputs identifiers using https:// and not http://
Modified the regexp to make it work. Also checked that identifiers.org accepts both http and https connections.


laurentheirendt:
Thanks @akaviaLab 👍

PR #1502, opened 2019-09-12
HH to new model format support

rmtfleming:
Support for draft HH model (https://www.biorxiv.org/content/10.1101/255885v1) conversion into new format. Improvements to conversion and verification files.

- [x ] Tested my code on my own machine


tpfau:
This PR was never tested and completely breaks SBML IO. 
@rmtfleming could you please revert these changes and have them pass the tests before resubmitting them? At least one issue is in preparseGPR which no longer works with individual grRule strings, which is required for GPR parsing in SBML files.

src/base/io/utilities/convertOldCouplingFormat.m
```diff
@@ -0,0 +1,135 @@
...
+
+if isfield(model,'A') && isfield(model,'S')
+%     if printLevel >=1
+%         warning('The inserted Model contains an old style coupling matrix (A). The MAtrix will be converted into a Coupling Matrix (C) and fields will be adapted.')
+%     end
+    [nMets,nRxns] = size(model.S);
+    if size(model.S,1) == size(model.A,1)
+        bool=strncmp('slack_',model.mets,length('slack_'));
+        if any(bool)
+            %this is a draft harvey or harvetta
```
tpfau: Can you provide me with a model of this type? 
Because at the moment, this code breaks a few things and I would like to correct it in a way that works for both these draft models and other models.

PR #1504, opened 2019-09-18
Test ARTENOLIS trigger

laurentheirendt:


laurentheirendt:
@Bounciness, all tests are failing (actually, the initialisation of the COBRA Toolbox fails), as https://github.com/Bounciness/constrained-logconcave-sampler has been set as a dependency in https://github.com/opencobra/cobratoolbox/pull/1502.

Do you mind making this repository public? If not, we need to remove the dependency.

rmtfleming:
Pls remove the dependencey on constrained-logconcave-sampler for the moment.

PR #1511, opened 2019-10-23
Enabled cycle-free pan-models up to family level

almut-heinken:
- Error message was added for wrong input of taxonomic level
- Enabled creation of futile cycle-free pan-models up to family level in AGORA 1.03
- removed creation of gene rules in pan models (unnecessary)


almut-heinken:
Note: a few AGORA reconstructions have syntax errors in the rules field and currently cause the readCbModel function to fail. I will put a new version of AGORA 1.03 on vmh.life to address this.

PR #1513, opened 2019-10-27
transcriptomics data integration: use gene significance along with expression value

chaitrasarathy:
Hi,
This pull request is for a small modification in the functions for integrating with transcriptomics data. The inputs for the functions - mapExpressionToReactions, findUsedGenesLevels, selectGeneFromGPR have been modified to also include significance level along with an expression value for each gene.

Best regards,
Chaitra

### DIFF
--- a/src/dataIntegration/transcriptomics/preprocessing/findUsedGenesLevels.m
+++ b/src/dataIntegration/transcriptomics/preprocessing/findUsedGenesLevels.m
@@ -1,9 +1,11 @@
-function [gene_id, gene_expr, gene_sig] = findUsedGenesLevels(model, exprData, printLevel)
+function [gene_id, gene_expr, gene_sig] = findUsedGenesLevels(model, exprData, exprSig, printLevel)
 % Returns vectors of gene identifiers and corresponding gene expression
 % levels for each gene present in the model ('model.genes').
 %
 % USAGE:
+%
 %    [gene_id, gene_expr] = findUsedGenesLevels(model, exprData)
+%    [gene_id, gene_expr, gene_sig] = findUsedGenesLevels(model, exprData, exprSig, printLevel)
 %
 % INPUTS:
 %
@@ -13,9 +15,11 @@ function [gene_id, gene_expr, gene_sig] = findUsedGenesLevels(model, exprData, p
 %       .gene                cell array containing GeneIDs in the same
 %                            format as model.genes
 %       .value               Vector containing corresponding expression value (FPKM)
-%       .sig                 Vector containing corresponding significance values
+
 %
 % OPTIONAL INPUTS:
+%    exprSig:            Vector containing significance values of
+%                        expression corresponding to expression values in exprData.value (ex. p-values)
 %    printLevel:         Printlevel for output (default 0);
 %
 % OUTPUTS:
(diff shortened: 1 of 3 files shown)


src/dataIntegration/transcriptomics/preprocessing/selectGeneFromGPR.m
```diff
@@ -1,4 +1,4 @@
-function [expressionCol, gene_used] = selectGeneFromGPR(model, gene_names, gene_exp, parsedGPR, minSum)
+function expressionRxns = selectGeneFromGPR(model, gene_names, gene_exp, gene_sig, parsedGPR, minSum)
```
tpfau: Please refactor this so that gene_sig is an optional argument, or implement it thus that the old signature can still be used.
Since the number of optional arguments starts to increase, I would suggest to modify the function to use parameter/value pairs while still accepting the original signature. For a potential way to do so, have a look at e.g. addReaction. 
Modifying the header of the function in the way you currently suggest would potentially break a lot of existing code. 

Also the current outputs should be retained. Also admittedly, they are currently lacking documentation.

src/dataIntegration/transcriptomics/preprocessing/selectGeneFromGPR.m
```diff
@@ -19,55 +19,55 @@
...
-gene_used={};
-for i=1:length(model.rxns)
-	gene_used{i}='';
-end
-    
 % -1 means unknown/no data
-expressionCol = -1*ones(length(model.rxns),1); 
+expressionRxns.rxnSig = zeros(length(model.rxns),1); 
+expressionRxns.rxnExp = -1*ones(length(model.rxns),1); 
 for i = 1:length(model.rxns)
```
tpfau: Please modify the code so that it allows using significance as an optional input.

chaitrasarathy: Thank you @tpfau for your comments. It is a good point to keep the new variables optional and not break existing code. I will modify the function to take significance value as optional

chaitrasarathy:
In the latest commit, I have updated all the three functions - mapExpressionToReactions, findUsedGenesLevels and selectGeneFromGPR to take significance value as optional input. Old signature can still be used.

src/dataIntegration/transcriptomics/preprocessing/findUsedGenesLevels.m
```diff
@@ -1,9 +1,11 @@
-function [gene_id, gene_expr] = findUsedGenesLevels(model, exprData, printLevel)
+function [gene_id, gene_expr, gene_sig] = findUsedGenesLevels(model, exprData, exprSig, printLevel)
```
tpfau: `exprSig` should go to the end of the list (again, backwards compatability).
Alternatively (and I think that's the better option):
Include `exprSig` as an optional field in `exprData` as an optional field. test whether the field is present and if, use the information from the field. if not, just don't use it.
That would make the function:
a) backward compatible
b) more flexible

chaitrasarathy: in the latest commit, significance value is an optional input in the struct `exprData`. there is also a check to see if this field exists, if yes this input is used. hope this retains the previous usage of these functions

rmtfleming:
@tpfau 
For some of the tests, there are errors like this:

21:59:37 testGetModelSizes.m:
21:59:37 Cell contents reference from a non-cell array object.
21:59:37 Error in preparseGPR (line 20)

It is hard to see why that would be platform specific. Can you see why?

PR #1514, opened 2019-10-30
preparseGPR crashes if the input is a single gene rule

almut-heinken:
The preparseGPR function caused data2model to crash since it was given a single gene rule as input but only worked with the entire grRules field. I fixed this. Hopefully, this does not introduce any other problems.


laurentheirendt:
> Incidently, the CI currently does not start any checks, so a passing CI does unfortunately not mean, that the PR does indeed pass the tests.

This has been fixed in #1515.

PR #1427, opened 2019-02-20
Patch 1

axelvonkamp:
Basic implementation of multi-threaded FVA via an external JAVA virtual machnine with CPLEX as solver.

**I hereby confirm that I have:**

- [ X] Tested my code on my own machine
- [X ] Followed the guidelines in the [Contributing Guide](https://opencobra.github.io/cobratoolbox/docs/contributing.html)
- [X ] Selected `develop` as a target branch (top left drop-down menu)

*(Note: You may replace [ ] with [X] to check the box)*

PR #1538, opened 2019-12-21
Add back mtFVA option in fluxVariability

shjchan:
@rmtfleming 
Re-solve the previous conflict and re-introduce the option for `mtFVA` in `fluxVariability`
#1534 #1427 
Updated `testMtFVA`

@axelvonkamp, please also take a look. In particular I changed one character in `mtFVA.m` for adding classpath in windows machine. Is that correct? That's how I could run it on a window machine.


axelvonkamp:
> @axelvonkamp, please also take a look. In particular I changed one character in `mtFVA.m` for adding classpath in windows machine. Is that correct? That's how I could run it on a window machine.
> 
Yes, I think that is correct, because on Windows the classpath separator is a ";" instead of a ":", thanks for fixing this.

PR #1541, opened 2020-01-13
Fix lrs output read ray

htpusa:
A new attempt since the bug remains, this time via devTools. If the input file of lrsOutputReadRay is an empty file, the function gets stuck in an infinite loop. Personally I have never been able to run testAll successfully because of this. A quick fix where an empty input file throws an error.


rmtfleming:
Thanks.

PR #1590, opened 2020-04-15
feat: gene SBO terms in XML output

BenjaSanchez:
hi! I'm working in increasing the compatibility of models exported with cobrapy/cobratoolbox so that no fields are lost when going from one to the other one. I noticed that cobratoolbox does not record SBO terms for genes by default (which cobrapy does), so here I'm adding that setting, as afaik the gene field is solely intended to store that type of species, so there should not be a need to ask the user to do it. Let me know if this makes sense to you.

http://www.ebi.ac.uk/sbo/main/SBO:0000243


rmtfleming:
CI being migrated at the moment. Watch for bugs when it is back up.

PR #1599, opened 2020-06-09
 Changes to support the matlab NLP solver

sowmyamanojna:


initCobraToolbox.m
```diff
@@ -336,7 +336,7 @@ function initCobraToolbox(updateToolbox)
 SOLVERS.pdco.type = {'LP', 'QP'};
 SOLVERS.quadMinos.type = {'LP'};
 SOLVERS.dqqMinos.type = {'LP','QP'};
-SOLVERS.matlab.type = {'LP', 'NLP'};
+SOLVERS.matlab.type = {'NLP'};
```
tpfau, line 339: Matlab is a valid LP solver. Maybe 2020a+ versions need some adaption here, but removing it from the solversa can't be a solution.

sowmyamanojna: I agree that removing the solver may not be a solution. It was a quick fix I had used to incorporate the `NLP` solver. 

I had tried finding a solution to the `matlab` LP problem and fixed all the calls made to `optimoptions` in `solveCobraLP`, however, the error still stems from a call to `optimoptions` inside the inbuilt `linprog` function (Line: 292). As `linprog` is a READ-ONLY file in the Optimization Toolbox, the changes made there cannot be reflected back to the files in the Cobra Toolbox. 

Thanking you,
Sowmya Manojna

rmtfleming: Did you find a workaround?

sowmyamanojna: I replaced the `optimoptions.p` file with the `optimoptions.m` file from my online `MATLAB` account. This fixed the LP and NLP problem. No additional changes were required.

PR #1631, opened 2020-10-05
feat: SBO terms for reaction bounds

BenjaSanchez:
Added SBO terms for all lower/upper bounds, based on the definitions from SBO:

* Measured/estimated bounds: http://www.ebi.ac.uk/sbo/main/SBO:0000625
* Default bounds (`-Inf`, `-1000`, `0`, `+1000` or `+Inf`): http://www.ebi.ac.uk/sbo/main/SBO:0000626

These definitions are consistent with cobrapy's after merge of https://github.com/opencobra/cobrapy/pull/949

Tested here: https://github.com/BenjaSanchez/yeast-GEM/commit/086b1cb58a936efe745009b0431ff78e77b0a4be

### DIFF
--- a/src/base/io/utilities/writeSBML.m
+++ b/src/base/io/utilities/writeSBML.m
@@ -387,6 +387,11 @@ if ~isempty(listUniqueValues)
         
         fbc_parameter.id=listUniqueNames{i,1};
         fbc_parameter.value=listUniqueValues(i);
+        if fbc_parameter.value == -1000 || fbc_parameter.value == +1000 || fbc_parameter.value == 0 || isinf(fbc_parameter.value)
+            fbc_parameter.sboTerm=626;
+        else
+            fbc_parameter.sboTerm=625;
+        end
         if i==1
             sbmlModel.parameter=fbc_parameter;
         else
--- a/test/verifiedTests/base/testIO/testWriteSBML.m
+++ b/test/verifiedTests/base/testIO/testWriteSBML.m
@@ -38,6 +38,19 @@ testModelXML = rmfield(testModelXML,'rxnReferences');
 % assess any potential differences
 assert(~any(numDiff))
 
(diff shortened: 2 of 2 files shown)


rmtfleming:
Thanks for this contribution Ben. Please can you add a feature to the associated test function, or create a new one associated with this new feature:
https://github.com/opencobra/cobratoolbox/tree/master/test/verifiedTests/base/testIO

BenjaSanchez:
@rmtfleming I have now added the corresponding test in `testWriteSBML.m`. Note that as far as I can tell, nowhere in the code is the SBML structure itself (the output from `writeSBML.m`) tested; you might consider in the future testing all other fields of said structure (`compartment`, `species`, `reaction`, `fbc_geneProduct`, etc).

tpfau:
@BenjaSanchez how would you test the structure? using validateSBML?
We know that IO does not necessarily produce the exact same output model/xml file, but in testWriteSBML we do check whether the model is essentially the same (with the exception of a few fields which are different as commented in the test.
Implicitly this also tests whether the model can be read by libsbml, as otherwise we would not be able to read it back in. But yes, an explicit test would probably be good.

rmtfleming:
cobratoolbox/external/base/utilities/cellstructeq/structeq.m might be useful

BenjaSanchez:
@tpfau I imagine similar checks to the one I introduced here could be performed with all other fields in the SBML structure (i.e. check they have the proper SBO terms). However, maybe checking that the structure doesn't change + remains SBML-compatible is already enough; I only added the test in this PR after @rmtfleming's request, and noticed that there's not really an equivalent to it in the rest of the test suite, so the test looks a bit out of place at the moment.

@rmtfleming not sure how `structeq.m` would help here, as `testWriteSBML.m` already tests if loaded & written models are the same.

PR #1726, opened 2021-03-04
minor edits to findRxnFromGenes

ithiele:

### DIFF
--- a/src/analysis/exploration/findRxnsFromGenes.m
+++ b/src/analysis/exploration/findRxnsFromGenes.m
@@ -25,6 +25,7 @@ function [results, ListResults] = findRxnsFromGenes(model, genes, numericFlag, L
 %       - edited 04/05/09 (MUCH faster now -- NL)
 %       - edited 06/11/10 (yet even faster now -- NL)
 %       - Ronan, Ines - edited interface for backward compatibility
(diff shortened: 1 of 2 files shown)


tpfau (changes requested):
AddMultipleRxns intentionally throws an error here as otherwise you can end up with duplicate reaction ids in the model making the model infeasible.
If an ID already exists you should not be allowed to add it again.

rmtfleming:
@tpfau please can you edit AddMultipleRxns so it does not add any duplicates and informs the end user if there is a duplicate, but still continue without an error?
Closing pull request as changes to findRxnFromGenes in new PR

tpfau:
That is something that can easily be checked by the user prior to adding reactions and removing them. 
Since the experience is that warnings are commonly ignored I would advise against such a change as it is likely to lead to a situation where someone writes a script adding stuff and then is puzzled because a reaction is not what he expects it to be since it was already present with different properties. 
If you really want it like that it should be something like what you did for the else case and looping over varargin removing the corresponding entries (if the variable name is not 'genes').
The fields "should" all be one-dimensional (if I'm not mistaken) so it should really be a simple loop.
Unfortunately, I don't have a working Matlab copy any more and thus cannot test anything I code for bugs, so I don't think it's sensible for me to do this.

PR #1776, opened 2021-05-22
Proposal buildFluxDistLayout with a color gradient

nmendozam:
This is a proposal to use color gradient when building a layout to plot a flux distribution in ReconMap, instead of the current approach that is based on size of the lines.

**Advantages**
- It is easier to understand
- It could represent negative fluxes as well (Not implemented yet)

**Example**
imagen

This is still a work in progress. What do you think, should I continue?


rmtfleming:
Please can you add a param.method option that defaults to 'color' but also provides a 'width' option for those that prefer that?

nmendozam:
The way to use this is by running the code below for gradients of two or three colors, and if the input is only one color, it works as before.
```
serverResponse = buildFluxDistLayout(minerva, model, solution, 'gradiet of two colors', ["#ff0000", "#6617B5"])
serverResponse = buildFluxDistLayout(minerva, model, solution, 'gradient of three colors', ["#ff0000", "#6617B5", "#0000ff"])
```
So if there is no other comment, this is ready to merge.

PR #1882, opened 2021-11-24
minor changes to functions eg converting errors to warnings

ithiele:


tpfau:
This is breaking the assertion, that Identifiers in rxns are unique,

rmtfleming:
We need to make a distinction between a reconstruction and a model.

tpfau:
Many cobra functions implicitly assume that identifiers are unique. A lot of analysis tools might simply produce wrong results without failing if there are multiple rxns/mets with the same name. 
In the end, it's up to you what you want, but I think this approach is likely to cause a lot of trouble in the future.

PR #1995, opened 2022-05-31
fix the bug 'metaboAnnotator'

rosieluo2021:
fix the bug on:
1. generating VMH ID
2. missing fields
3. retrievePotHitsHMDB


rosieluo2021:
fix the non-breaking space error

rmtfleming:
Already incorporated via separate approach

PR #2224, opened 2024-02-10
better dealing with models that have stat3 when performing QP

FaizLAB:


rmtfleming:
maxDeviation 
param.rounding 

need io commenting as per style guide:
https://opencobra.github.io/cobratoolbox/stable/contributing.html#style-guide

Note that checking for optimality is better to be done at a lower level because it applies to more problems:
https://github.com/opencobra/cobratoolbox/blob/master/src/base/solvers/solveCobraLP.m#L1598
https://github.com/opencobra/cobratoolbox/blob/master/src/base/solvers/solveCobraQP.m#L1030

Re-running solveCobraQP or solveCobraLP on a rescaled model is better done in optimiseCbModel for similar reasons.

PR #2266, opened 2024-06-25
Bugfix printRxnFormula

bramnap:
printRxnFormula had an update for duplicated rxnIDs. The update however made it so that the iteration became the index being extracted from the S-matrix instead of the index belonging the the reaction. I.e., If you wanted to print a formula of a 1 reaction, the function would always print the formula of the first reaction, regardless of the input.
Changed the code to find index of reactions and put an error in to stop the code when duplicated reaction IDs are found. The code also tells the user which rxnabbrevation caused the error.

### DIFF
--- a/src/analysis/exploration/printRxnFormula.m
+++ b/src/analysis/exploration/printRxnFormula.m
@@ -132,19 +132,22 @@ formulas = cell(size(rxnAbbrList));
 for i = 1:length(rxnAbbrList)
 
     rxnAbbr = rxnAbbrList{i};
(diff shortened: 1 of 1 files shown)


farid-zare:
Hi Bram,
Could you please use "ismember" or "findRxnIDs" other than "contains" function? This will prevent matching reactions that share part of their name with each other.
Thanks!

bramnap:
Hi Farid,

Done :)

farid-zare:
Fantastic, 
Thanks a million!

PR #2361, opened 2024-11-02
Edit icongems file

ThummaratPaklao:
The changes to the icongems.m file are as follows:

- Added an additional input for the human recon model.
- Resolved division-by-zero issues.


ThummaratPaklao:
I have revised the code to ensure consistency with MATLAB testing requirements.

PR #2448, opened 2025-02-25
Hotfix for seqc_mama primary bash script

TheWileyB:
Added check to prevent use of more cpus than available on the system.
Fix of header suffix removal for all columns in taxa tables.


TheWileyB:
Combining changes with broader addition of Persephone update

PR #2571, opened 2025-10-23
Resolving QPold method of optimizeWBModel

faesslerd:
I resolved small errors of the "QPold" option in optimizeWBModel, and added  case 'QPold' to the function buildOptProblemFromModel


rmtfleming (changes requested):
What is the justification for changing the parameters for how 'ibm_cplex' solves WBMs?
e.g.         param.scaind = -1; was put in there for a purpose.
Most WBMs are on the borderline being too multiscale for a double precision solver so the default is to be careful, perhaps slower, to identify a truly optimal solution, than faster but perhaps only approximate the optimal solution. 
e.g. test https://opencobra.github.io/cobratoolbox/stable/tutorials/tutorial_numCharactWBM.html with your model

faesslerd:
I changed solverName to solverNameQP in case QPold as solverName is already defined as character variable in line 145 and this would then result in an error: "Unable to perform assignment because brace indexing is not supported for variables of this type.".

rmtfleming:
Farid can this be merged now?

PR #2642, opened 2026-04-09
Feature/gecko relax fba

PkiwiBird:
*- expand relaxedFBA to deal with extra variables and constraints
- add test for toy gecko model with enzyme usage*


.github/workflows/testAllCI_step1.yml
farid-zare, line 1: Hi Tania,
Why is this workflow deleted? 
This is the COBRA Toolbox CI workflow and I think it has been deleted by accident.

PR #2643, opened 2026-04-11
replace inf bounds before setting param.maxUB to avoid unbounded solution

PkiwiBird:
*replace inf bounds before setting param.maxUB to avoid unbounded solution*

- [x ] Selected `develop` as a target branch (top left drop-down menu)

### DIFF
--- a/src/analysis/relaxedFBA/relaxedFBA.m
+++ b/src/analysis/relaxedFBA/relaxedFBA.m
@@ -451,8 +451,22 @@ if hasC || hasD
     model = rmfield(model, {'ctrs', 'dsense', 'd', 'C', 'D'});
 end
 
-% % bounds, .toBeUunblockedReactions default
-[nMets,nRxns] = size(model.S);
+% % deal with infinite bounds
+%TODO properly incorporate inf bounds
+if any(model.lb==inf)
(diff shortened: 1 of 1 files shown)


rmtfleming:
infeasibility \ne unbounded

PkiwiBird:
This PR is correcting a mistake I made in the previous PR that was already merged. I originally wrote:
```
if ~isfield(param,'maxUB')
   param.maxUB = max(max(model.ub),-min(model.lb));
   param.minLB = min(-max(model.ub),min(model.lb));
end
```
before:
```
infLB = (model.lb == -inf);
infUB = (model.ub ==  inf);

%add a large finite lower bound here
model.lb(infLB) = -1/feasTol;
%add a large finite upper bound here
model.ub(infUB) = 1/feasTol;
```
Since the lifted gecko model had some inf in ub, param.maxUB became Inf then propagated into relaxFBA_cappedL1 as the cap for the relaxation slacks (p, q, r), so the slacks become unbounded. that's why now I changed the order and put the 2nd block first.

The PR also now corrects another issue: models without D were failing at rmfield(model, {'ctrs','dsense','d','C','D'}). Now C/D are only removed when they exist

PR #2651, opened 2026-04-22
fix(writePajekNet): directed Pajek output, tolerance filter, robust filtering & checks

marouenbg:
## Summary

Audit of `writePajekNet` surfaced five issues. This PR fixes all of them and rewrites the regression test to assert semantics rather than byte-equality with a frozen fixture.

## Fixes

1. **Directed output (`*Arcs` instead of `*Edges`)** — the function already encodes direction (forward/reverse based on flux sign), but the previous header marked the file as undirected, so downstream tools could discard the directionality.
2. **Tolerance-based active-reaction filtering** — replace `FBA.full(i) == 0` with `abs(v(i)) < fluxTol`. `fluxTol` defaults to the LP solver's `feasTol` (via `getCobraSolverParams`) and can be overridden via the new optional argument. Prevents tiny solver residuals (e.g. 1e-12) from being serialised as real arcs.
3. **Solver status check** — validate `FBA.stat == 1` before consuming `FBA.full`. On failure, raise `writePajekNet:lpFailed` instead of silently emitting a graph from an invalid solution or crashing on indexing.
4. **Robust biomass / objective exclusion** — make the substring filter case-insensitive (`Biomass_*` no longer leaks through) and additionally exclude any reaction whose linear-objective coefficient (`model.c`) is non-zero.
5. **Test rewrite** — the previous test only compared the produced `.net` byte-for-byte against a frozen fixture, so any of the bugs above could pass. The new test:
   - parses the file and asserts the `*Vertices` block matches `model.mets`,
   - requires a `*Arcs` header (directed) with a count matching the number of arc lines,
   - validates each arc has in-range vertex ids and a positive weight,
   - confirms biomass / objective fluxes do not appear as arc weights,
   - exercises tolerance filtering (large `fluxTol` ⇒ zero arcs),
   - exercises the LP-failure path (infeasible bounds ⇒ `writePajekNet:lpFailed`).
   The obsolete `COBRAmodeltest.net` fixture is removed.

## Backward compatibility

- Calling `writePajekNet(model)` still works; new arguments are optional.
- The output file remains a valid Pajek `.net`; consumers that ignored the previous `*Edges` vs `*Arcs` distinction are unaffected, while consumers that respect it now get the correct semantics.
- The default `fluxTol` matches the configured solver tolerance, so any reaction the solver itself considers active is still emitted.

## Files

- `src/base/io/utilities/writePajekNet.m` — fixes 1–4, plus optional `fluxTol` and `fileName` arguments and `onCleanup`-guarded `fclose`.
- `test/verifiedTests/base/testwritePajekNet/testwritePajekNet.m` — semantic test (fix 5).
- `test/verifiedTests/base/testwritePajekNet/COBRAmodeltest.net` — removed (obsolete fixture).


farid-zare:
@codex

PR #2658, opened 2026-05-08
Fix existence check for 'param' variable

farid-zare:
I believe this one also should not be input as a string


farid-zare:
@codex

